PR #2, opened 2022-05-05
Add Spiegelman Drucker-Prager case

stephankramer:


drhodrid:
Looks good to me. Just one typo that I spotted: converge --> convergence on plotting script. Otherwise good to go.

PR #3, opened 2022-05-05
Add testing framework

angus-g:
There are two places we select the actual tests to run: `.github/workflows/test.yml` can select pytest cases using the `-k` flag, and the `Davies_etal_GMD_2021/Makefile` has a list of subdirectories to recurse into for tests. Because the 3D cases are so expensive, I've disabled them for the moment.

Another important consideration is that I've added a header row to the `params.log` files so we can easily pull out a pandas dataframe in the test script — this is a modification from the listings in the paper. If we want to leave the source files completely unmodified, the Makefile could add this header out-of-band.

For this first iteration of the testing framework, I'm just ensuring that the `u_rms` and `nu_top` match in the last row of this file between a run, and the expected results. I've also disabled all the PETSc monitor flags so the output on Actions isn't so noisy. Although, because we're still using `DumbCheckpoint` the output is still littered with deprecation warnings!


stephankramer (changes requested):
This is nice! All happy with that. Just two further request:
* re. your comment on changing the base case. The actual base code that appears in the paper is in minimal_base_case.py. Looking at the diff between minimal_base_case.py and base_case.py I'm getting a little nervous that they have diverged it bit too much - and just discussing this @drhodrid we don't want to change the base_case.py at this stage to make it more inline with the minimal case (as the former is consistent with what's run in later cases). Could you just add a third, say minimial_base_case_regression.py that is exactly the same as minimal_base_case.py but just with the minimal amount of lines added (I guess `log_params` definition and calls) so that it can be tested as well? Just so I can see immediately from the diff that it is the same as minimal_base_case.py?

* I find the dual use of test_all.py to both check the output of all tests, and generate it to be a little confusing, and couldn't immediately figure out what was going on from the docs/testing.md without reading the code. Would it maybe make more sense to split the `if __name__ == "__main__":` bit in a separate, say generate_expected_output.py, script which simply imports test_all.py ?

angus-g:
Thanks!

1. Just so I understand, you'd like an additional regression on `minimal_base_case.py`, but keep everything else as well? If so, can do!
2. Yeah, makes sense — that was probably just me being lazy when I was prototyping this stuff ages ago. I'll split the functionality.

stephankramer:
Yes, exactly - so we test both base_case.py and minimal_case.py.

angus-g:
I think it should be good to go, assuming this latest commit gets the green tick!

stephankramer:
Maybe you want to update the docs/testing.md ?

angus-g:
Oops, good spot! I got too excited about the other changes...

stephankramer (approved):
Splendid - merge away!

PR #4, opened 2022-05-11
CFL based timestep should be based on max *absolute* ref. velocity

stephankramer:


stephankramer:
Rebased on testing branch so that it runs regressions tests.

angus-g:
Looks like this one is failing because the number of timesteps is different, I don't know if that's a useful metric to even test but I guess pandas is labelling the dataframes anyway.

stephankramer:
Hm yeah, this probably requires some more thought. Getting the exact same answer for all timesteps isn't really expected for these cases. They're steady state cases and we're using the timestepping as a means to converge to it. Maybe we should have a separate output file where for each case we carefully select what we think the values are that shouldn't change - so for these steady state cases that might just be the Nusselt and u_rms when it's reached steady state (also rounded to some precision that we think is appropriate) and perhaps number of timesteps, again with some precision. Hm yeah so maybe it would record a tolerance, just specifying by n/o digits is a bit limited - at which point we may as well explicitly write the test in the case itself - you know just a np.testing.assert_almost_equal(...).  Thoughts?

There's always a bit of an art of finding the right statistics to test against. If the number changes with every trivial code change, it just gets annoying and you'll get into the habit of just blindly regenerating the expected output, but if it's too loose then it of course starts defeating the purpose as well...

angus-g:
I think a few of your points are still covered by us using a pandas `Series` for recording the expected result: the `pd.testing.assert_series_equal` function has a bunch of parameters similar to the numpy equivalent, such as `atol` and `rtol`. We can also pass `check_names=False` to not check that the conversion took the same number of timesteps. If we want some ballpark comparison of the number of timesteps, we have that available too.

If all the tests are going to be for convergence, then I think we just need the flexibility to choose different metrics to compare, so there is definitely room to make the testing framework a little bit more general.

I do agree that the tests shouldn't be so tightly-constrained that minor changes are going to require regenerating the reference results though! It is nice to have full-precision results that we can then subsample as needed in the tests.

drhodrid:
Looks fab!

drhodrid:
These are steady-state tests - is it converging towards a steady-state in less or more time steps? Is it the same steady-state solution (i.e. are all other diagnostics the same aside from # time steps)? I think it’s worth testing this actually - but we may need to update our “reference value” for the # time steps?

R

On 12 May 2022, at 10:19 am, Angus Gibson ***@***.******@***.***>> wrote:



Looks like this one is failing because the number of timesteps is different, I don't know if that's a useful metric to even test but I guess pandas is labelling the dataframes anyway.

PR #5, opened 2023-01-04
Add CI to restructure branch

angus-g:
This adds the Lint and Testing CI steps to the restructure branch. Most of the code changes are related to the linting, but there are a few additional changes:
- the `GADOPT_LOGLEVEL` environment variable can be used to control the logging verbosity. It's very noise to have every solve on stdout, maybe we could capture this and only output it if an error occurs? For now, I'm just changing to the `WARNING` loglevel for testing
- currently only testing the base case, but I've added a line to the output to name the columns for the parameters, so we can reuse the same testing code as before

If this is okay, don't merge through GitHub, I'll need to change the branch names to trigger the workflows, etc.


stephankramer (approved):
Nice, I like the GADOPT_LOGLEVEL. Tbh I think the default at INFO should be what it is, i.e. as a minimum show nonlinear convergence for stokes, and a single line for passing the energy solve - I prefer noisy over not knowing what it's up to: in any realistic run these timesteps (or even each nonlinear iteration) are not gonna go by all that fast. If you think it's too noisy for the test suite, sure let's just set it to WARNING for those then.

angus-g:
> Tbh I think the default at INFO should be what it is, i.e. as a minimum show nonlinear convergence for stokes, and a single line for passing the energy solve - I prefer noisy over not knowing what it's up to: in any realistic run these timesteps (or even each nonlinear iteration) are not gonna go by all that fast. If you think it's too noisy for the test suite, sure let's just set it to WARNING for those then.

Yep, fair enough. Maybe for the CI I'll pipe the output somewhere and just dump it (or maybe I can upload it as an artifact) only if there's a failure.

PR #8, opened 2023-06-01
Defer setup of timestepper+solver for energy equation

stephankramer:
This is to be consistent with Stokes solver and allows users to override parameters setup at initialisation.

### DIFF
--- a/gadopt/energy_solver.py
+++ b/gadopt/energy_solver.py
@@ -1,4 +1,4 @@
-from firedrake import DirichletBC
+from firedrake import DirichletBC, Function
 from .scalar_equation import EnergyEquation
 from .utility import is_continuous, ensure_constant
 from .utility import log_level, INFO, DEBUG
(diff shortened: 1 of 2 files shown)


angus-g:
I expect this is probably fine, there's just some weird issue with Docker preventing the CI from running correctly!

gadopt/energy_solver.py
```diff
@@ -68,10 +70,22 @@ def __init__(self, T, u, approximation,
...
+        # so people can overwrite parameters we've setup here
+        self._solver_setup = False
+
+    def setup_solver(self):
+        """Setup timestepper and associated solver"""
+        self.ts = self.timestepper(self.eq, self.T, self.fields, self.delta_t,
+                                   bnd_conditions=self.weak_bcs, solution_old=self.T_old,
+                                   strong_bcs=self.strong_bcs,
+                                   solver_parameters=self.solver_parameters)
+        self._solver_setup = True
```
angus-g, line 85: I guess an alternative would be to just check for the presence of `self.ts`, but this is probably better for enforcing construction of the timestepper.

angus-g (approved):
Looks fine to me!

PR #9, opened 2023-06-23
Fix direct solver options for energy equation

angus-g:
Well spotted by @rhodrid!


drhodrid (approved):

drhodrid:
Assuming the test suite passes, all good!

PR #10, opened 2023-07-07
Convert Constants with domains to Functions

angus-g:
This will be removed, currently raising a FutureWarning. Hopefully this is the right way to do things?

### DIFF
--- a/demos/2d_cylindrical/2d_cylindrical.py
+++ b/demos/2d_cylindrical/2d_cylindrical.py
@@ -110,8 +110,8 @@ for timestep in range(0, max_timesteps):
     f_ratio = rmin/rmax
     top_scaling = -1.3290170684486309  # log(f_ratio) / (1.- f_ratio)
     bot_scaling = -0.7303607313096079  # (f_ratio * log(f_ratio)) / (1.- f_ratio)
-    nusselt_number_top = (assemble(dot(grad(T), n) * ds_t) / assemble(Constant(1.0, domain=mesh)*ds_t)) * top_scaling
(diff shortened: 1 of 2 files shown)


demos/2d_cylindrical/2d_cylindrical.py
```diff
@@ -110,8 +113,8 @@
     f_ratio = rmin/rmax
     top_scaling = -1.3290170684486309  # log(f_ratio) / (1.- f_ratio)
     bot_scaling = -0.7303607313096079  # (f_ratio * log(f_ratio)) / (1.- f_ratio)
-    nusselt_number_top = (assemble(dot(grad(T), n) * ds_t) / assemble(Constant(1.0, domain=mesh)*ds_t)) * top_scaling
-    nusselt_number_base = (assemble(dot(grad(T), n) * ds_b) / assemble(Constant(1.0, domain=mesh)*ds_b)) * bot_scaling
+    nusselt_number_top = (assemble(dot(grad(T), n) * ds_t) / top_area) * top_scaling
+    nusselt_number_base = (assemble(dot(grad(T), n) * ds_b) / bottom_area) * bot_scaling
```
angus-g, line 117: Or maybe we can just remove `domain=mesh` from the Constant without penalty? It does seem a bit dumb to have to use Functions to do it...

stephankramer (approved):
Yes, I think this is right. And I think in the places you've changed it, it actually makes more sense to attach it to the measure.

The only thing we have to watch out for is that if you assign values to a Constant it no longer works in the adjoint (without a domain). Then you do need to use a Function on the R space instead (i.e. the thing you actually get at the moment when you do Constant(..., domain=...) with a warning).

I _think_ currently we only do this for the timestep? And having a variable timestep doesn't really work with the adjoint anyway, so not a problem. But something to keep in mind if you do an inversion/sensitivity analysis with respect to a constant control as well....

PR #11, opened 2023-08-01
Reduce convergence requirement of 2D cylindrical

angus-g:
The tolerance requirement of 1e-9 for convergence of the 2D cylindrical case is pretty tight. This is probably a good thing in general, but it takes a lot of our CI budget (the other big user being the analytical cases). We can chop a lot of time off this case by dropping the tolerance to 1e-7.


angus-g:
Oops, I guess the CI on this won't pass because the convergence numbers require the Ttheta changes first. @drhodrid maybe I can merge the changes for the cases other than the parallel scaling (which we can do later?)

drhodrid:
Hey @angusg - I just looked at the remove theta branch again and I think it’s good to go - aside from the fact that we haven’t updated the 2D cylindrical case there (which you may have done as part of this commit). So aside from that caveat, I’m happy for you to merge - and we can sort parallel scaling once we have a better idea of GAMG outcomes.

On 1 Aug 2023, at 5:52 pm, Angus Gibson ***@***.***> wrote:



Oops, I guess the CI on this won't pass because the convergence numbers require the Ttheta changes first. @drhodrid<https://github.com/drhodrid> maybe I can merge the changes for the cases other than the parallel scaling (which we can do later?)

demos/test_all.py
```diff
@@ -23,5 +22,4 @@ def test_benchmark(benchmark):
 
     pd.testing.assert_series_equal(df[["u_rms", "nu_top"]], expected, check_names=False)
 
-    if df.name != expected.name:
-        warnings.warn(f"Convergence changed: expected {expected.name}, got {df.name}")
+    assert abs(df.name - expected.name) <= 2
```
stephankramer, line 25: What's going here? What is df.name and why is it a value that is allowed to be 2 different from expected?

angus-g: Yeah I'm not much of a pandas user... The "name" of the series is its row number, which ends up being the iteration. So we want to (by @drhodrid's suggestion) make sure we're within 2 of the expected number of iterations to convergence. I guess extra documentation would be good here!

stephankramer: Same here - I find their API rather ugly. Yes, makes sense - a comment would be good, yeah

stephankramer (approved):
Nice!

angus-g:
In case it's useful @drhodrid, here are the two parameter logs for the cylindrical case (original first, then the new version with the Ttheta fix):
- [params_orig.log](https://github.com/g-adopt/g-adopt/files/12247009/params_orig.log)
- [params_new_fixed.log](https://github.com/g-adopt/g-adopt/files/12247007/params_new_fixed.log)

drhodrid:
Hey @angus-g - I looked at this case and I think it's fine to go ahead and update the expected values. I suspect the 3 timestep difference is because: (i) we expect 1 timestep variation from the Ttheta change; (ii) minor roundoff issues in this case, perhaps due to nullspaces which we don't have in any of the other examples we have examined thus far. But the results look as expected - so happy for you to push on.

angus-g:
Should I make the expected value for the cylindrical case the results of the old code run + slacken the iteration tolerance?

drhodrid:
I’d recommend updating the expected values to the new run…

R

On 4 Aug 2023, at 10:47 am, Angus Gibson ***@***.***> wrote:



Should I make the expected value for the cylindrical case the results of the old code run + slacken the iteration tolerance?

angus-g:
Easy, will do!

PR #13, opened 2023-08-07
Add 3D spherical and cartesian as tested cases

angus-g:
The `3d_cartesian` case takes somewhere between 5-10 hours to run (unless I'm doing something wrong!), so that's not included here.


drhodrid:
Hey @angus-g  - this looks good to me. I will look at the 3D Cartesian case now to see if I can make it quicker.

angus-g:
Strange that I can't use `GeodynamicalDiagnostics` -- seems the Nusselt number definition is different from the one used in the script:

```python
# script
nusselt_number_base = (assemble(dot(grad(T), n) * ds_b) / assemble(Constant(1.0) * ds_b(domain=mesh))) * (rmin*(rmax-rmin)/rmax)

# gadopt
    def Nu_bottom(self):
        return assemble(dot(grad(self.T), self.n) * self.ds_b) * (1./assemble(self.T * self.ds_b))
```

I guess we normalise by the surface temperature in gadopt, but not in the script...

drhodrid:
Hi Angus,

Could you still use the Geodynamical Diagnostics, for consistency, and just do the scaling ( * (rmin*(rmax-rmin)/rmax)) on the result?

Or am I missing something?

R

On 9 Aug 2023, at 8:49 pm, Angus Gibson ***@***.***> wrote:



Strange that I can't use GeodynamicalDiagnostics -- seems the Nusselt number definition is different from the one used in the script:

# script
nusselt_number_base = (assemble(dot(grad(T), n) * ds_b) / assemble(Constant(1.0) * ds_b(domain=mesh))) * (rmin*(rmax-rmin)/rmax)

# gadopt
    def Nu_bottom(self):
        return assemble(dot(grad(self.T), self.n) * self.ds_b) * (1./assemble(self.T * self.ds_b))

I guess we normalise by the surface temperature in gadopt, but not in the script...

angus-g:
> Could you still use the Geodynamical Diagnostics, for consistency, and just do the scaling ( * (rmin*(rmax-rmin)/rmax)) on the result?

That's what I did in the latest commit, but `nu_top` changes pretty significantly (from -0.999 to 3.308):
```
E   [index]: [u_rms, nu_top]
E   [left]:  [1.258677490685418, 3.30766875676134]
E   [right]: [1.258687832524867, -0.9989337081904872]
```

angus-g:
I've just added the cartesian case too, but still expecting a failure on the check because I haven't fixed the issue with the diagnostics on the spherical case.

PR #19, opened 2023-09-21
Fix timestep adaptor

stephankramer:
Two bugs:
1) I was taking np.abs(uref.dat.data.max()), i.e. the maximum over
   _signed_ reference velocities first, thus ignoring any large negative
values
2) (less important) the expression that computes the reference velocity
   is disconintinuous but gets interpolated to the continuous velocity
space (and thus the result may depend on mesh ordering). What we want instead is take the maximum over all adjacent elements in each node, which is what you can get using an interpolator with acccess=op2.MAX

Since I'm now using an interpolator (also possibly marginally faster) the (physical) velocity that is used needs to be supplied already when creating the TimestepAdaptor. Updated demos to corresponding interface change.


stephankramer:
@drhodrid : were you previously getting different results on the TALA case than the expected.pkl ? If so could you rerun with this branch and see if they are now consistent?

angus-g:
> @drhodrid : were you previously getting different results on the TALA case than the expected.pkl ? If so could you rerun with this branch and see if they are now consistent?

I think the different results were reproducible by everybody. The `expected.pkl` is generated using 4 cores, but @drhodrid's results were on a single core.

stephankramer:
So for the TALA case, on 1 core on my laptop:
```
7612 0.5765643482639519 7.49303697578308e-05 9.990851401584303e-10 156.29198042226741 178.23483897152877 231.72971386780142 7.912901115561309 7.795334758038918 0.11756635752239042 0.5455412460985543 3.247259920922624 3.276450828771609 0.029190907848985148
```
, and on 4 cores on Kerguelen:
```
7612 0.5765643482639525 7.493036975783056e-05 9.9908513700258e-10 inf 178.23483897152772 231.7297138678002 7.912901115561208 7.795334758038864 0.11756635752234423 inf 3.247259920922522 3.2764508287715746 0.02919090784905265
```
So the n/o timesteps and the Nusselt numbers are bang on, but I'm getting `inf`s for `u_rms` and `T_avg`. Hm, must be something to do with `domain_volume` in the GeodynamicalDiagnostics. But I don't see what?!

angus-g:
Possibly unrelated, but in `diagnostics.py` we get the mesh through `u.ufl_domain()`, whereas in `utility.py` it's `u.function_space().mesh()`.

stephankramer:
Hm on Kerguelen:
```
$ python -c "from firedrake import *; mesh=UnitSquareMesh(40,40, quadrilateral=True); print(assemble(Constant(1)*dx(domain=mesh)))"
0.0
```
Something is very broken on my fresh firedrake build on Kerguelen!

stephankramer:
> Possibly unrelated, but in `diagnostics.py` we get the mesh through `u.ufl_domain()`, whereas in `utility.py` it's `u.function_space().mesh()` (and there have been a lot of UFL changes recently!)

Yes, unfortunately we also need `u.function_space()` to create a new function in a function space that is (equivalent to) the velocity space, which isn't straightforward to peel out of a `u,p = split(z)` velocity (also depends on whether it's a VectorFunctionSpace or a vector-valued FunctionSpace). So I've now gone for the approach of separately specifying which reference velocity space to use - should be more generic as well....

angus-g (approved):
Looks good!

stephankramer:
Thanks, and yes, let's do this one next. If it passes

PR #26, opened 2023-09-27
Add spherical analytical with long test framework

angus-g:
This finally adds the more expensive spherical analytical cases, as well as the testing framework that will let us fire off automated tests on HPC (just Gadi for the moment).

- [x] check out and copy repository to Gadi
- [x] set up SSH access to run tests
- [x] run tests using existing `make` framework
- [x] execute tests in parallel
- [x] retrieve failed runs
- [x] test results
  - maybe do this on the remote machine too, by adding a pytest tag for longtest results? the SSH command will error if pytest fails
- [x] clean up test artefacts

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -7,7 +7,7 @@ jobs:
     name: "Run linter"
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Setup Python
         uses: actions/setup-python@v4
         with:
--- a/.github/workflows/longtest.yml
+++ b/.github/workflows/longtest.yml
@@ -33,7 +33,7 @@ jobs:
 
       - name: Test SSH
         run: |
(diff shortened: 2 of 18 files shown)


angus-g:
Successful workflow run: https://github.com/g-adopt/g-adopt/actions/runs/6506115687

I've disabled it on pull request, so it won't be run for the actual commit that will eventually be merged.

drhodrid (approved):
This all looks good to me. 

One thing that I think we need, now that the CI is getting close to what we feel is robust (which shouldn't delay this pull request!) is documentation in relation to the test suite. How is it set up? How are individual tests run? How can these tests be run locally? What triggers long-tests versus standard tests etc...? 

For now, this looks great @angus-g! Thanks for your efforts! Am I right that we next need GAMG merges into Firedrake and then parallel testing can be incorporated?

angus-g:
Good points, documentation would definitely be a good focus while it's all still fresh in my head too!

PR #28, opened 2023-10-04
Adjust penalty term in weak imposition of velocity boundary conditions.

drhodrid:
The previous value of 6.25 was set to match the value in Davies et al. 2022 (GMD), for 2-D cases. I have found improved solver convergence with a lower value of 2. My tests also suggest improved accuracy for analytical test cases with free-slip boundary conditions. Not to be merged immediately - would prefer for this to go in after analytical spherical tests are in and we have fully resolved all outstanding GAMG related issues, but getting the pull request in now, to make sure things remain green in the CI.


drhodrid:
Just an update. Given issues with upstream PETSc, which has been merged into Firedrake, the plan is now to merge this branch first, pending green actions and a review.

angus-g:
So in addition to the slight change to the converged result `2d_cylindrical` and `3d_spherical` cases, there's a change in the number of iterations to converge the `2d_cylindrical` case. This goes from 3027 iterations to 3038. @drhodrid is that expected? If so, we'll have to update the expected values for this test.

Also, should we do this by bumping down the relative tolerance from the default 1e-4 to 1e-3, or update the expected values? Arguably this does change the system we're solving, so it seems to me that both approaches are justifiable.

drhodrid:
Hi @angus-g - I think this change in the number of iterations is fine. With the new penalty term, we are solving an ever so slightly different problem, which just takes a little longer to reach the steady-state that we have specified. My preference would be to keep the steady-state tolerance at 1e-4 and adjust the expected values: both strategies are probably fine, but the former keeps some consistency with how we have done things previously.

angus-g:
Tada! Were you happy with this too @stephankramer?

stephankramer (approved):
Ship it!

angus-g (approved):
I think we are good to go on this.

PR #31, opened 2023-10-09
Lower tolerances, to avoid the occasional inner solve failure.

drhodrid:
Slacken solver tolerances marginally for analytical cases. This (for my testing at least!) seems to reduce the number of inner solve failures, and in the long-term should improve stability of the tests. I wonder if some of this has contributed to the timing issues @stephankramer has been seeing. I'll dig deeper tomorrow.


angus-g (approved):
> 1e-14 is right at the limit of what the solvers can do... and convergence towards analytical solutions still seems to satisfy the test suite

Sounds fine to me.

PR #32, opened 2023-10-10
Enforce implementation of abstract methods

Patol75:
I was reading through `approximations.py` and discovered the concept of an abstract base class. However, I think there is an issue with how it is currently implemented. From what I read, the `@abc.abstractmethod` decorator is only useful if the method it targets belongs to a class that is derived from `abc.ABCMeta`. Indeed, in the current implementation, one could write
```python
from gadopt.approximations import BaseApproximation


class Foo(BaseApproximation):
    pass


a = Foo()
```
and no error would be raised.

### DIFF
--- a/gadopt/approximations.py
+++ b/gadopt/approximations.py
@@ -3,7 +3,7 @@ from firedrake import sym, grad, inner, div, Identity
 from .utility import ensure_constant, vertical_component
 
 
-class BaseApproximation(metaclass=abc.ABCMeta):
+class BaseApproximation(abc.ABC):
     """
     Base class to provide expressions in (Navier?)-Stokes + energy equations
 
--- a/gadopt/approximations.py
+++ b/gadopt/approximations.py
@@ -3,7 +3,7 @@ from firedrake import sym, grad, inner, div, Identity
 from .utility import ensure_constant, vertical_component
 
 
-class BaseApproximation:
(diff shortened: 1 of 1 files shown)


gadopt/approximations.py
```diff
@@ -3,7 +3,7 @@
 from .utility import ensure_constant, vertical_component
 
 
-class BaseApproximation:
+class BaseApproximation(metaclass=abc.ABCMeta):
```
angus-g, line 6: I think inheriting `abc.ABC` directly would be a little cleaner and more idiomatic, i.e.

```python
class BaseApproximation(abc.ABC):
```

Patol75: I saw that argument, yes, but then I also found that Python developers seem to use the `metaclass` keyword (at least [here](https://github.com/python/cpython/blob/67e8d416cc5e14c181e4f883ffb4d87f2a647ae1/Lib/_collections_abc.py#L131)). Up to you honestly. :)

angus-g: I think it's a bit of both (e.g. [PathLike](https://github.com/python/cpython/blob/67e8d416cc5e14c181e4f883ffb4d87f2a647ae1/Lib/os.py#L1084)). Consistency be damned!

angus-g (approved):
Excellent, thanks!

PR #33, opened 2023-10-11
Fix adoint cylindrical

sghelichkhani:
This PR fixes an inconsistency in defining the strain-rate for non-linear cases. The definition of `forward.py` currently uses an additional factor of $\sqrt{1/2}$, that should actually appear in the definition for `epsii`. 

To make the two scripts consistent, and also the same as the original scripts for running the adoints we are now using a higher `sigma_y`.

### DIFF
--- a/demos/adjoint_2d_cylindrical/forward.py
+++ b/demos/adjoint_2d_cylindrical/forward.py
@@ -43,7 +43,6 @@ def run_forward():
     # Set up function spaces for the Q2Q1 pair
     V = VectorFunctionSpace(mesh, "CG", 2)  # Velocity function space (vector)
     W = FunctionSpace(mesh, "CG", 1)  # Pressure function space (scalar)
-    Q = FunctionSpace(mesh, "CG", 2)  # Temperature function space (scalar)
     Q1 = FunctionSpace(mesh, "CG", 1)  # Average temperature function space (scalar, P1)
     Z = MixedFunctionSpace([V, W])
 
--- a/demos/adjoint_2d_cylindrical/forward.py
+++ b/demos/adjoint_2d_cylindrical/forward.py
@@ -51,13 +51,19 @@ def run_forward():
     z = Function(Z)  # a field over the mixed function space Z.
     u, p = split(z)  # Returns symbolic UFL expression for u and p
 
-    T = Function(Q, name="Temperature")
     X = SpatialCoordinate(mesh)
+    r = sqrt(X[0] ** 2 + X[1] ** 2)
+    Ra = Constant(1e7)  # Rayleigh number
+    approximation = BoussinesqApproximation(Ra)
+
+    # Define time stepping parameters:
+    max_timesteps = 200
+    delta_t = Constant(5e-6)  # Constant time step
+
     with CheckpointFile("Checkpoint230.h5", mode="r") as f:
         T = f.load_function(mesh, "Temperature")
 
     Taverage = Function(Q1, name="Average Temperature")
-    r = sqrt(X[0] ** 2 + X[1] ** 2)
 
     # A step function designed to design viscosity jumps
     # Build a step centred at "centre" with given magnitude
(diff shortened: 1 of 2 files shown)


drhodrid (changes requested):
I think it's worth adding a short comment next to your definitions of epsii of why you are including the factor of 0.5 (for clarity - because this is different to how we have it in the visco-plastic demo, for example). You should also perhaps say a little about where the 1.41 and 2.82 come from (a scaling from a sqrt(2) error - right)? Fianlly - remove adjoint_spherical_test directory (looks like you just did!).

sghelichkhani:
Note that this also fixes an issue with loading mesh from `Checkpoint230.h5`, whereas this should load `Checkpoint_State.h5` to preserve the partitioning information from the forward run.

demos/adjoint_2d_cylindrical/forward.py
```diff
@@ -85,9 +85,9 @@ def step_func(centre, mag, increasing=True, sharpness=50):
     mu_lin *= exp(-ln(Constant(80)) * T)
 
     # Assemble the viscosity expression in terms of velocity u
-    eps = 0.5 * sym(grad(u))
-    epsii = sqrt(inner(eps, eps))
-    sigma_y = 1e4 + 2.0e5 * (rmax - r)
+    eps = sym(grad(u))
+    epsii = sqrt(0.5 * inner(eps, eps) + 1e-10)
+    sigma_y = 1.4142e4 + 2.8284e5 * (rmax - r)
```
angus-g, line 90: Should these be e.g. `1e4 * sqrt(2)` and `2e5 * sqrt(2)` or something instead of just floats?

sghelichkhani:
> I think it's worth adding a short comment next to your definitions of epsii of why you are including the factor of 0.5 (for clarity - because this is different to how we have it in the visco-plastic demo, for example). You should also perhaps say a little about where the 1.41 and 2.82 come from (a scaling from a sqrt(2) error - right)? Fianlly - remove adjoint_spherical_test directory (looks like you just did!).

Yeah thanks! I got rid of an additional factor of 2 (I don't know why I thought that's how we define `epsii`). The values are consistent with Coltice et. al 2017: We use a surface yield stress, its depth derivative of 2e4, and 4e5, compared to 2e4 and 2.5e5 used in Coltice etl al. I have now added a reference there. 

Also counting for the additional factor of 2, we don't need to use floats anymore.

demos/adjoint_2d_cylindrical/forward.py
```diff
@@ -85,9 +85,11 @@ def step_func(centre, mag, increasing=True, sharpness=50):
     mu_lin *= exp(-ln(Constant(80)) * T)
 
     # Assemble the viscosity expression in terms of velocity u
-    eps = 0.5 * sym(grad(u))
-    epsii = sqrt(inner(eps, eps))
-    sigma_y = 1e4 + 2.0e5 * (rmax - r)
+    eps = sym(grad(u))
+    epsii = sqrt(inner(eps, eps) + 1e-10)
+    # yeild stress and its depth dependence
```
stephankramer, line 90:
```suggestion
    # yield stress and its depth dependence
```

drhodrid:
OK - so this inverse run now fails when I run it (KSP_DIVERGED_MAX_IT). It turns out that your forward and adjoint models are now inconsistent again in the definitions of epsII. This needs to be fixed. Could you also ensure that the differences in forward.py and inverse.py are solely those arising from the addition of the adjoint related functionality? For example, at the moment, some comments are inconsistent and the ordering of some things seems a little different too (although the latter maybe required - I'm not sure). This makes things harder to debug (e.g. diffs highlight things that shouldn't be different). So perhaps use meld or similar to highlight the differences, and line things up better.

angus-g:
Suggests that maybe the two should use the same code (inherit the same function)… Then it’s not as prone to manual errors.

Also, given the checkpoint file was renamed, does the file in Git LFS need to be renamed too?

sghelichkhani:
> OK - so this inverse run now fails when I run it (KSP_DIVERGED_MAX_IT). It turns out that your forward and adjoint models are now inconsistent again in the definitions of epsII. This needs to be fixed. Could you also ensure that the differences in forward.py and inverse.py are solely those arising from the addition of the adjoint related functionality? For example, at the moment, some comments are inconsistent and the ordering of some things seems a little different too (although the latter maybe required - I'm not sure). This makes things harder to debug (e.g. diffs highlight things that shouldn't be different). So perhaps use meld or similar to highlight the differences, and line things up better.

@drhodrid  Hmm... I am not sure what you mean with inconsistency.  Might be you had not pull the changes? The two files were originally different after the original PR. Now  they are as similar as necessary. 

@angus-g Yeah! It would make sense to inherit to ensure consistency. But maybe it's a bit of an overkill for a synthetic test?

Re checkpointfile: No. Renaming of the checkpoint file in `inverse.py` ensures that we use the same partitioning that is used for the reference fields. Currently we load the mesh from `Checkpoint230.h5` but then load the fields from `Checkpoint_State.h5`. So we are not being consistent. By using `Checkpoint_State.h5` we stay consistent.

drhodrid:
Thanks @sghelichkhani - this now runs for me and things look sensible. I'm happy to approve (and we can revisit @angus-g comment about the inverse code inheriting the forward model at a later date).

drhodrid (approved):

PR #35, opened 2023-10-11
Adding multi-material capabilities to g-adopt through level sets

Patol75:
This is currently a rough proof of concept for the integration of level-set methods in g-adopt. I have used the most simple approach (no conservative level set, no reinitialisation) to model the benchmark of [van Keken et al. (1997)](https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/97JB01353). I will keep on adding functionalities. Any feedback welcome.


Patol75:
Just a quick update here:
- Conservative level set works fine
- High-order DQ approximations work fine
- Projection from DQ to CG works fine
- Time-stepping of the reinitialisation (equation 17 of [Parameswaran and Mandal, 2023](https://www.sciencedirect.com/science/article/pii/S0997754622001364)) does not give anything sensible

For the latter, I do not quite understand what is going on. I imagine the reinitialisation solver should be almost identical to the level-set solver except for the terms included in the equation, which is how I set things up. If I run the code with the reinitialisation solver, it is as if the level set function is constantly brought back to the initial condition, but the gradient projection still shows some advection. I am confused.

Patol75:
Reinitialisation of the conservative level set now works. Happy to get reviews while I implement a few tests.

Just a comment that [W503 should be disabled](https://stackoverflow.com/questions/73568036/flake8-line-break-before-binary-operator-how-to-fix-it).

Patol75:
Example results from a run on Gadi (69 minutes on a single Cascade Lake node)
[rms_velocity_and_entrainment.pdf](https://github.com/g-adopt/g-adopt/files/13111657/rms_velocity_and_entrainment.pdf)
[g-adopt_level_set_van_keken.pdf](https://github.com/g-adopt/g-adopt/files/13111658/g-adopt_level_set_van_keken.pdf)

Patol75:
Following the latest commit, here is a 512x128 solution for the thermochemical benchmark of van Keken et al. (1997).
https://youtu.be/YATMAk0CMdw

Patol75:
@stephankramer @drhodrid I believe this is now ready for a review.

Known issues:
- The Schmalholz benchmark does not behave well if the level set is defined on DQ2.
- Thermochemical benchmarks require a lower CFL than chemical benchmarks to converge to published results. I do not think this should be the case.
- Subcycling degrades the solution, and I believe it is related to the above point.
- Results from Trim are not reproduced.

Patol75:
Given the previous commits, I would not mind if someone could have a look at the present code and, in particular, see if they have a feeling as to why, when the CFL increases to 0.65 (which should be stable), the solution to the Robey benchmark degrades.

Patol75:
Updated list of known issues:
- `schmalholz_2011` raises `DIVERGED_LINEAR_SOLVE` during the first Stokes solve if the viscosity around the interface is calculated with a geometric mean
- Increasing the degree of the level-set function space and the mesh resolution prevent lithospheric detachment in `schmalholz_2011`
- The air layer and its associated level set are prone to instabilities in `schmeling_2008`, especially when increasing the degree of the level-set function space or when the density contrast around the interface is not sharp
- The results of `robey_2019` seem to not exactly reproduce the published results
- The evolution of the temperature field in `trim_2023` is out of sync with the analytical internal heating rate; specifying `B = 0.46` seems to improve results, but that highlights a serious problem somewhere in the code

@stephankramer @drhodrid I would not mind having some help with the above; I am running short of ideas, unfortunately.

Patol75:
Regarding the first bullet point above, the issue seems to go away if I use
```python
@classmethod
def viscosity(cls, velocity):
    strain_rate = fd.sym(fd.grad(velocity))
    strain_rate_sec_inv = fd.sqrt(fd.inner(strain_rate, strain_rate) / 2)

    visc = cls.visc_coeff * strain_rate_sec_inv ** (1 / cls.stress_exponent - 1)

    return fd.conditional(
        visc > cls.visc_bounds[1],
        cls.visc_bounds[1],
        fd.conditional(visc < cls.visc_bounds[0], cls.visc_bounds[0], visc),
    )
```
instead of
```python
@classmethod
def viscosity(cls, velocity):
    strain_rate = fd.sym(fd.grad(velocity))
    strain_rate_sec_inv = fd.sqrt(fd.inner(strain_rate, strain_rate) / 2)

    return fd.min_value(
        fd.max_value(
            cls.visc_coeff * strain_rate_sec_inv ** (1 / cls.stress_exponent - 1),
            cls.visc_bounds[0],
        ),
        cls.visc_bounds[1],
    )
```
Don't ask me why!

Patol75:
Regarding the second bullet point, failure to detach only seems to occur if I set `level_set_func_space_deg = 2`. Results with `level_set_func_space_deg = 1` look as expected.

stephankramer:
On the first point: this might be to do with some subtlety of the symbolic Jacobian not being defined in u=0. If you have a nonlinear strain-rate dependent velocity, the linearised system in the Newton iteration is not only your normal \mu(u) \nabla^2 del_u term (where del_u is the change in velocity), but also has a d\mu(u)/du del_u \nabla^2 u. That d\mu/du is not defined for u=0 due to the sqrt. Two ways to deal with this is: 1) to have the initial (guess for) velocity be just slightly nonzero:
```
u.interpolate(as_vector(1e-12, 1e-12))
```
or 2)  add an epsilon inside the sqrt:
```
strain_rate_sec_inv = fd.sqrt(fd.inner(strain_rate, strain_rate) / 2+eps)
```
where `eps` is suitably small wrt the expected strain rate.
Oh actually maybe 1) doesn't work here, since the strain rate of (1e-12, 1e-12) would still be zero. So yeah, maybe try 2)

Patol75:
@stephankramer Thank you for the explanation. Adding `1e-99` solves the issue when using `max_value` and `min_value`.

Patol75:
Another update on the third bullet point: from a few high-resolution tests on Gadi, it seems that having a sharp density interface helps a lot, as mentioned before.  Now to try to understand whatever is going wrong with the fourth and fifth bullet points. Again, any help is appreciated.

stephankramer (changes requested):
Some feedback on code structure. As a general point, which I think I mentioned before, this is really an impressive amount of work, getting all these different benchmarks to work - and this genuinely looks really very nice!
Sorry this doesn't really address the numerical issues you mentioned (except the coupled solver stuff might subtly change things)

stephankramer:
Oh and well done for getting further regarding the importance of the sharpness of the interface. I wonder how we compare in the details of how the levelset solution is "translated" into viscosity and density structure for the Stokes equation. I suspect some of the other codes do everything on the nodes (either P1/Q1 or P2/Q2) which would smear things out more than what we do which is evaluated on the gauss points. Oh actually, you could try using a higher, or automatic quadrature degree. gadopt uses a default quadrature degree in the stokes and energy solvers - you could for instance set it back to None (for "automatic").

demos/multi_material/benchmarks/trim_2023.py
```diff
@@ -0,0 +1,164 @@
...
+        ) / cls.Ra
+
+    @classmethod
+    def internal_heating_rate(cls, int_heat_rate, simu_time):
+        node_coords_x, node_coords_y = node_coordinates(int_heat_rate)
+
+        analytical_values = []
+        for coord_x, coord_y in zip(node_coords_x, node_coords_y):
+            analytical_values.append(
+                flib.h_python(
```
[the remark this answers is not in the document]
Patol75: Intriguing, isn't it? In Trim et al. (2023), they manufacture a benchmark solution. Most of it can be expressed analytically, but some parts (including the internal heating rate) are determined through some symbolic solver I believe. Accordingly, they provide a Fortran program and associated Python library to query values of these functions at any point and time. `flib.h_python` gives the internal heating rate.

demos/multi_material/material_interface.py
```diff
@@ -0,0 +1,42 @@
...
+                    + (1 - ls) / diffuse_interface(level_set, material_value, method)
+                )
+    else:
+        match method:
+            case "arithmetic":
+                return material_value[0] * (1 - ls) + material_value[1] * ls
+            case "geometric":
+                return material_value[0] ** (1 - ls) * material_value[1] ** ls
+            case "harmonic":
+                return 1 / ((1 - ls) / material_value[0] + ls / material_value[1])
```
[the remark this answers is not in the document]
Patol75: If I change some naming to make it more general, and we are happy with the fact that it should accept a Function with values in [0, 1], then why not.

demos/multi_material/run_benchmark.py
```diff
@@ -0,0 +1,303 @@
...
+from gadopt.energy_solver import EnergySolver
+from gadopt.level_set_tools import (
+    LevelSetEquation,
+    ProjectionSolver,
+    ReinitialisationEquation,
+    TimeStepperSolver,
+)
+from gadopt.stokes_integrators import StokesSolver, create_stokes_nullspace
+from gadopt.time_stepper import ImplicitMidpoint, eSSPRKs3p3, eSSPRKs10p3
+from gadopt.utility import TimestepAdaptor
```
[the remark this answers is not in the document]
Patol75: I am not sure if I am allergic to it, but VSCode is!

demos/multi_material/run_benchmark.py
```diff
@@ -0,0 +1,303 @@
+import firedrake as fd
+import numpy as np
+from benchmarks.schmeling_2008 import Simulation
```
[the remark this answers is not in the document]
Patol75: We could execute `run_benchmark.py` with a command-line argument, which would be the relative path to the setup file — for example, `"benchmarks.schmeling_2008"`. Then, in `run_benchmark.py`, we read the argument through `argparse` and import the `Simulation` class using the following logic:
```python
import importlib
Simulation = importlib.import_module(relative_path).Simulation
```

Patol75:
Current benchmark evolutions given the latest commit: https://www.youtube.com/playlist?list=PLCgkpRltN3N9CO-NXOtCww1bdSt4xpkB9
I will add diagnostic plots next week.

@stephankramer @drhodrid If you have any time next week to look at the PR, the most crucial point is the [Trim et al. (2023)](https://scholar.google.com/scholar_url?url=https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/2022GC010807&hl=en&sa=T&oi=gsb&ct=res&cd=0&d=4634233625786422446&ei=kFZ1ZcuSD8KXywTVk6bgCg&scisig=AFWwaeZbhteQY04alxyHLgbhqpRM) benchmark, which refuses to behave...

Patol75:
@angus-g Something weird: https://github.com/g-adopt/g-adopt/actions/runs/7156028140
"Annotations" report a failure in the 2D compressible ALA benchmark, which corresponds to the failure listed in the table under "Benchmark regression test summary". However, "Case timings" reports "Success" for all "Cases". Is this "Success" attribute supposed to highlight execution success or test validation success?
Moreover, I am pretty sure that the failure is unrelated to my commit changes, so you may want to investigate what happened there. :)

angus-g:
> Is this "Success" attribute supposed to highlight execution success or test validation success?

The former. Because the test runs on the latest upstream Docker image, there's probably been some change to Firedrake that has very slightly changed the convergence to be out of tolerance.

gadopt/approximations.py
```diff
@@ -75,19 +75,32 @@ class BoussinesqApproximation(BaseApproximation):
     Viscous dissipation is neglected (Di << 1)."""
     compressible = False
 
-    def __init__(self, Ra, kappa=1, g=1, rho=1, alpha=1):
+    def __init__(self, Ra, rho=1, g=1, alpha=1, kappa=1, RaB=0, delta_rho=1, H=0):
```
angus-g, line 78: Shouldn't affect your uses of `BoussinesqApproximation` or any of the existing demos, but I wonder if maybe this should enforce keyword-only args. Lots of numbers whose order could get confused!

```suggestion
    def __init__(self, Ra, *, rho=1, g=1, alpha=1, kappa=1, RaB=0, delta_rho=1, H=0):
```

demos/multi_material/benchmarks/crameri_2012.py
```diff
@@ -0,0 +1,173 @@
+from functools import partial
+
+import initial_signed_distance as isd
+
+import gadopt as ga
+
+
+class Mantle(ga.AbstractMaterial):
```
stephankramer, line 8: I'm wondering what the best structure here is. So it seems you are sub-classing specific classes for each specific Material from AbstractMaterial, and then you're only using these as a class, they're never substantiated, but the way you use them as a class, they really just act like an object of which you call some methods. So why not just:
```
Mantle = Material(B=None, RaB=None, density=3300, viscosity=1e21, thermal_expansion=1, ...)
```
where Material is a concrete (data) class that just sets a bunch of attributes - in fact you could use an actual [datclass](https://docs.python.org/3/library/dataclasses.html) for that.
Of course that would make them attributes rather than methods, but I wasn't immediately clear what the use case for them being methods was. In the AbstractMaterial docstring you mention "to allow for temporal and spatial dependency" but I don't really follow tbh. The only example where it's actually used seems to be the viscosity of the lithosphere in the Schmallholz case, but there it's just referring to some extra attributes that have just been added on the fly, and really don't need to be a part of the class/object. In the end of the day you still end up with an expression that is always the same, so this could just be set as an attribute. If in some specific case you really want some code to "calculate" what the expression should be (and this calculation is not just a one off that you could have just done beforehand), you could always subclass Material, and change sub-class the attribute to become a property.

The other slight smell about AbstractMaterial is that it's defined in gadopt, but there's no associated code in gadopt. The only place we actually _use_ the materials is in the run_benchmark demo code. An abstract class is like a contract that defines what data and methods will be available but the meaning of that data is always intimately tied with the code that is written to handle it. So for instance you seem to assign specific meaning to B or RaB having the value None, but that logic lives in that bit of demo code.
Note I'm not so much concerned about your Simulation class and its class methods, that seems a kind of nice and reasonable way to avoid code duplication between the different cases, where you just stuff the specific bits in a class, and that's just the way you have chosen to set up the demos. My concern is more with the structure of AbstractMaterial which would be part of gadopt/ and also as a by-effect forces uses to actually know what a classmethod is (we have very intelligent users but not all of them are CS wizards).

gadopt/level_set_tools.py
```diff
@@ -0,0 +1,239 @@
...
+            if step % self.reini_params["frequency"] == 0:
+                for reini_step in range(self.reini_params["iterations"]):
+                    self.reini_ts.advance(
+                        0, update_forcings=self.update_level_set_gradient
+                    )
+
+                self.ls_ts.solution_old.assign(self.level_set)
+
+
+def sharp_interface(level_set, material_value, method):
```
stephankramer, line 201:
```suggestion
def sharp_interface(level_set, material_value, method):
    """docstring!"""
    sharp_interface_popped(level_set.copy(), material_value.copy(), method)
    
def sharp_interface_popped(level_set, material_value, method):
    """Construct UFL expression defining sharp interface by popping list of level_sets and material_values"""
```

stephankramer: I think it's better to be safe with these user-facing functions, and not eat the arguments provided by the user. Copies are trivially cheap anyway

gadopt/level_set_tools.py
```diff
@@ -0,0 +1,239 @@
...
+        return conditional(
+            ls > 0.5,
+            material_value.pop(),
+            sharp_interface(level_set, material_value, method),
+        )
+    else:  # Final level set; specify values for both sides of the interface
+        return conditional(ls < 0.5, *material_value)
+
+
+def diffuse_interface(level_set, material_value, method):
```
stephankramer, line 214: Same here

demos/multi_material/benchmarks/trim_2023.py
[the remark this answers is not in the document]
stephankramer: Intriguing indeed! Ok, we do need a comment with url/installation instructions then. If we decide this case needs to be included in the CI suite, then there's a few more things to think about.

stephankramer (changes requested):
Pretty much there afaic, but it would be good to have a think about `AbstractMaterial`

The other thing we need to think about is what (subset/reduced versions? of) cases do we want to run as part of the CI.

stephankramer:
I'll try to have a closer look at the Trim et al case next week to see if I can spot anything...

Patol75:
> I'll try to have a closer look at the Trim et al case next week to see if I can spot anything...

I have been exchanging emails with Sean Trim in the past few days. Most issues are resolved, but I am not sure we will be able to perfectly match that benchmark. The reason is the fact that the provided internal heating rate is exact, whilst the composition and temperature fields are numerical solutions.

For the composition, the massive issue is that the interface must be diffused with quite a large transition (equivalent to a CLS epsilon value of 1/70, which can easily be an order of magnitude bigger than what is used for other benchmarks). Such a large transition can become problematic during DG advection, especially at high resolution. I was not able to find any clever way to advect a sharp-transition level set while keeping the composition diffused. Indeed, floating-point arithmetic erases the memory of the signed distance function after the hyperbolic tangent is applied, and it does not seem possible to convert the sharp profile into a diffuse profile after advection. Do you see a way?

For the temperature, any small variations from the analytical solution will immediately result in oscillations generated through the loss of synchronicity with the internal heating rate. When running at high resolution, I notice that the temperature loses the synchronicity much faster than the composition, especially towards domain boundaries. Is there anything we could do to improve temperature advection accuracy in the energy solver?

demos/multi_material/benchmarks/crameri_2012.py
Patol75: I agree with the current structure not being ideal, but I fail to see how a `dataclass` could improve things. I can see two issues here. First, it would be nice to be able to provide an integer, float, or tuple if the value of a material property is constant. However, we also have to account for the fact that the material property could be `Callable`. For example, viscosity may depend on velocity and temperature, which evolve from one timestep to another, and thus it needs to be evaluated at each timestep. Indeed, this involves a function call, and objects of `float` type, for example, do not define `__call__`. Say we enforce material properties as callables. Then how different is it to subclass an `ABC` from writing functions and providing them to a `dataclass`? In the current format, all material properties are methods of a class, and each class corresponds to a material, so there is a sense of structure. Happy to have more thoughts.

Patol75: How I imagine it could look like as a `dataclass`.
```python
@dataclass(kw_only=True)
class Material:
    """Dataclass to specify relevant material properties that affect simulation's
    evolution. Each material property is provided as a class attribute. All physical
    properties should be dimensional."""

    B: int | float | Callable | None = None  # Buoyancy number
    RaB: int | float | Callable | None = None  # Compositional Rayleigh number
    density: int | float | Callable | None = None
    viscosity: int | float | Callable | None = 1  # Dynamic viscosity
    thermal_expansion: int | float | Callable | None = 1  # Volumetric expansion
    thermal_conductivity: int | float | Callable | None = 1
    specific_heat_capacity: int | float | Callable | None = 1  # At constant pressure
    internal_heating_rate: int | float | Callable | None = 0  # Per unit mass

    def __post_init__(self):
        self.thermal_diffusivity = (
            self.thermal_conductivity / self.density / self.specific_heat_capacity
        )
```

Patol75: How would you implement the lithospheric viscosity from the Schmalholz benchmark here?

Patol75:
trim_2023
This is the kind of evolution I now get for the Trim et al. (2023) benchmark (t = 7.7e-3). Background colours represent the analytical internal heating rate. Blue and green contours represent the 0.5 temperature isocontour and 0.5 CLS isocontour (material interface). I believe the CLS instabilities come from the excessive value of epsilon at high resolution (512x512). For the temperature, instabilities could be related to the energy solver advection scheme or could come as a consequence of numerical errors in CLS advection that get amplified by the loss of synchronicity with the internal heating rate.

Patol75:
@stephankramer My last push does not address all of your requests yet, as I am still working on the Trim benchmark. On the latter, I would like your input on the issue I describe below.

I had another email exchange with Sean Trim, and he wondered if I was updating the internal heating rate within the time-stepping algorithm. Indeed, I was not, so I tried to implement it, and the result is embedded in the last commit above. I was hoping for a noticeable improvement in the temperature solution, but instead, I got a significantly worse outcome, in the sense that the loss of synchronicity between temperature and internal heating rate is amplified. I do not quite understand how this is possible. If I modify line 221 of `demos/multi_material/run_benchmark.py` by removing any argument to `energy_solver.solve()`, I recover the previous behaviour, which yields a more accurate solution. Any thoughts much appreciated.

demos/multi_material/benchmarks/crameri_2012.py
angus-g: Since you've effectively defined an implicit contract where dynamic viscosity is _always_ provided a velocity argument, you could provide that as an argument.

I've provided a demo of this: https://github.com/g-adopt/g-adopt/commit/b9127724bd979ef67a4acee2c05e2e8fab28cb36. Because none of the other properties take an argument, they can be turned into a function as a `@property`, in an overriding class.

Patol75: Thanks for the suggestion! I have tried to combine elements from Steph's suggestion, my example above, and your example. Let me know what you think of what I just pushed.

stephankramer:
That's really weird indeed. So to clarify: are you commenting out the update_forcings argument *and* the explicit call to simulation.internal_heating_rate() or both? In any case, these individually should be very similar, it's just that the update_forcings will evaluate at t=t0+delta_t, whereas the explicit call beforehand does it at t=t0 (t0 being the beginning of timestep). Have you tried outputting and looking at the evaluated internal_heating at every timestep? I can't immediately find anything that's in contradiction between how the equations are defined in the paper and what we do....

There's something a bit funny btw about how the energy source is scaled - also in master, you just did the same for Boussinesq as was done for EBA - but that should only affect things if the reference density is not 1, so no the Trim case: with BA and EBA, we assume we have divided the energy equation by rho*c_p, by setting the prefactor rhocp used in the mass and advection term to 1, and redefining conductivity to be diffusivity. But that means if we provide *volumetric* heat rate (as we say in the doc string), it should be divided by rho*c_p to get the source term in the temperature equation. If it is actually a specific heat rate (as the rho*H term suggests) then actually the source should be that specific heat rate divided by c_p, except we don't have a mechanism to specify cp at the moment. For ALA and TALA (where don't divide by rhocp) it therefore seems that what we are expected to provide is the specific heat rate (despite what the doc string says).

demos/multi_material/benchmarks/schmalholz_2011.py
```diff
@@ -1,63 +1,27 @@
...
-        return None
 
-    @classmethod
-    def density(cls):
-        return 3300
+@dataclass
+class Lithosphere(ga.Material):
+    visc_coeff: ClassVar = 4.75e11
+    stress_exponent: ClassVar = 4
+    visc_bounds: ClassVar = (1e21, 1e25)
```
angus-g, line 20: I think `ClassVar` needs to be instantiated with an inner type

Patol75: It seems that it is running without them, but I will still add them nevertheless :)

Patol75:
Given the recent effort towards matching published diagnostics, I believe I will make any further changes to the source code supporting the level set functionality. Happy to have another round of comments on the code and opinions on how to set up meaningful tests for the CI.

Patol75:
There seems to be a failure with the Spiegelman cases. After doing a clean installation of Firedrake and G-ADOPT, I tried to reproduce the error locally, but with no success (see below). Any idea?

```bash
(firedrake) thomas@Legion-5i-Pro:~/Software/g-adopt_fork/demos/Drucker-Prager_rheology$ make all
echo "running spiegelman 0.0025_1e+23_64_16_50_False" >&2
running spiegelman 0.0025_1e+23_64_16_50_False
/usr/bin/time --format="spiegelman 0.0025_1e+23_64_16_50_False finished in %E" tsp -f python3 spiegelman.py 0.0025_1e+23_64_16_50_False
0
spiegelman 0.0025_1e+23_64_16_50_False finished in 0:48.48
echo "running spiegelman 0.0025_1e+23_64_16_50_True" >&2
running spiegelman 0.0025_1e+23_64_16_50_True
/usr/bin/time --format="spiegelman 0.0025_1e+23_64_16_50_True finished in %E" tsp -f python3 spiegelman.py 0.0025_1e+23_64_16_50_True
1
spiegelman 0.0025_1e+23_64_16_50_True finished in 0:46.77
echo "running spiegelman 0.0025_1e+23_64_16_0_False" >&2
running spiegelman 0.0025_1e+23_64_16_0_False
/usr/bin/time --format="spiegelman 0.0025_1e+23_64_16_0_False finished in %E" tsp -f python3 spiegelman.py 0.0025_1e+23_64_16_0_False
2
spiegelman 0.0025_1e+23_64_16_0_False finished in 0:22.52
echo "running spiegelman 0.0025_1e+23_64_16_0_True" >&2
running spiegelman 0.0025_1e+23_64_16_0_True
/usr/bin/time --format="spiegelman 0.0025_1e+23_64_16_0_True finished in %E" tsp -f python3 spiegelman.py 0.0025_1e+23_64_16_0_True
3
spiegelman 0.0025_1e+23_64_16_0_True finished in 0:23.87
echo "running spiegelman 0.0025_1e+23_64_16_5_False" >&2
running spiegelman 0.0025_1e+23_64_16_5_False
/usr/bin/time --format="spiegelman 0.0025_1e+23_64_16_5_False finished in %E" tsp -f python3 spiegelman.py 0.0025_1e+23_64_16_5_False
4
spiegelman 0.0025_1e+23_64_16_5_False finished in 0:21.69
echo "running spiegelman 0.0025_1e+23_64_16_5_True" >&2
running spiegelman 0.0025_1e+23_64_16_5_True
/usr/bin/time --format="spiegelman 0.0025_1e+23_64_16_5_True finished in %E" tsp -f python3 spiegelman.py 0.0025_1e+23_64_16_5_True
5
spiegelman 0.0025_1e+23_64_16_5_True finished in 0:20.95
echo "running spiegelman 0.0025_1e+23_64_16_15_False" >&2
running spiegelman 0.0025_1e+23_64_16_15_False
/usr/bin/time --format="spiegelman 0.0025_1e+23_64_16_15_False finished in %E" tsp -f python3 spiegelman.py 0.0025_1e+23_64_16_15_False
6
spiegelman 0.0025_1e+23_64_16_15_False finished in 0:22.96
echo "running spiegelman 0.0025_1e+23_64_16_15_True" >&2
running spiegelman 0.0025_1e+23_64_16_15_True
/usr/bin/time --format="spiegelman 0.0025_1e+23_64_16_15_True finished in %E" tsp -f python3 spiegelman.py 0.0025_1e+23_64_16_15_True
7
spiegelman 0.0025_1e+23_64_16_15_True finished in 0:26.60
echo "running spiegelman 0.0025_1e+23_64_16_25_False" >&2
running spiegelman 0.0025_1e+23_64_16_25_False
/usr/bin/time --format="spiegelman 0.0025_1e+23_64_16_25_False finished in %E" tsp -f python3 spiegelman.py 0.0025_1e+23_64_16_25_False
8
spiegelman 0.0025_1e+23_64_16_25_False finished in 0:29.52
echo "running spiegelman 0.0025_1e+23_64_16_25_True" >&2
running spiegelman 0.0025_1e+23_64_16_25_True
/usr/bin/time --format="spiegelman 0.0025_1e+23_64_16_25_True finished in %E" tsp -f python3 spiegelman.py 0.0025_1e+23_64_16_25_True
9
spiegelman 0.0025_1e+23_64_16_25_True finished in 0:28.98
echo "running spiegelman 0.005_1e+24_64_16_50_False" >&2
running spiegelman 0.005_1e+24_64_16_50_False
/usr/bin/time --format="spiegelman 0.005_1e+24_64_16_50_False finished in %E" tsp -f python3 spiegelman.py 0.005_1e+24_64_16_50_False
10
spiegelman 0.005_1e+24_64_16_50_False finished in 0:43.88
echo "running spiegelman 0.005_1e+24_64_16_50_True" >&2
running spiegelman 0.005_1e+24_64_16_50_True
/usr/bin/time --format="spiegelman 0.005_1e+24_64_16_50_True finished in %E" tsp -f python3 spiegelman.py 0.005_1e+24_64_16_50_True
11
spiegelman 0.005_1e+24_64_16_50_True finished in 0:41.44
echo "running spiegelman 0.005_1e+24_64_16_0_False" >&2
running spiegelman 0.005_1e+24_64_16_0_False
/usr/bin/time --format="spiegelman 0.005_1e+24_64_16_0_False finished in %E" tsp -f python3 spiegelman.py 0.005_1e+24_64_16_0_False
12
spiegelman 0.005_1e+24_64_16_0_False finished in 0:29.21
echo "running spiegelman 0.005_1e+24_64_16_0_True" >&2
running spiegelman 0.005_1e+24_64_16_0_True
/usr/bin/time --format="spiegelman 0.005_1e+24_64_16_0_True finished in %E" tsp -f python3 spiegelman.py 0.005_1e+24_64_16_0_True
13
spiegelman 0.005_1e+24_64_16_0_True finished in 0:31.72
echo "running spiegelman 0.005_1e+24_64_16_5_False" >&2
running spiegelman 0.005_1e+24_64_16_5_False
/usr/bin/time --format="spiegelman 0.005_1e+24_64_16_5_False finished in %E" tsp -f python3 spiegelman.py 0.005_1e+24_64_16_5_False
14
spiegelman 0.005_1e+24_64_16_5_False finished in 0:36.29
echo "running spiegelman 0.005_1e+24_64_16_5_True" >&2
running spiegelman 0.005_1e+24_64_16_5_True
/usr/bin/time --format="spiegelman 0.005_1e+24_64_16_5_True finished in %E" tsp -f python3 spiegelman.py 0.005_1e+24_64_16_5_True
15
spiegelman 0.005_1e+24_64_16_5_True finished in 0:31.36
echo "running spiegelman 0.005_1e+24_64_16_15_False" >&2
running spiegelman 0.005_1e+24_64_16_15_False
/usr/bin/time --format="spiegelman 0.005_1e+24_64_16_15_False finished in %E" tsp -f python3 spiegelman.py 0.005_1e+24_64_16_15_False
16
spiegelman 0.005_1e+24_64_16_15_False finished in 0:35.44
echo "running spiegelman 0.005_1e+24_64_16_15_True" >&2
running spiegelman 0.005_1e+24_64_16_15_True
/usr/bin/time --format="spiegelman 0.005_1e+24_64_16_15_True finished in %E" tsp -f python3 spiegelman.py 0.005_1e+24_64_16_15_True
17
spiegelman 0.005_1e+24_64_16_15_True finished in 0:30.96
echo "running spiegelman 0.005_1e+24_64_16_25_False" >&2
running spiegelman 0.005_1e+24_64_16_25_False
/usr/bin/time --format="spiegelman 0.005_1e+24_64_16_25_False finished in %E" tsp -f python3 spiegelman.py 0.005_1e+24_64_16_25_False
18
spiegelman 0.005_1e+24_64_16_25_False finished in 0:38.47
echo "running spiegelman 0.005_1e+24_64_16_25_True" >&2
running spiegelman 0.005_1e+24_64_16_25_True
/usr/bin/time --format="spiegelman 0.005_1e+24_64_16_25_True finished in %E" tsp -f python3 spiegelman.py 0.005_1e+24_64_16_25_True
19
spiegelman 0.005_1e+24_64_16_25_True finished in 0:45.80
(firedrake) thomas@Legion-5i-Pro:~/Software/g-adopt_fork/demos/Drucker-Prager_rheology$ python3 -m pytest test_spiegelman.py 
=============================================================================================================================== test session starts ===============================================================================================================================
platform linux -- Python 3.10.12, pytest-8.0.2, pluggy-1.4.0
rootdir: /home/thomas/Software/g-adopt_fork
configfile: pytest.ini
plugins: nbval-0.10.0, anyio-4.3.0, mpi-0.1, xdist-3.5.0
collected 2 items                                                                                                                                                                                                                                                                 

test_spiegelman.py ..                                                                                                                                                                                                                                                       [100%]

================================================================================================================================ 2 passed in 0.08s ================================================================================================================================
```
```

File: gadopt/approximations.py [hunk 10]
```diff
@@ -74,20 +76,35 @@ class BoussinesqApproximation(BaseApproximation):
...
+        self.RaB = ensure_constant(RaB)
+        self.delta_rho = ensure_constant(delta_rho)
+        self.H = ensure_constant(H)
 
     def buoyancy(self, p, T):
-        return self.Ra * self.g * self.alpha * self.rho * T
+        return (
+            self.Ra * self.rho * self.alpha * (T - self.T0) * self.g
+            - self.RaB * self.delta_rho * self.g
+        )
```
angus-g, line 107: I've nailed the test failure on Spiegelman down to this section. I think because it has been modified to include subtractions, it goes from an expression that was identically 0 to something more like `-1 * [1.] * [0.] * [1.]` (or at least that's what it prints out for me). For some reason, this must very slightly change the order of operations or otherwise impact the accuracy of the result. I guess Kerguelen's hardware in particular is sensitive enough for this to show up, even though it passes elsewhere.

I don't know if @stephankramer knows more about the numerical implementation here to comment? I think it may end up being that we modify the expected result of this test if we verify that this is only a small relative difference, since there's not really another way around adding the extra terms to this expression.

Patol75: The one thing I do not get is that `spiegelman.py` uses `approximation = BoussinesqApproximation(0)` (line 202), so `buoyancy` necessarily returns 0, even in the updated expression. All the terms in `buoyancy` are `Constant`s except for `T`. That is confusing...

angus-g: Yes, all the terms are 0, but they're overloaded as `Constant` or whatever (some UFL subclass), so the entire expression is a `Sum` containing a few terms, which is different to when it *only* contains multiplications of terms (which reduces to identically 0). Yes, it's silly, but clearly it changes the assembly slightly!

angus-g:
Locally you need to run `tsp -S 8` first to indicate that there are 8 slots present, otherwise it defaults to 1 and it will never run.

Patol75:
Thank you! So testing passes locally for me, but it seems that running the benchmark fails on the CI, and it is not very verbose:
```
make[2]: *** [Makefile:9: gerya_2003] Error 1
make[2]: Leaving directory '/__w/g-adopt/g-adopt/demos/multi_material'
make[1]: *** [Makefile:14: multi_material] Error 2
make[1]: *** Waiting for unfinished jobs....
```
How do I debug that?

angus-g:
Look at the "output timing information" step: you never install shapely. I'd suggest adding it as a testing optional dependency in pyproject.toml

Patol75:
Thank you, Angus. That should do the trick. By the way, after some testing, it seems that the problem we see with Drucker-Prager which you tracked down to `approximations.py` comes from the fact that the expression now contains terms that evaluate to `Constant(0)`. If, instead, these terms are `0`, then the `test_spiegelman.py` succeeds.

Patol75:
@angus-g @drhodrid @sghelichkhani @stephankramer We have a green tick; happy to have your thoughts on what you would like to see materialise (!) before the PR gets merged.

drhodrid (commented):
I'm slowly working my way through this and will leave comments sporadically through the day. 

First thing - as your demo includes a lot of very nice multi-material cases, of gradually increasing complexity, I think it's important to add a README to that demos/multi-material folder, with a brief description of the cases being run and how they build on each other. To avoid duplication on the webpage, perhaps we could have a consistent format for these README's across all demos, such that out benchmark tab auto populates. @angus-g  - would you have any advice on this? 

I think its perhaps also worth adding short note on how to run the individual cases -- e.g. mpiexec -np 4 python3 run_benchmark.py benchmarks.tosi_2015.

drhodrid:
The other thing I'm wondering is how we test this moving forward. Testing a selection of the most complex cases on a regular basis (i.e. with every pull request), is vital, but we also need to make sure all demos work and I think its intractable to run all cases for every pull request - they take a lot of time. With that in mind, can the testing be split into a "short" suite and a "long" suite? The long suite could be run on gadi, for example, in a similar manner to what we currently do with parallel scaling?

drhodrid:
When I run the Tosi case, as is, on Kerguelen, with a clean Firedrake (yesterday's build) and Level_Set branch, I fail with: 
`DIVERGED_MAX_ITfiredrake.exceptions.ConvergenceError: Nonlinear solve failed to converge after 50 nonlinear iterations.`

Any chance you could look into running this there @Patol75 to see if you do too?

drhodrid:
Running:
` mpiexec -np 4 python3 run_benchmark.py  benchmarks.tosi_2015` 
Seems to go fine (excluding the error I flagged above which you're looking into). 

However, running:
` mpiexec -np 4 python3 run_benchmark.py  benchmarks.crameri_2012` returns:
`FileNotFoundError: [Errno 2] No such file or directory: 'benchmarks/crameri_2012.msh'.

Can we get the runner setup to generate the meshes, such that this doesn't happen? It would be nice to have a consistent way of running them all (regardless of whether we are using in-built meshes or gmsh).

sghelichkhani:
> Running: ` mpiexec -np 4 python3 run_benchmark.py benchmarks.tosi_2015` Seems to go fine (excluding the error I flagged above which you're looking into).
> 
> However, running: ` mpiexec -np 4 python3 run_benchmark.py benchmarks.crameri_2012` returns: `FileNotFoundError: [Errno 2] No such file or directory: 'benchmarks/crameri_2012.msh'.
> 
> Can we get the runner setup to generate the meshes, such that this doesn't happen? It would be nice to have a consistent way of running them all (regardless of whether we are using in-built meshes or gmsh).

Yeah this is not ideal! In other demos where we generate mesh by gmesh (if we still have them, like in case of 2d annulus analytical case) we used Makefiles for that. 

A nice way to deal with this would be probably to do everything in python using gmsh package. For example the Schmalholz case would turn into:

```Python
import gmsh

def generate_mesh():
    vertical_resolution = 4e3
    domain_dimension_x = 1e6
    domain_dimension_y = 6.6e5
    layer_min_x = 4.2e5
    layer_thickness = 1.6e5
    layer_horizontal_resolution = 1e3
    
    # Initialize Gmsh
    gmsh.initialize()
    gmsh.model.add("mesh")

    # Define Points
    p1 = gmsh.model.geo.addPoint(0, 0, 0, vertical_resolution)
    p2 = gmsh.model.geo.addPoint(0, domain_dimension_y, 0, vertical_resolution)

    # Define Line
    l1 = gmsh.model.geo.addLine(p1, p2)

    # Extrusion 1
    gmsh.model.geo.extrude([(1, l1)], layer_min_x, 0, 0, numElements=[42], recombine=True)

    # Curve Tags update after Extrusion
    l2 = l1 + 1  # Update based on Gmsh's internal tagging system

    # Extrusion 2
    num_layers = int(layer_thickness / layer_horizontal_resolution)
    gmsh.model.geo.extrude([(1, l2)], layer_thickness, 0, 0, numElements=[num_layers], recombine=True)

    # Extrusion 3
    l6 = l2 + 4  # Updating to the new line generated after second extrusion
    gmsh.model.geo.extrude([(1, l6)], domain_dimension_x - layer_min_x - layer_thickness, 0, 0, numElements=[42], recombine=True)

    # Synchronize and Mesh
    gmsh.model.geo.synchronize()
    gmsh.model.mesh.generate(3)

    # Save and Finalize
    gmsh.write("output_mesh.msh")
    gmsh.finalize()

```

Patol75:
I think I have now addressed quite a few of your points @drhodrid and @sghelichkhani:
- Basic README is in.
- Stokes solver parameters are handled properly.
- Mesh generation is handled automatically; options are using a Firedrake meshing utility, providing a GMSH `.msh` file, providing a GMSH `.geo` file, or adding a method called `generate_mesh` to the Simulation class that uses the GMSH Python API.

Open issues currently include how to best test the new functionality and if the current approach with `run_benchmark.py` is appropriate.

gadopt/__init__.py
drhodrid, line 1: Is there a reason the diff is so huge for this file? It looks like changes are only minor yet the diff has lines changing places all over. If there is no need for these changes, revert and just include the relevant ones?

gadopt/stokes_integrators.py
```diff
@@ -122,35 +123,38 @@ class StokesSolver:
...
 
     def __init__(
         self,
         z: fd.Function,
         T: fd.Function,
         approximation: BaseApproximation,
-        bcs: Optional[dict[int, dict[str, int | float]]] = None,
-        mu: fd.Function | int | float = 1,
+        bcs: Optional[dict[int, dict[str, Number]]] = None,
+        mu: fd.Function | Number = 1,
```
drhodrid, line 143: Opening up the solver options to a string definition is nice. But perhaps that should be a seperate pull request? Or am I missing something?

pyproject.toml
```diff
@@ -17,7 +17,7 @@ classifiers = [
 ]
 
 [project.optional-dependencies]
-testing = ["assess", "pandas"]
+testing = ["assess", "pandas", "shapely"]
```
drhodrid, line 20: Do we need to add this dependency to gadopt in such a way that those doing --install gadopt after this merge automatically get shapely? This maybe a dumb point, but flagging in case.

gadopt/time_stepper.py
```diff
@@ -496,7 +822,7 @@ class ImplicitMidpointAbstract(AbstractRKScheme):
 
 class CrankNicolsonAbstract(AbstractRKScheme):
     """
-    Crack-Nicolson scheme
+    Crank-Nicolson scheme
```
drhodrid, line 825: I prefer a goog Crack scheme!

gadopt/time_stepper.py
```diff
@@ -67,7 +67,7 @@ def __init__(
         self.test = firedrake.TestFunction(solution.function_space())
         self.solution = solution
         self.fields = fields
-        self.dt = dt
+        self.dt = float(dt)
```
drhodrid, line 70: Why was this necessary? How haven't we hit an issue with this before?

gadopt/energy_solver.py
```diff
@@ -139,9 +139,8 @@ def setup_solver(self):
                                    solver_parameters=self.solver_parameters)
         self._solver_setup = True
 
-    def solve(self):
+    def solve(self, t=0, update_forcings=None):
```
drhodrid, line 142: Why is this t=0 option there? Just something Sia also mentioned, so important to have documented.

gadopt/approximations.py
```diff
@@ -166,13 +182,16 @@ def rhocp(self):
         return 1
 
     def kappa(self):
-        return self.thermal_diffusivity
+        return self.kappa_ref
```
drhodrid, line 185: What is this change for?

drhodrid (commented):
A few minor things from me and checks to make sure the changes included are all needed.

drhodrid:
So from my perspective, this largely looks good. I have flagged a few minor things for you to address/consider, but after that, I think @sghelichkhani and @stephankramer should take a look.

What this review has taught me is that we need a really consistent format for demos, tests and tutorials, and we don't currently have that - this is not your fault and should not really delay this merge, but it's something I think we need to carefully consider as our code base, user base and application base grows. I think the setup you have here is nice, in the sense that there is minimal code duplication, and it's a great way to test all of the level-set (and other) functionality, but I'm not sure I yet believe they work as "demos" per se (as we discussed today). 

That said - I don't think it should hold up a merge. I have a feeling that making demos and tests consistent across the board is likely to be the subject of another sprint in the future. 

Before merge, however, please ensure that you set up careful regression tests for cases with and without material dependent buoyancy. So perhaps testing one 2-material case, one 3-material case, and one thermo-chemical case, as a starting point (as we discussed today). For these, first ensure that at high resolution your results still match the benchmark values, then change across to a coarser resolution (and a shorter temporal period) for the regression testing, and set acceptable values that you have computed for these coarser cases. 

In summary - a huge effort @Patol75 and a massive contribution. Just some minor things to tidy up from me, and a subsequent approval from Sia and Steph.

gadopt/__init__.py
Patol75: In my opinion, the file looks much neater with alphabetically sorted imports, and it makes it easier and faster to find what you are looking for.

gadopt/stokes_integrators.py
Patol75: I tend to always add commits to the same PR; I just cannot help it. Up to you if you want it as a separate PR. :)

pyproject.toml
Patol75: I agree that having `shapely` under optional testing dependencies does not feel right. Either we make it a required dependency of `g-adopt` or we add an optional `multi-material` dependency alongside `testing` and `optimisation`.

gadopt/time_stepper.py
Patol75: Joke rejected.

gadopt/time_stepper.py
Patol75: It is not necessary. However, it seemed reasonable to me to ensure that `self.dt` is a `float` if we ensure that `self.dt_const` is a `Constant`.

gadopt/energy_solver.py
Patol75: It was always there, but not used. `t` is given to `update_forcings` and is needed for the Trim case, for example.

gadopt/approximations.py
Patol75: I am just reverting something I did in the documentation PR and do not like anymore. `thermal_diffusivity` just does not fit with how other attributes are named. Before, there was `_kappa`, which does not work either because it implies that the attribute is private, which I believe is not intended. I decided to put `kappa_ref` instead. It could be `kappa`, but in that case, the method's name needs to change.

Patol75:
All benchmarks included in `demos/multi_material/benchmarks` are now successfully tested as part of the CI. Judging from the Actions tab, these new tests added about 20 minutes of execution; let me know if this is reasonable. I will just remove the commented lines in the workflow file that I included in the previous commit; they were necessary to run it locally using `act`.

pyproject.toml
Patol75: Having thought a little bit more about it, I think it is fine the way it is.

.github/workflows/test.yml
```diff
@@ -54,13 +54,29 @@ jobs:
...
         run: |
           . /home/firedrake/firedrake/bin/activate
           python3 -m pip install .[testing,optimisation]
 
+      - name: Check out repository linked to Trim et al. (2023)
+        uses: actions/checkout@v4
+        with:
+          repository: seantrim/exact-thermochem-solution
+          path: ./flib
+
```
drhodrid, line 70: Just flagging this because I'm not sure what's right and wrong here. Do we really want to be checking out this repository every time? Would it not be easier (particularly for regression testing) to copy the current version to our test suite and make its original source clear?

demos/multi_material/README.md
```diff
@@ -0,0 +1,47 @@
...
+- Compositional and thermal buoyancy
+- Manufactured solution
+
+**benchmarks.van_keken_1997_isothermal**
+Isothermal Rayleigh-Taylor instability.
+- Compositional buoyancy
+
+**benchmarks.van_keken_1997_thermochemical**
+Entrainment of a thin, compositionally dense layer by thermal convection.
+- Compositional and thermal buoyancy
```
drhodrid, line 47: Nice!

demos/multi_material/benchmarks/__init__.py
drhodrid, line 1: Empty file?

demos/multi_material/benchmarks/tosi_2015.py
drhodrid, line 1: For clarity, could we perhaps call this multi_material_tosi or similar? My concern is that Tosi did not undertake any multi-material simulations, to this may cause confusion for some.

demos/multi_material/test_multi_material.py
drhodrid, line 1: Looks like a number of tests have been added here. Well done.

gadopt/stokes_integrators.py
drhodrid: Please seperate this solver option change out to a seperate PR.

drhodrid (commented):
This now is good from my perspective. Two things needed before merge:
1. Seperate out solver changes (strings) to seperate PR.
2. A last once over from @stephankramer.

.github/workflows/test.yml
Patol75: I do not have an answer to that, sadly. Could you clarify what you mean by easier?

demos/multi_material/benchmarks/tosi_2015.py
Patol75: Given we now have a README, I would not worry too much.

gadopt/diagnostics.py
```diff
@@ -11,20 +14,21 @@ class GeodynamicalDiagnostics:
       u:         Firedrake function for the velocity
       p:         Firedrake function for the pressure
       T:         Firedrake function for the temperature
-      bottom_id: bottom boundary identifier.
-      top_id:    top boundary identifier.
-      degree:    degree of the polynomial approximation.
+      bottom_id: bottom boundary identifier
+      top_id:    top boundary identifier
+      diag_vars: dictionary containing additional diagnostic-related functions
```
stephankramer, line 19: This is very vague: the user will have no idea what to put here.

gadopt/diagnostics.py
```diff
@@ -35,37 +39,54 @@ def __init__(
...
+        mesh_coords = SpatialCoordinate(level_set.function_space().mesh())
+        target_region = mesh_coords[1] >= entrainment_height
+        material_entrained = conditional(level_set < 0.5, 1, 0)
 
-    def T_avg(self) -> float:
-        return assemble(self.T * self.dx) / self.domain_volume()
+        return (
+            assemble(conditional(target_region, material_entrained, 0) * self.dx)
+            / material_area
+        )
```
stephankramer, line 92: Yeah, I would just make this a separate function. Tbh I can't quite remember why we thought combining these other basic diagnostics in one class was a good idea, maybe the idea was to have it as a diagnostic _file_ class originally? But in any case this one in particular seems quite separated from the class as a whole

setup.cfg
```diff
@@ -7,10 +7,11 @@
 # - E226 (missing whitespace around arithmetic operator)
 # - E741 (ambiguous variable name)
 # - E701 (multiple statements on one line (colon))
+# - W503 (line break before binary operator)
 # - W504 (line break after binary operator)
 
 ignore =
-    E501,F403,F405,E226,E741,E701,W504
+    E501,F403,F405,E226,E741,E701,W503,W504
```
stephankramer, line 14: Agree, this one's annoying - but in the future please do this in a separate PR.

stephankramer (changes requested):
Alright then, some final, minor requests.
Amazing piece of work!

.github/workflows/test.yml
Patol75: Something to note here is that I believe this repository only exists to support the associated publication. Essentially, it can be seen as static.

gadopt/approximations.py
```diff
@@ -166,13 +182,16 @@ def rhocp(self):
...
 
     def linearized_energy_sink(self, u):
         return 0
 
     def energy_source(self, u):
-        return 0
+        return self.rho * self.H
+
+    def density(self, p, T):
```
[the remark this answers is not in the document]
Patol75: I introduced `density` with @w3168 when we looked at the drunken sailor benchmark. I believe it is a convenient way to directly query density from within `stokes_integrators.py`. It is called `rho_field` in gh-51; have a look [here](https://github.com/g-adopt/g-adopt/blob/cd372506610e559b4715c5e230b4f0cd3e5b6250/gadopt/approximations.py#L167) and [here](https://github.com/g-adopt/g-adopt/blob/cd372506610e559b4715c5e230b4f0cd3e5b6250/gadopt/stokes_integrators.py#L236). The reason I am putting it here is that it is more general than what is in gh-51. Would you advise I keep it and document it or I remove it and we will figure it out in gh-51?

gadopt/diagnostics.py
Patol75: Well, I find it pretty neat that you can just instantiate a class by providing velocity, pressure, temperature, and boundary IDs and then just query diagnostics as methods. I think it is a good idea. Regarding `entrainment`, it is a very common diagnostic in multi-material geodynamical simulations. Given the class is called `GeodynamicalDiagnostics`, perhaps it should be included as a method. But it indeed requires arguments, whereas other existing methods do not. So it is not as straightforward as the other ones. @drhodrid @sghelichkhani Any opinions?

gadopt/diagnostics.py
Patol75: Agreed I should document it better. Note that it depends on what is decided below.

drhodrid:
I think for now my advice would be to keep generic single material diagnostics in the gd.diagnostics and move the entrainment elsewhere…

We can perhaps then think about a nicer longer term strategy for diagnostics at a later date...

On 24 May 2024, at 12:13 PM, Thomas Duvernay ***@***.***> wrote:



@Patol75 commented on this pull request.

________________________________

In gadopt/diagnostics.py<https://github.com/g-adopt/g-adopt/pull/35#discussion_r1612602665>:

> +    def entrainment(self, level_set_index, material_area, entrainment_height):
+        level_set = self.diag_vars["level_set"][level_set_index]

-    def Nu_bottom(self) -> float:
-        return assemble(dot(grad(self.T), self.n) * self.ds_b) * (1./assemble(Constant(1) * self.ds_b))
+        mesh_coords = SpatialCoordinate(level_set.function_space().mesh())
+        target_region = mesh_coords[1] >= entrainment_height
+        material_entrained = conditional(level_set < 0.5, 1, 0)

-    def T_avg(self) -> float:
-        return assemble(self.T * self.dx) / self.domain_volume()
+        return (
+            assemble(conditional(target_region, material_entrained, 0) * self.dx)
+            / material_area
+        )


Well, I find it pretty neat that you can just instantiate a class by providing velocity, pressure, temperature, and boundary IDs and then just query diagnostics as methods. I think it is a good idea. Regarding entrainment, it is a very common diagnostic in multi-material geodynamical simulations. Given the class is called GeodynamicalDiagnostics, perhaps it should be included as a method. But it indeed requires arguments, whereas other existing methods do not. So it is not as straightforward as the other ones. @drhodrid<https://github.com/drhodrid> @sghelichkhani<https://github.com/sghelichkhani> Any opinions?

gadopt/approximations.py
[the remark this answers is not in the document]
Patol75: @w3168 Given what @stephankramer wrote above about equation terms, perhaps it is not the most logical thing to add a density method to approximations. In that case, the method content could just be pasted directly in `StokesSolver` where you need it for the free surface. What do you think?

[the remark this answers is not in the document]
stephankramer: If it's not actually used in this PR, then indeed it shouldn't be included in this PR. My main objection was that it was a little unclear to me what this specific definition of density means (there's always various $\rho$s, $\bar\rho$, $\rho^\ast$, $\rho'$, $\Delta\rho$ etc. floating around); I guess this is what you would call "full density" under the assumption that buoyancy=-g(full density-reference density) - which I guess makes sense for most (all?) our applications - this split is already a little arbtitrary for multi-material as you know. So yeah, I don't mind it being part of approximation, but let's do that in #51 then and as long as it's documented.

Patol75:
Provided that CI passes, this should now be good to go.

drhodrid (approved):
What a milestone! Well done @Patol75 - a huge and important body of work here. Once CI is green, please go ahead and merge, and give yourself a big fat pat on the back!! :-)

Patol75:
@stephankramer I think it may need approval from your side as well.

gadopt/diagnostics.py
```diff
@@ -1,7 +1,10 @@
-import firedrake
-from firedrake import assemble, Constant, Function, sqrt, dot, grad, FacetNormal, DirichletBC
+from firedrake import (
+    Constant, FacetNormal, Function,
```
angus-g, line 2: Needs `DirichletBC` (for `ux_max`)

gadopt/level_set_tools.py
```diff
@@ -0,0 +1,529 @@
...
+        )
+        self.level_set_grad_proj = fd.Function(
+            self.func_space_lsgp,
+            name=f"Level-set gradient (projection) #{level_set.name()[-1]}",
+        )
+        self.proj_solver = self.gradient_L2_proj()
+
+        self.reini_params = (
+            reini_params_default if reini_params is None else reini_params
+        )
```
angus-g, line 236: Bit less verbose:
```suggestion
        self.reini_params = reini_params or reini_params_default
```

PR #36, opened 2023-10-12
Change GAMG default to use mis_k_minimum_degree_ordering

drhodrid:
Do as the title says! I have found with a number of test cases that this improves solver behaviour at all levels (reduces the number of inner solve iterations). Would like to run through test suite as an independent change, to ensure I didn't miss anything, prior to merging.


drhodrid:
This looks good to go - just needs a kind reviewer :-)

stephankramer (approved):
Amazing!

PR #38, opened 2023-10-16
Create workflow summaries for test status and timings

angus-g:
These now show up on the workflow summary page, rather than needing to go into the run log itself.


stephankramer (approved):
Very nice!

PR #39, opened 2023-10-23
Change default restored checkpoint in optimisation to the last saved

rubyturner:
This PR changes the default iteration for the restored checkpoint to be the last stored checkpoint instead of requiring a given iteration number to restore from.

Note that if a user adds other subdirectories to `checkpoint_dir` then this might/will break as it assumes the only things in there are the subdirectories for the iterations' checkpoints. I think this is a non-issue as we probably don't need to allow for that?


angus-g (approved):
Awesome, thanks @rubyturner! This looks good to me.

sghelichkhani (approved):
Looking! Thanks Ruby!

PR #41, opened 2023-10-27
Add parallel scaling to longtest framework

angus-g:
This now executes and tests on Gadi, as with the spherical analytical cases. I've also taken the liberty of updating the plotting script and hopefully collecting the data for the plots in a synchronised way with the testing. Because we're not regularly running the long tests, here's a manually triggered, successful run off this branch: https://github.com/g-adopt/g-adopt/actions/runs/6660055066/job/18100425186


drhodrid (approved):
This looks great @angus-g ! Much neater than what I committed. I think it's good to go.

Building on my comment from the previous PR (analytical tests) it would be good to document how these cases are run, but that can be another PR at a later date. 

Thanks for your huge efforts here!

sghelichkhani (approved):

PR #42, opened 2023-10-29
Remove pc_sor_diagonal_shift option, by default.

drhodrid:
We found with parallel scaling tests that this option led to poorer performance in the velocity solve. In previous versions of PETSc, some of our tests required this option to ensure velocity convergence. However, with the recent merge to PETSc upstream, alongside changes within G-ADOPT itself, that is no longer the case - we have a green CI. 

Given that we never really liked this option, this PR removes it. 

Note that if this is merged post parallel scaling tests, we will need to remove the dictionary popping of this option in all parallel scaling demos.


angus-g (changes requested):
Maybe now that #41 is merged you can remove the pop line from the parallel scaling test too?

drhodrid:
Done.

angus-g (approved):
All good now.

PR #44, opened 2023-11-06
Update expected results for parallel scaling

angus-g:
Results due to #43 were too good! Also adds in the run summary that we use for the non-HPC tests.

Successful run: https://github.com/g-adopt/g-adopt/actions/runs/6764504756


stephankramer:
> Also adds in the run summary that we use for the non-HPC tests.

Sounds good - where can I see this?

angus-g:
I only added it _after_ the successful run (oops). I was hoping to test without having to actually re-run the scaling, but I forgot that the rsync action wipes the logs, so the tests all fail! But you can see that here: https://github.com/g-adopt/g-adopt/actions/runs/6764993163

angus-g:
Successful run (with summary): https://github.com/g-adopt/g-adopt/actions/runs/6780491664

stephankramer (approved):
Nice one!

PR #45, opened 2023-11-07
Fix the iteration counter for restored optimiser

rubyturner:
This fixes the checkpointing iteration counter for restored optimisations.

Currently, the restored iteration gets saved again in the checkpointing directory as the next iteration. So after restoring from say iteration 10, this gets checkpointed as iteration 11 but is still iteration 10 as far as ROL is concerned. This means that with each restored run of the optimisation the directory names in the checkpoint directory and the actual ROL iteration values get more and more out of sync.

Please let me know if this was intended behaviour, but doesn't seem very user-friendly to me!

### DIFF
--- a/gadopt/inverse.py
+++ b/gadopt/inverse.py
@@ -215,6 +215,11 @@ class LinMoreOptimiser:
         self.rol_solver.rolvector.checkpoint_path = self.checkpoint_dir / "solution_checkpoint.h5"
         self.rol_solver.rolvector.load(self._mesh)
 
+        # ROL algorithms run in a loop like `while (statusTest()) { ... }`
+        # so we will double up on saving the restored iteration
+        # by rolling back the iteration counter, we make sure we overwrite the checkpoint
+        # we just restored, to keep the ROL iteration count, and our checkpoint iteration
+        # count in sync
         self.iteration -= 1
 
         # The various ROLVector objects can load all their metadata, but can't actually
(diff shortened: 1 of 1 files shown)


angus-g (approved):
Thanks Ruby, this is a good catch. The rest of this comment is just going to be my stream of consciousness as I think about it!

I am just wondering if it's us or ROL that's out of sync: the checkpointing is done during the status test, where we first increment the iteration count (which starts at -1) then write the checkpoint. Thus checkpoint 10 is written at the ~~end~~ start of the 11th iteration. When we restart from this checkpoint, we should effectively be at the start of the 12th iteration. The iteration counter should go up to 11 in this case.

Clearly this isn't happening, so is it because there's an extra status test when we first run the ROL algorithm that's checkpointing the *start* of the iteration? In that case, maybe it would be more elegant to skip checkpointing in the first status test after a restore, to avoid overwriting a checkpoint (I don't know how sacredly we're treating checkpoints!)

Looking at the `LinMoreAlgorithm::run` method, the optimisation loop is wrapped in
```cpp
while (status_->check(*state_)) {
  ...
}
```
Which does indeed indicate that we double up on the status check.

Of course, rolling back the iteration counter and overwriting the checkpoint makes sense too! Maybe this is the nicest solution, but we can supplement the line with a comment explaining why it's needed (just did that).

rubyturner:
Thanks for your comments Angus! Absolutely agree that overwriting the checkpoint is not the safest of solutions. Happy to modify if you think that's the better option in the long run.

PR #46, opened 2023-11-17
Conform with new UFL element interface

angus-g:
Unfortunately things got shuffled around and completely broke some of the bare element fiddling we do within the library.

There are still lots of warnings of the form:
```
/home/firedrake/firedrake/src/ufl/ufl/core/expr.py:269: DeprecationWarning: Expr.ufl_domain() is deprecated, please use extract_unique_domain(expr) instead.
```

We are guilty of calling that, but only probably once per run. The main culprits are coming from Firedrake itself, somewhere.


gadopt/utility.py
```diff
@@ -148,27 +149,15 @@ def is_continuous(expr):
...
-        return False
-    elif isinstance(elem, ufl.HCurlElement) or isinstance(elem, ufl.HDivElement):
-        return False
-    elif family == 'TensorProductElement':
-        return all(is_continuous(sele) for sele in elem.sub_elements())
-    elif family == 'EnrichedElement':
-        return all(is_continuous(e) for e in elem._elements)
-    else:
-        raise NotImplementedError("Unknown finite element family")
+    return elem in ufl.H1
```
angus-g, line 160: Well that's significantly less complicated! :grin:

angus-g (commented):
Thanks for taking a look at this @stephankramer, and adding the test is a good idea!

I think having a separate `test` (or `tests`?) directory for unit tests that aren't explicitly demos is probably the way to go. The setuptools config is already set up to ignore the top level directories that aren't `gadopt` as packages, so that's fine.

gadopt/utility.py
stephankramer: Yes, well thanks for finding that - I hadn't seen this Sobolev space logic before, pretty neat indeed.

stephankramer:
On second thought I prefer `tests/` indeed. Seems the various firedrake packages disagree on the issue:
```
$ ls -d */test*
fiat/test        libspatialindex/test  PyOP2/test           tsfc/tests
FInAT/test       loopy/test            thetis/test          ufl/test
firedrake/tests  pyadjoint/tests       thetis/test_adjoint
```

stephankramer:
Ready to go afaic. Thwaites is happy now as well: https://github.com/thwaitesproject/thwaites/pull/6

drhodrid:
This all looks good to me. A few minor questions:
1. How are the unit "tests" executed? Are they going to be a part of the test suite?  
2. Post (or pre) merge, can we run all of the long tests too, just to confirm there are no hidden gremlins?

angus-g:
Long test still passes (https://github.com/g-adopt/g-adopt/actions/runs/6941809152/job/18883402852), but you do need to make sure the Firedrake prefix on Gadi matches the g-adopt version you're running.

PR #48, opened 2023-12-19
Add pyGplates interface

sghelichkhani:
# PR Description

This PR is focused on integrating the often stand-alone `libgplates` into `gadopt`.

## Changes Made

1. **Elimination of Interpolation to Intermediary Grid Points**: 
   - Previously, interpolation to intermediary grid points was used to avoid velocities containing NaN values.
   - This issue primarily arose when there was inconsistency between the plate reconstruction model (often the topologies) and the rotation model, leading to scenarios where `pygplates` could not assign a plate ID for certain points at some geologic times.
   - After consultations with the EarthByte group, it's been determined that such discrepancies should not occur. Therefore, we have removed the need for this interpolation.

2. **Addition of a Tester**:
   - A new testing mechanism has been added.
   - This tester evaluates specific combinations of topology and rotation files over a given time period.
   - It is designed to raise an Exception if, for any of the time IDs, there is a point that cannot be associated with a plate ID.

## Dependency
- [x] This PR depends on [PR #85](https://github.com/g-adopt/g-adopt/pull/85) which introduces assignment of a 1d profile to a function.


angus-g:
Is there a demo case that goes along with this?

sghelichkhani:
> Is there a demo case that goes along with this?

Just added two demos. `test_minimal_gplates.py` can be used for testing.

drhodrid:
Hey @sghelichkhani - just about to look at this. Could you please just provide a bit of background? Is this essentially the test case that I had set up in gplates_demo, albeit with a new, improved, fantastic treatment for the coupling to pygplates?

sghelichkhani:
> Hey @sghelichkhani - just about to look at this. Could you please just provide a bit of background? Is this essentially the test case that I had set up in gplates_demo, albeit with a new, improved, fantastic treatment for the coupling to pygplates?

Yes! This is basically the same script that we published as part of our Davies et al GMD. The changes include moving gplates interface as part of our gadopt  package. I will enhance the docstring for `pyGplatesConnector` for our future documentation efforts.

drhodrid:
So this looks good to me. 

I will test it against the old version on a large run, over the coming days, to verify that everything works as expected, and report back.

We also need to get pygplates installed within our testing environment. @angus-g - can you look into this (it seems to be failing to find that right now)?

Finally, in addition to test_minimal_gplates.py, it's important that we check other diagnostics of the main run - now called forward.py (such as VRMS, Nu_top, Nu_bottom) and ensure the solvers pass. We only need to run this for 5 time-steps or so, but this case has a different null-space treatment to the other tests in our framework, and GAMG sometimes wobbles with this, so good to have that checked regularly.

drhodrid:
Just a short update here. Sia and I are trying to figure out when I am seeing some different behaviour in comparison to previous GPlates driven runs. This could be a few weeks, given new daddy duties!

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+        assigned a plate id. A warning is raised for any point not assigned a plate id.
+
+        Attributes:
+            rotation_filenames (str or list of str): Collection of rotation file names for pygplates.
+            topology_filenames (str or list of str): Collection of topology file names for pygplates.
+            geologic_zero (float): The oldest time present in the plate reconstruction model.
+            delta_time (float, optional): The time window range outside which plate velocities are updated. Defaults to 1.0.
+            scaling_factor (float, optional): Scaling factor for surface velocities. Defaults to 1.0.
+            nseeds (int, optional): Number of seed points used in the Fibonacci sphere generation. Defaults to 1000.
+            nneighbours (int, optional): Number of neighboring points to consider in velocity calculations. Defaults to 3.
```
angus-g, line 143:
```suggestion
        Arguments:
            rotation_filenames (Union[str, List[str]]): Collection of rotation file names for pygplates.
            topology_filenames (Union[str, List[str]]): Collection of topology file names for pygplates.
            geologic_zero (float): The oldest time present in the plate reconstruction model.
            delta_time (Optional[float]): The time window range outside which plate velocities are updated.
            scaling_factor (Optional[float]): Scaling factor for surface velocities.
            nseeds (Optional[int]): Number of seed points used in the Fibonacci sphere generation.
            nneighbours (Optional[int]): Number of neighboring points to consider in velocity calculations.
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+            for f in pygplates.FeatureCollection(fname):
+                self.topology_features.append(f)
+        # geologic_zero is the same as model_time=0
+        self.geologic_zero = geologic_zero
+
+        # time velocity interpolations
+        self.delta_time = delta_time
+
+        # seeds are equidistantial points generate on a sphere
+        self.seeds = self._fibonacci_sphere(samples=int(nseeds))
```
angus-g, line 173: You did already specify that `nseeds` should be an int.

```suggestion
        self.seeds = self._fibonacci_sphere(samples=nseeds)
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+            self._make_GPML_velocity_feature(
+                self.seeds
+            )
+        )
+
+        # last reconstruction time
+        self.reconstruction_time = None
+
+        # Flag to know when to recalculate surface velocities.
+        self.recalculation_flg = False
```
angus-g, line 187: I'd suggest a name that suggests what the flag is signalling, maybe `self.needs_recalculation` or similar?

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+
+        # Factor to scale plate velocities to RMS velocity of model,
+        # the definition: RMS_Earth / RMS_Model is used
+        # This is specially used in low-Rayleigh-number simulations
+        self.scaling_factor = scaling_factor
+
+    # setting the time that we are interested in
```
angus-g, line 194:
```suggestion
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+            nseeds (int, optional): Number of seed points used in the Fibonacci sphere generation. Defaults to 1000.
+            nneighbours (int, optional): Number of neighboring points to consider in velocity calculations. Defaults to 3.
+
+        Methods:
+            assign_plate_velocities(model_time):
+                Assigns plate velocities based on the specified model time.
+            ndtime2geotime(ndtime):
+                Converts non-dimensionalised time to geologic time.
+            geotime2ndtime(geotime):
+                Converts geologic time to non-dimensionalised time.
```
angus-g, line 151: These don't need to be here (at the class level) if there are docstrings for the methods themselves.

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+
+        # Factor to scale plate velocities to RMS velocity of model,
+        # the definition: RMS_Earth / RMS_Model is used
+        # This is specially used in low-Rayleigh-number simulations
+        self.scaling_factor = scaling_factor
+
+    # setting the time that we are interested in
+    def get_plate_velocities(self, target_coords, model_time):
+        """
+        Returns plate velocities for the specified target coordinates at the top boundary of a sphere,
```
angus-g, line 197:
```suggestion
        """Returns plate velocities for the specified target coordinates at the top boundary of a sphere,
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+    # Factor to dimensionalise model time: d^2/kappa
+    time_dim_factor = 2890e3**2/1.0e-6
+    # 1 yr in seconds
+    yrs2sec = 365*24*60*60
+    # 1 Myr in seconds
+    myrs2sec = 1e6*yrs2sec
+
+    def __init__(self, rotation_filenames, topology_filenames, geologic_zero, delta_time=1., scaling_factor=1.0, nseeds=1000, nneighbours=3):
+        """
+        An interface to pygplates, used for updating top Dirichlet boundary conditions
```
angus-g, line 126:
```suggestion
        """An interface to pygplates, used for updating top Dirichlet boundary conditions
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+
+        Notes:
+            - The method uses conversions between non-dimensionalised time and geologic time.
+            - Velocities are non-dimensionalised and scaled for the simulation before being applied.
+        """
+
+        # Raising an error if the user is asking for invalid time
+        if self.ndtime2geotime(ndtime=model_time) < 0:
+            raise Exception(
+                ("pyGplates: geologic time is being negative!"
```
angus-g, line 222: Maybe a `ValueError` rather than a more generic `Exception`? Also, I don't think this error should be attributed to pygplates, since it's specific to this connector code.

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+                 f"maximum: {self.geologic_zero/(pyGplatesConnector.time_dim_factor/pyGplatesConnector.myrs2sec/self.scaling_factor)}")
```
angus-g, line 223: Not sure it's clear what this is the maximum of? Wouldn't this be the minimum time before going negative?

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+        self.reconstruction_time = self.ndtime2geotime(model_time)
+        self.interpolated_u = self._interpolate_seeds_u(target_coords)
+        return self.interpolated_u
+
+    def ndtime2geotime(self, ndtime):
+        """
+        Converts non-dimensionalized time to geologic time with respect to present-day (Ma).
+
+        This function takes a non-dimensionalized time value and converts it to geologic time,
+        measured in millions of years ago (Ma), with respect to the present day.
```
angus-g, line 236: The additional description seems a bit redundant...

```suggestion
        """Converts non-dimensionalized time to geologic time with respect to present-day (Ma).
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+
+        Args:
+            ndtime (float): The non-dimensionalized time to be converted.
+
+        Returns:
+            float: The converted geologic time in millions of years ago (Ma).
+        """
+
+        geotime = self.geologic_zero - float(ndtime) * pyGplatesConnector.time_dim_factor/pyGplatesConnector.myrs2sec/self.scaling_factor
```
angus-g, line 245: You said `ndtime` should be a float in the function signature already? Yes, that's not enforced, but I think it could introduce subtle bugs if somebody ends up relying on the implicit conversion.

```suggestion
        geotime = self.geologic_zero - ndtime * pyGplatesConnector.time_dim_factor/pyGplatesConnector.myrs2sec/self.scaling_factor
```

Also, you end up using `pyGplatesConnector.time_dim_factor/pyGplatesConnector.myrs2sec/self.scaling_factor` a few times, maybe that should get its own constant?

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+        return geotime
+
+    def geotime2ndtime(self, geotime):
+        """ converts geologic time with respect to present-day (Ma) to non-dimensionalise time:
+
+        Args:
+            geotime (float): geologic time (before presentday in Myrs)
```
angus-g, line 252:
```suggestion
        """Converts geologic time with respect to present-day (Ma) to non-dimensionalized time.

        Args:
            geotime (float): geologic time (before present day in Myrs)
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+
+        Returns:
+            float: non-dimensionalised time
+         """
+        ndtime = (self.geologic_zero - geotime)*(pyGplatesConnector.myrs2sec*self.scaling_factor / pyGplatesConnector.time_dim_factor)
+        return ndtime
+
+    # convert seeds to Gplate features
```
angus-g, line 260:
```suggestion
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+    def _make_GPML_velocity_feature(self, coords):
+        """
+        Creates a pygplates.GPML velocity feature from specified coordinates.
```
angus-g, line 263:
```suggestion
        """Creates a pygplates.GPML velocity feature from specified coordinates.
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+        meshnode_feature.set_geometry(multi_point)
+        meshnode_feature.set_name('Velocity Mesh Nodes from pygplates')
+
+        output_feature_collection = pygplates.FeatureCollection(meshnode_feature)
+
+        return output_feature_collection
+
+    def _interpolate_seeds_u(self, target_coords):
+        """
+        Interpolates seed velocities onto target coordinates.
```
angus-g, line 295:
```suggestion
        """Interpolates seed velocities onto target coordinates.
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+        # calculate velocities here
+        seeds_u = self._calc_velocities(
+            velocity_domain_features=self.velocity_domain_features,
+            topology_features=self.topology_features,
+            rotation_model=self.rotation_model,
+            time=round(self.reconstruction_time, 0),
+            delta_time=self.delta_time)
+
+        seeds_u = np.array([i.to_xyz() for i in seeds_u]) *\
+            ((1e-2 * pyGplatesConnector.velocity_non_dim_factor) / (self.scaling_factor * pyGplatesConnector.yrs2sec))
```
angus-g, line 318: What's the magic `1e-2` in here? Should that get its own named constant?

```suggestion
        seeds_u = (
            np.array([i.to_xyz() for i in seeds_u]) *
            ((1e-2 * pyGplatesConnector.velocity_non_dim_factor) / (self.scaling_factor * pyGplatesConnector.yrs2sec))
        )
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+
+        # Then, calculate the weighted average using safe_dists
+        res_u = np.einsum(
+            'i, ij->ij',
+            1/np.sum(1/safe_dists, axis=1),
+            np.einsum('ij, ijk ->ik', 1/safe_dists, seeds_u[seeds_u[:, 0] == seeds_u[:, 0]][idx])
+        )
+
+        # Now handle the case where points are too close to each other:
+        res_u[dists[:, 0] <= 1e-8, :] = seeds_u[seeds_u[:, 0] == seeds_u[:, 0]][idx[dists[:, 0] <= 1e-8, 0]]
```
angus-g, line 339: `1e-8` here is a bit of a magic constant too.

I think this would be clearer if you assigned `dists[:,0] <= 1e-8` to its own variable.

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+
+        return res_u
+
+    def _calc_velocities(self, velocity_domain_features, topology_features, rotation_model, time, delta_time):
+        """
+        Calculates velocity vectors for domain points at a specific geological time.
```
angus-g, line 345:
```suggestion
        """Calculates velocity vectors for domain points at a specific geological time.
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+            velocity_domain_features: Domain features representing the points at which velocities are to be calculated.
+            topology_features: Topological features of tectonic plates used in the partitioning.
+            rotation_model: The rotation model defining plate movements over time.
+            time (float): The geological time at which velocities are to be calculated.
+            delta_time (float): The time interval over which velocity is calculated.
+
+        Returns:
+            list of pygplates.Vector3D: Velocity vectors for all domain points.
+
+        Raises:
```
angus-g, line 363:
```suggestion
        Warns:
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+                        # NB: velocity unit is fixed to cm/yr, but we convert it to m/yr and further on non-dimensionalise it later.
+                        velocity_vectors = pygplates.calculate_velocities(
+                            [velocity_domain_point],
+                            equivalent_stage_rotation,
+                            delta_time, velocity_units=pygplates.VelocityUnits.cms_per_yr)
+
+                        # add it to the list
+                        all_velocities.extend(velocity_vectors)
+                    else:
+                        warnings.warn("No plate id found. There is an issue with the reconstruciton model.", category=RuntimeWarning)
```
angus-g, line 405:
```suggestion
                        warnings.warn("No plate id found. There is an issue with the reconstruction model.", category=RuntimeWarning)
```

If this warning does end up getting raised and you return a NaN vector, is that going to cause a weird crash somewhere else? If so, maybe it's worth raising an exception directly?

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+                        all_velocities.extend([pygplates.Vector3D(np.NaN, np.NaN, np.NaN)])
+
+        return all_velocities
+
+    def _fibonacci_sphere(self, samples):
+        """
+        Generates points on a sphere using the Fibonacci sphere algorithm, which
```
angus-g, line 412:
```suggestion
        """Generates points on a sphere using the Fibonacci sphere algorithm, which
```

tests/test_gplates.py
```diff
@@ -0,0 +1,49 @@
...
+    # compute surface velocities
+    rec_model = pyGplatesConnector(
+        rotation_filenames=[
+            '../demos/gplates_global/gplates_files/Zahirovic2022_CombinedRotations_fixed_crossovers.rot',
+        ],
+        topology_filenames=[
+            '../demos/gplates_global/gplates_files/Zahirovic2022_PlateBoundaries.gpmlz',
+            '../demos/gplates_global/gplates_files/Zahirovic2022_ActiveDeformation.gpmlz',
+            '../demos/gplates_global/gplates_files/Zahirovic2022_InactiveDeformation.gpmlz',
+        ],
```
angus-g, line 30: Maybe define `gplates_files = Path(__file__).parent / "demos/gplates_global/gplates_files"` and then pull the files out of there? Just to be sure with relative paths and pytest.

demos/gplates_global/forward.py
```diff
@@ -0,0 +1,370 @@
...
+        # Log diagnostics:
+        plog.log_str(f"{num_timestep} {time} {float(dt)} "
+                     f"{gd.u_rms()} {gd.T_avg()}")
+    plog.close()
+
+
+def generate_mesh():
+    import os
+    if os.path.isfile("./simulation_state.h5"):
+        return
```
angus-g, line 224:
```suggestion
    from pathlib import Path
    if Path("simulation_state.h5").exists():
        return
```

demos/gplates_global/forward.py
```diff
@@ -0,0 +1,370 @@
...
+        extrusion_type="radial",
+    )
+
+    with CheckpointFile("simulation_states.h5", mode="w") as fi:
+        fi.save_mesh(mesh)
+
+
+def T_initialise(T, average):
+    import scipy
+    import math
```
angus-g, line 262: Generally, I don't like putting imports within methods, especially when they'll always be called. It makes it harder to see the dependencies of the file.

demos/gplates_global/forward.py
angus-g, line 1: Is there any intention to run this e.g. on the Gadi regression tests?

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+            topology_features=self.topology_features,
+            rotation_model=self.rotation_model,
+            time=round(self.reconstruction_time, 0),
+            delta_time=self.delta_time)
+
+        seeds_u = np.array([i.to_xyz() for i in seeds_u]) *\
+            ((1e-2 * pyGplatesConnector.velocity_non_dim_factor) / (self.scaling_factor * pyGplatesConnector.yrs2sec))
+
+        # generate a KD-tree of the seeds points that have a numerical value
+        tree = cKDTree(data=self.seeds[seeds_u[:, 0] == seeds_u[:, 0], :], leafsize=16)
```
angus-g, line 321: I don't understand the point of doing `seeds_u[:,0] == seeds_u[:,0]`, doesn't this just give an array of `True` that's as long as the first dimension? You end up using this a few times in the indexing too, maybe factor it out?

Any reason to specify `leafsize=16`? This is the default anyway, but maybe let scipy decide that?

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+        Raises:
+            RuntimeWarning: If no plate ID is found for a given point, indicating an issue with the reconstruction model.
+        """
+        # All domain points and associated (magnitude, azimuth, inclination) velocities for the current time.
+        all_domain_points = []
+        all_velocities = []
+
+        # Partition our velocity domain features into our topological plate polygons at the current 'time'.
+        # Note: pygplates can only deal with rounded ages in million years
+        plate_partitioner = pygplates.PlatePartitioner(topology_features, rotation_model, round(time, 0))
```
angus-g, line 372: Given you use `round(time, 0)` several times, you could factor that out.

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+            We use this method for generating seed points that can be interpolated onto our mesh
+
+        References:
+            The algorithm is based on the concept of the golden angle, derived from
+            the Fibonacci sequence and the golden ratio.
+        """
+
+        phi = np.pi * (3. - np.sqrt(5.))  # golden angle in radians
+
+        y = 1 - (np.array(list(range(samples)))/(samples-1)) * 2
```
angus-g, line 444:
```suggestion
        y = 1 - (np.arange(samples)/(samples-1)) * 2
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+        radius = np.sqrt(1 - y * y)
+        theta = phi * np.array(list(range(samples)))
```
angus-g, line 446:
```suggestion
        theta = phi * np.arange(samples)
```

gadopt/gplates.py
```diff
@@ -0,0 +1,449 @@
...
+        x = np.cos(theta) * radius
+        z = np.sin(theta) * radius
+        return np.array([[x[i], y[i], z[i]] for i in range(len(x))])
```
angus-g, line 449:
```suggestion
        return np.stack((x, y, z)).T
```

angus-g (commented):
Generally looks pretty good! A few formatting tweaks on the docstrings, and some minor code issues. I trust that you know what you're doing for the gplates bits!

I'll add gplates into the Docker environment so we can run the unit test.

demos/gplates_global/forward.py
sghelichkhani: Agreed! The only motivation was that `T_initialise` itself is called only when initialising. But now moved to beginning.

gadopt/gplates.py
sghelichkhani: Agreed.

gadopt/gplates.py
sghelichkhani: So this would be maximum non-dimensionalised time. Non-dimensionalised time starts from 0 which in geologic age is the oldest age available in pygplates files. The maximum non-dimensionalised time then equals the minimum geologic age, which is now, and is zero. 

Now I reliase I have to use the appropriate term and rename geologic_time to geologic_age.

angus-g: Oh right, everything is backwards! :upside_down_face:

gadopt/gplates.py
sghelichkhani: Correct! That is cm2m conversion.

tests/test_gplates.py
sghelichkhani: Makes sense! Thanks Agnus.

gadopt/gplates.py
sghelichkhani: When pygplates calculates the surface velocities, it should find the associated plate polygon for each point. For some geologic ages, the produced files include polygons that do not entirely cover the Erath's surface. These inevitably result in an empty plate id, which implies that no surface velocity can be calculated by gplates.

sghelichkhani: In the absense of any solution to deal with these points, we do interpolation between the all that actually has a solution. Here, instead of building arrays of coordinates and the velocities that are found for those coordinates, I keep the coordinates the same and assign Nan for points that don't have values. 
Then for building the tree, I only pass in seeds with `seeds_u[:,0] == seeds_u[:,0]`, which are essentially non Nan points.

gadopt/gplates.py
sghelichkhani: So this is basically to say, if we are extremely close to any point, let's not interpolate and assign the value itself. It is just to avoid non-sense results when dividing by zero distance. But a valid point re it being a magic constant. I define it as a constant now.

gadopt/gplates.py
sghelichkhani: Makes sense!

gadopt/gplates.py
sghelichkhani: See discussion above regarding interpolation.

gadopt/gplates.py
sghelichkhani: Beautiful!

demos/gplates_global/forward.py
sghelichkhani: Thanks!

gadopt/gplates.py
Patol75: It is probably more readable to use `numpy.isfinite` or `numpy.isnan`, if you do not mind.

.github/workflows/test.yml
```diff
@@ -54,6 +54,9 @@ jobs:
           sudo apt update
           sudo apt install -y task-spooler time
           tsp -S 48
+          curl https://www.earthbyte.org/download/9079/ -o ./pygplates_0.36.0_py310_ubuntu-22.04.deb
+          sudo apt install -y --fix-broken ./pygplates_0.36.0_py310_ubuntu-22.04.deb
+          ln -s /usr/lib/pygplates.so /home/firedrake/firedrake/lib/python3.10/site-packages/
       - uses: actions/checkout@v4
```
drhodrid, line 61: Just tagging @angus-g to check the above installation design!

demos/gplates_global/forward.py
```diff
@@ -0,0 +1,370 @@
...
+
+    element_family = q.function_space().ufl_element()
+    X = Function(VectorFunctionSpace(mesh=mesh, family=element_family)).interpolate(SpatialCoordinate(mesh))
+    rad = Function(q.function_space()).interpolate(sqrt(X**2))
+    averager = LayerAveraging(mesh, cartesian=False)
+    averager.extrapolate_layer_average(q, interp1d(rshl, visc, fill_value="extrapolate")(averager.get_layer_average(rad)))
+
+
+if __name__ == "__main__":
+    forward()
```
drhodrid, line 370: Please run this example, get a set of diagnostic values at the end of the simulation and then add it to the testing environment to ensure ongoing functionality.

demos/gplates_global/forward.py
```diff
@@ -0,0 +1,370 @@
...
+            )
+
+    # Initiating a plate reconstruction model
+    plate_receonstion_model = pyGplatesConnector(
+        rotation_filenames=[
+            './gplates_files/Zahirovic2022_CombinedRotations_fixed_crossovers.rot'],
+        topology_filenames=[
+            './gplates_files/Zahirovic2022_PlateBoundaries.gpmlz',
+            './gplates_files/Zahirovic2022_ActiveDeformation.gpmlz',
+            './gplates_files/Zahirovic2022_InactiveDeformation.gpmlz'],
```
drhodrid, line 82: Our of interest, instead of uploading these files to the repo, is there an option to grab them via some sort of wget command at runtime?

gadopt/gplates.py
drhodrid: Agreet with @Patol75 here. Please update.

gadopt/gplates.py
```diff
@@ -0,0 +1,444 @@
...
+    #   L [cm] X cm2m = L [m]
+    cm2m = 1e-2
+    #   u [cm/year] * velocity_dimDcmyr = u []
+    velocity_dimDcmyr = velocity_non_dim_factor * cm2m / yrs2sec
+
+    # minimum distance, bellow which we do not interpolate
+    #   this is just to avoid division by zero when weighted averaging
+    epsilon_distance = 1e-8
+
+    def __init__(self, rotation_filenames, topology_filenames, oldest_age, delta_t=1., scaling_factor=1.0, nseeds=1000, nneighbours=3):
```
drhodrid, line 136: Are you happy with your default parameter choice for nseeds and nneighbours here? When we tested this together around a month ago, we found this made a difference. For our previous GPlates simulations (in the 2022 paper and Tom's stuff), we have used  nseeds=1e5, nneighbours=4. This seems to lead to sharper velocity gradients at the plate boundaries. Some may say this is a better representation of rigid plate motions, but it also makes the Stokes solution process more challenging, so there is a balance here. I'm happy with these defaults if you are, but perhaps add a comment similar to what I've written here, so that we remember this discussion!

drhodrid (commented):
Aside from the minor things I have raised here, this is good to go. Please address at your earliest convenience, then I will be happy to approve.

.github/workflows/test.yml
angus-g: I think this is reasonable (quite close to what they do in the [upstream Dockerfile](https://github.com/GPlates/dockerfile/blob/main/Dockerfile.pygplates-py3-base-jammy-amd64)), although I don't think we need `--fix-broken`?

demos/gplates_global/forward.py
angus-g: If they do end up being required for the library, we could do something like xarray and split into a different repository: https://github.com/pydata/xarray-data. However, definitely not subprocessing out to wget at runtime! Use a library like https://pypi.org/project/pooch/ which handles downloading and caching at the Python level.

gadopt/gplates.py
sghelichkhani: Done

gadopt/gplates.py
sghelichkhani: Agreed! Thanks @Patol75

demos/gplates_global/forward.py
sghelichkhani: Well we have the legendary g-drift repo for that. For now I am thinking of simply downloading it from Earthbyte website itself, and do a test for it.

sghelichkhani:
@angus-g  I probably need to add pooch to dependencies?

demos/gplates_global/forward.py
sghelichkhani: Didn't we want to move this level of testings from now on to testing_long directory? @angus-g what do you say?

.github/workflows/test.yml
sghelichkhani: eeee ...  This is I think what I pulled from Angus' branch?  Do you need me to change this?

angus-g: Oh, weird! Past me must've put it there for a reason. Can't really hurt, so this is fine.

angus-g:
> @angus-g I probably need to add pooch to dependencies?

Yep.

.github/workflows/test.yml
angus-g: Although having said that, I might look into caching that file so we don't hit their server all the time...

demos/gplates_global/forward.py
angus-g: I can't actually run this as-is:
```
  File "/home/firedrake/g-adopt/demos/gplates_global/forward.py", line 370, in <module>
    forward()
  File "/home/firedrake/g-adopt/demos/gplates_global/forward.py", line 77, in forward
    muller_2022_files = obtain_Muller_2022_SE("./gplates_files")
  File "/home/firedrake/g-adopt/gadopt/gplatefiles.py", line 37, in obtain_Muller_2022_SE
    raise FileNotFoundError("Some files are missing and download_mode is set to False. Cannot proceed without downloading the required files.")
FileNotFoundError: Some files are missing and download_mode is set to False. Cannot proceed without downloading the required files.
```

Having said that, I think this is big enough that we only want it in the long test suite? If you give me an idea of core count and runtime, I can wire that up.

gadopt/gplates.py
```diff
@@ -0,0 +1,450 @@
...
+        # Use np.where to avoid division by very small values
+        # Replace tiny distances with pyGplatesConnector.epsilon_distance to avoid division
+        # by tiny values while keeping original distances intact for later use
+        safe_dists = np.where(dists < pyGplatesConnector.epsilon_distance, pyGplatesConnector.epsilon_distance, dists)
+
+        # Then, calculate the weighted average using safe_dists
+        res_u = np.einsum(
+            'i, ij->ij',
+            1/np.sum(1/safe_dists, axis=1),
+            np.einsum('ij, ijk ->ik', 1/safe_dists, seeds_u[non_nan_values][idx])
```
Patol75, line 335: Just out of curiosity, what is the mathematical operation represented here in the inner call to `einsum`?

demos/gplates_global/forward.py
sghelichkhani: Yeah! So part of the run case should be calling `obtain_Muller_2022_SE("./gplates_files", download_mode=True)` from `gadopt.gplatefiles`. And then calling `forward()`, noting that we call the `obtain_Muller_2022_SE("./gplates_files", download_mode=False)` on  https://github.com/g-adopt/g-adopt/blob/26ad2bd8c2ed47bd2106ba2096548207de79dd51/demos/gplates_global/forward.py#L77

This is basically a level 6 spherical case. So only in long suite. Core count should be the same as level 6 in other spherical case, with the same runtime (here we run it for only 20 timesteps). I will give more accurate numbers later.

angus-g: Thanks! That's enough for me to go on for now :)

gadopt/gplates.py
sghelichkhani: The inner einsum is multiplying the weighting associated with each point with all three velocity vectors.

sghelichkhani: "i" is the index for number of points, "j" is the index for n number of nearest points, and "k" is the number of coordinates which is 3. This multiplies all three components with the same weight and then sums all the weighted values togeth to give "ik"

Patol75: That is interesting. I gave it a go and indeed:
```python
>>> a = np.ones((2, 3))
>>> b = 2 * np.ones((2, 3, 4))
>>> np.einsum("ij,ijk->ik", a, b)
array([[6., 6., 6., 6.],
       [6., 6., 6., 6.]])
>>> np.sum(np.atleast_3d(a) * b, axis=1)
array([[6., 6., 6., 6.],
       [6., 6., 6., 6.]])
```
I may start using `einsum` more often!

sghelichkhani:
> > @angus-g I probably need to add pooch to dependencies?
> 
> Yep.

Done!

sghelichkhani:
I have now removed all the faff regarding downloading of rotation+polygon files from Earthbyte webpage. This was primarily needed because of our testing. The functionality for Muller et al 2022 now only checks for certain files of the reconstruction model to exists and returns their absolute path. Notice that I have now moved `assign_1d_function` to a seperate PR after discussion from @drhodrid.

drhodrid (changes requested):
This looks mainly good to go. My only concern is with the tutorial/demo. In the spirit of the sprint we just had, this really needs to be the spherical case that we have already, with minimal changes: i.e. what we want is to highlight the changes required to run a case with GPlates relative to one without. With that in mind, it should be incompressible, and doesn't need the variable mesh spacing. My advice would be that it:

1. Has a depth dependent viscosity, from a file - see comment on 1-D pull request.
2. Applies GPlates as a kinematic surface BC.
3. Updates nullspaces relative to free-slip case. 

And that's it...

sghelichkhani:
@drhodrid Once you give approval for this, this should be good to go.

drhodrid (commented):
I have made minor edits to the text of the demo, which I think improve clarity, but please confirm. In theory I'm now happy with this and it looks great - yay! Just a few final things to add:

1. This demo now puts a hard g-adopt requirement on pygplates - because it needs to be tested. At present I don't have that installed on my mac. How are we going to handle that requirement @angus-g ?
2. You'll need to update the demos map, to show an arrow from 3-D spherical --> this. 
3. Forward.py still has radii at 1.22 and 2.22. Did you want to change that? That case (and it's pkl file etc...) should now also be moved across to the new testing directory, since Angus has got that in.
4. Are you also going to update the radii in 3-D spherical.py for consistency?

angus-g:
> This demo now puts a hard g-adopt requirement on pygplates - because it needs to be tested. At present I don't have that installed on my mac. How are we going to handle that requirement

I'm not sure how those statements correlate - pygplates is installed in the testing environment. Additionally, you don't need to have pygplates installed unless you specifically import `gadopt.gplates`.

drhodrid:
Sorry for confusion. You addressed my concern though - pygplates is in the testing environment, which is what matters.

R

On 24 Jun 2024, at 7:23 PM, Angus Gibson ***@***.***> wrote:



This demo now puts a hard g-adopt requirement on pygplates - because it needs to be tested. At present I don't have that installed on my mac. How are we going to handle that requirement

I'm not sure how those statements correlate - pygplates is installed in the testing environment. Additionally, you don't need to have pygplates installed unless you specifically import gadopt.gplates.

sghelichkhani:
Addressed comments 2, 3, 4 of Rhodri's comments. See Angus' response for comment 1. God bless squash, or this PR would be a novel in commits. 😂

demos/.diagram.mermaid
```diff
@@ -10,8 +10,8 @@ graph TD
   cartesian_3d[3D Cartesian]
   end
 
-  subgraph geometry[Geometry]
-  cylindrical_2d[2D Cylindrical] --> spherical_3d[3D Spherical]
+  subgraph geometry["Geometry & Boundary Condition"]
+  cylindrical_2d[2D Cylindrical] --> spherical_3d[3D Spherical] --> gplates_global["Global Glates"]
```
angus-g:
```suggestion
  cylindrical_2d[2D Cylindrical] --> spherical_3d[3D Spherical] --> gplates_global["Global GPlates"]
```

demos/.diagram.mermaid
```diff
@@ -35,4 +35,5 @@ graph TD
   click cartesian_3d "3d_cartesian"
   click cylindrical_2d "2d_cylindrical"
   click isothermal "compositional_buoyancy"
-  click thermochemical "thermochemical_buoyancy"
\ No newline at end of file
+  click thermochemical "thermochemical_buoyancy"
+  click gplates_global"gplates_global"
```
angus-g, line 39:
```suggestion
  click gplates_global "gplates_global"
```

demos/gplates_global/gplates_global_demo.py
angus-g, line 1: This file should be called `gplates_global.py` to make it easier to slot into the demos Makefile -- this will also need to go into `demo_cases` in there.

demos/.diagram.mermaid
drhodrid, line 14: GPlates

drhodrid (commented):
Lovely - looks good but 1 minor typo. @angus-g - is this all configured to run through your testing, or do we need to update the list of demos somewhere too?

.github/workflows/test.yml
```diff
@@ -50,10 +50,18 @@ jobs:
...
           sudo apt update
-          sudo apt install -y task-spooler time
+          sudo apt install -y task-spooler time s3cmd unzip
           tsp -S 48
+          mkdir -p $HOME/.s3cfg
+          echo "${S3CONFIG}" > $HOME/.s3cfg/config
+          chmod 600 $HOME/.s3cfg/config
+          s3cmd --config=$HOME/.s3cfg/config get s3://gadopt/github-actions/pygplates_0.36.0_py310_ubuntu-22.04.deb
+          sudo apt install -y --fix-broken ./pygplates_0.36.0_py310_ubuntu-22.04.deb
+          ln -s /usr/lib/pygplates.so /home/firedrake/firedrake/lib/python3.10/site-packages/
```
angus-g, line 64: Given Firedrake has updated to 24.04, this will have to change (as you've probably seen from the failing runs) -- I'm building a new .deb now that should work for us.

.github/workflows/test.yml
```diff
@@ -62,6 +70,12 @@ jobs:
           . /home/firedrake/firedrake/bin/activate
           python3 -m pip install .[demos,optimisation]
 
+      - name: Get Muller et al 2022 plate reconstructions
+        run: |
+          s3cmd --config=$HOME/.s3cfg/config get s3://gadopt/github-actions/Muller_etal_2022_SE_1Ga_Opt_PlateMotionModel_v1.2.zip .
+          unzip Muller_etal_2022_SE_1Ga_Opt_PlateMotionModel_v1.2.zip -d demos/gplates_global
+        shell: bash
```
angus-g, line 77:
```suggestion
```

This shouldn't be needed, right?

angus-g (changes requested):

gadopt/gplates/gplates.py
```diff
@@ -0,0 +1,450 @@
...
+                        velocity_vectors = pygplates.calculate_velocities(
+                            [velocity_domain_point],
+                            equivalent_stage_rotation,
+                            delta_t, velocity_units=pygplates.VelocityUnits.cms_per_yr)
+
+                        # add it to the list
+                        all_velocities.extend(velocity_vectors)
+                    else:
+                        warnings.warn("Issues finding plate ids for some of the seeds.", category=RuntimeWarning)
+                        all_velocities.extend([pygplates.Vector3D(np.NaN, np.NaN, np.NaN)])
```
angus-g, line 408:
```suggestion
                        all_velocities.extend([pygplates.Vector3D(np.nan, np.nan, np.nan)])
```

.github/workflows/test.yml
sghelichkhani: Considering this is the only place I am having it, yes!

sghelichkhani (commented):
Fixed the issues pointed by Angus.

.github/workflows/test.yml
sghelichkhani: Thanks!

sghelichkhani:
Just a note for testing: This PR has a unit test for gplates functionality. But I do have a separate branch for gplates_testing that will introduce gplates run for long tests.

sghelichkhani (commented):
Added the demo to demo list. renamed demo to be consistent

demos/3d_spherical/3d_spherical.py
```diff
@@ -25,7 +25,7 @@
 # using the optional keyword argument *degree$=2$*.
 
 # +
-rmin, rmax, ref_level, nlayers = 1.22, 2.22, 4, 8
+rmin, rmax, ref_level, nlayers = 1.208, 2.208, 4, 8
```
angus-g, line 28: This is why this test is failing, probably? Probably need to update answers, if you've verified the new values are correct.

sghelichkhani: Thanks @angus-g for sorting this out. Sorry for being so slow with this.

sghelichkhani (commented):
Thank again Angus for being quick with the changes! Should be good to go once it goes green.

sghelichkhani:
Alright! Give those approvals!

demos/gplates_global/gplates_global.py
angus-g, line 1: I'm going to assume @drhodrid has read this and checked it all looks good :grinning: My final comment here: is this file tested? We should at least *run* it, but maybe it could get a check on the final velocity or something like that?

angus-g: Also, is pyGPlates ever listed as a dependency (not a hard dependency in pyproject.toml)? Should that go in the general install instructions on gadopt.org and/or in this tutorial?

sghelichkhani: Hahha! Rhodri went through this. 

I am not really sure. I did not know you can have a non-hard dependency. I could have a sentence, referring users to the Earthbyte webpage for pygplates.

angus-g: > I did not know you can have a non-hard dependency.

What I mean here is explicitly not putting it in the `pyproject.toml`, i.e. don't change anything! But we do probably need to tell people that pyGPlates is indeed required

sghelichkhani: I see! I agree with that. Let me quickly add a sentence to the top.

demos/gplates_global/gplates_global.py
```diff
@@ -0,0 +1,263 @@
...
+# -
+
+# ## Loading a 1-D Axisymmetric Profile for Viscosity
+# ----------
+# Earth's physical properties are primarily characterised by spherically symmetric (depth-dependent) features influenced predominantly by hydrostatic pressure variations.
+# Here, we load a 1-D viscosity profile, as utilised by Ghelichkhan et al. (2021) in *Geophysical Journal International* to model the Earth's evolution during the Cenozoic era.
+# We first set up our viscosity function space. The 1-D profile data is located in the file `./mu2_radial.rad`, and we will use interpolation functionalities provided by G-ADOPT to populate the viscosity field.
+
+mu = Function(Q, name="Viscosity")
+interpolate_1d_profile(function=mu, one_d_filename="mu2_radial.rad", cartesian=False)
```
angus-g, line 80: I tried running this, and get the following error:
```
  File "/home/firedrake/g-adopt/demos/gplates_global/gplates_global.py", line 80, in <module>
    interpolate_1d_profile(function=mu, one_d_filename="mu2_radial.rad", cartesian=False)
  File "/home/firedrake/firedrake/lib/python3.12/site-packages/gadopt/utility.py", line 514, in interpolate_1d_profile
    rad = Function(function.function_space()).interpolate(upward_coord)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/firedrake/firedrake/src/firedrake/firedrake/function.py", line 413, in interpolate
    return assemble(interp, tensor=self, ad_block_tag=ad_block_tag)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/firedrake/firedrake/src/firedrake/firedrake/adjoint_utils/assembly.py", line 30, in wrapper
    output = assemble(form, *args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/assemble.py", line 133, in assemble
    return get_assembler(expr, *args, **kwargs).assemble(tensor=tensor)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/assemble.py", line 381, in assemble
    result = BaseFormAssembler.base_form_postorder_traversal(self._form, visitor, visited)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/assemble.py", line 603, in base_form_postorder_traversal
    visited[e] = visitor(e, *(visited[arg] for arg in operands))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/assemble.py", line 377, in visitor
    return self.base_form_assembly_visitor(e, t, *operands)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/assemble.py", line 537, in base_form_assembly_visitor
    interpolator = firedrake.Interpolator(expression, expr.function_space(), **interp_data)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/pyadjoint/pyadjoint/tape.py", line 110, in wrapper
    return function(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/interpolation.py", line 819, in __init__
    self.callable, arguments = make_interpolator(expr, V, subset, access, bcs=bcs)
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/firedrake/firedrake/src/firedrake/firedrake/interpolation.py", line 998, in make_interpolator
    loops.extend(_interpolator(V, tensor, expr, subset, arguments, access, bcs=bcs))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<decorator-gen-31>", line 2, in _interpolator
  File "/home/firedrake/firedrake/src/firedrake/firedrake/utils.py", line 89, in wrapper
    return f(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/interpolation.py", line 1074, in _interpolator
    kernel = compile_expression(cell_set.comm, expr, to_element, V.ufl_element(),
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/PyOP2/pyop2/caching.py", line 293, in wrapper
    v = func(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/interpolation.py", line 1208, in compile_expression
    return compile_expression_dual_evaluation(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/tsfc/tsfc/driver.py", line 268, in compile_expression_dual_evaluation
    evaluation, basis_indices = to_element.dual_evaluation(fn)
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/FInAT/finat/finiteelementbase.py", line 261, in dual_evaluation
    expr = sum_factorise(sum_indices, factors)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/tsfc/gem/optimise.py", line 358, in sum_factorise
    raise NotImplementedError("Too many indices for sum factorisation!")
NotImplementedError: Too many indices for sum factorisation!
```

angus-g: I guess `mu2_radial.rad` needs to be updated a) for the new radii; and b) for only 8 extruded layers?

sghelichkhani: hmmm, I don't think it has to do with `mu2_radial`. It's a `interpolate_1d_profile` issue that goes back to changing to  `upward_component`. Somehow this works fine for `cartesian=True` (even for spherical case below). But does not work for `cartesian=False` :

```python3
from gadopt import *
from gadopt.utility import vertical_component, upward_normal

rmin, rmax, ref_level, nlayers = 1.208, 2.208, 4, 8
mesh2d = CubedSphereMesh(rmin, refinement_level=ref_level, degree=2)
mesh = ExtrudedMesh(mesh2d, layers=nlayers, extrusion_type='radial')

V = VectorFunctionSpace(mesh, "CG", 2)  # Velocity function space (vector)
Q = FunctionSpace(mesh, "CG", 2)  # Temperature function space (scalar)

upward_component = vertical_component(SpatialCoordinate(mesh), cartesian=False)

q = Function(Q, name="r")
q.interpolate(upward_component)
```

drhodrid:
Just a comment on the testing of this. Like all the demos I think it needs to be tested. I would just take U_RMS and Nu top and bottom after the simulation ends, and verify it's consistent. 

Other minor things:

1. oldest_age=10000 - - should that be 1000?
2. radial viscosity file - I think it would be confusing to a user to have the radii as they are in that file (i.e. 1.22-2.22) because we are interpolating to a value that is out of bounds. I know it still works, but to avoid confusion I agree with @angus-g that they should be updated.

Patol75 (commented):
Any chance to limit the line length (e.g. 88 characters and return) of comments in the demo? It is pretty annoying to have to scroll horizontally to keep reading.

demos/gplates_global/gplates_global.py
sghelichkhani: This is now resolved via the recent PR.

drhodrid:
> Any chance to limit the line length (e.g. 88 characters and return) of comments in the demo? It is pretty annoying to have to scroll horizontally to keep reading.

Done.

drhodrid:
@angus-g - any chance you could take a last look at this before our dev meeting this arvo and flag any outstanding points? I think the tutorial is now good, but can't test locally as I don't currently have pygplates on my mac. So perhaps take values of U_RMS, Surface_RMS (if we have it) and Nu, from your local run, for regression testing? Sia noted above that he has a separate branch for gplates_testing (which I think will add a long test to our testing framework). We can get that merged when he is back.

angus-g:
Just fixed a typo that prevented the demo from actually running, but it still diverges for me:
```
(firedrake) firedrake@ccf2a46d583f:~/g-adopt/demos/gplates_global$ python3 gplates_global.py 
:/python/api/Crossovers.py:199: SyntaxWarning: invalid escape sequence '\@'
:/python/api/Crossovers.py:513: SyntaxWarning: invalid escape sequence '\@'
  0 SNES Function norm 9.793047096475e+05
      Linear Stokes_fieldsplit_1_ solve did not converge due to DIVERGED_NANORINF iterations 0
  File "/home/firedrake/g-adopt/demos/gplates_global/gplates_global.py", line 243, in <module>
    stokes_solver.solve()
  File "/home/firedrake/firedrake/lib/python3.12/site-packages/gadopt/stokes_integrators.py", line 267, in solve
    self.solver.solve()
  File "petsc4py/PETSc/Log.pyx", line 188, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "petsc4py/PETSc/Log.pyx", line 189, in petsc4py.PETSc.Log.EventDecorator.decorator.wrapped_func
  File "/home/firedrake/firedrake/src/firedrake/firedrake/adjoint_utils/variational_solver.py", line 89, in wrapper
    out = solve(self, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/variational_solver.py", line 329, in solve
    solving_utils.check_snes_convergence(self.snes)
  File "/home/firedrake/firedrake/src/firedrake/firedrake/solving_utils.py", line 124, in check_snes_convergence
    raise ConvergenceError(r"""Nonlinear solve failed to converge after %d nonlinear iterations.
firedrake.exceptions.ConvergenceError: Nonlinear solve failed to converge after 0 nonlinear iterations.
Reason:
   DIVERGED_LINEAR_SOLVE
```

angus-g (approved):
I think this is ready to go now?

drhodrid:
Go go go!

angus-g:
@drhodrid I think you'll need to change to an approving review

drhodrid (approved):

PR #49, opened 2024-01-10
Informing `StokesSolver` when our LHS is not changing in time. 

sghelichkhani:
## Issue
In many of demos, as well as some simplified Earth-like problems, we do  not have time-varying viscosity. For example for our adjoint optimisations, we might do many iterations with only a depth dependent viscosity to obtain a better "initial guess", before considering temperature-(or other temporally changing) dependencies. Currently we do assume the LHS of the stokes problem needs to be updated every timestep, whereas in the aforementioned cases this is not the case, as the Jacobian needs to be assembled once. 

## Change
With this PR, if `constant_jacobian=True` is provided to `StokesSolver` we setup a `LinearVariationalProblem` with `constant_jacobian=True`, which should result in considerable speed up in run-time for cases with temporally constant viscosities.

### DIFF
--- a/demos/3d_spherical/3d_spherical.py
+++ b/demos/3d_spherical/3d_spherical.py
@@ -103,7 +103,7 @@ stokes_bcs = {
 
 energy_solver = EnergySolver(T, u, approximation, delta_t, ImplicitMidpoint, bcs=temp_bcs)
 stokes_solver = StokesSolver(z, T, approximation, bcs=stokes_bcs,
-                             cartesian=False, J="constant",
+                             cartesian=False, constant_jacobian=True,
                              nullspace=Z_nullspace, transpose_nullspace=Z_nullspace,
                              near_nullspace=Z_near_nullspace)
 
--- a/demos/3d_spherical/3d_spherical.py
+++ b/demos/3d_spherical/3d_spherical.py
@@ -103,7 +103,7 @@ stokes_bcs = {
 
 energy_solver = EnergySolver(T, u, approximation, delta_t, ImplicitMidpoint, bcs=temp_bcs)
 stokes_solver = StokesSolver(z, T, approximation, bcs=stokes_bcs,
-                             cartesian=False,
+                             cartesian=False, J="constant",
                              nullspace=Z_nullspace, transpose_nullspace=Z_nullspace,
                              near_nullspace=Z_near_nullspace)
 
--- a/gadopt/stokes_integrators.py
+++ b/gadopt/stokes_integrators.py
@@ -2,6 +2,8 @@ from .momentum_equation import StokesEquations
 from .utility import upward_normal, ensure_constant
 from .utility import log_level, INFO, DEBUG, depends_on
 import firedrake as fd
+import ufl
+from firedrake import slate
 
 iterative_stokes_solver_parameters = {
     "mat_type": "matfree",
@@ -110,7 +112,17 @@ class StokesSolver:
         self.approximation = approximation
         self.mu = ensure_constant(mu)
         self.solver_parameters = solver_parameters
-        self.J = J
+
+        if isinstance(J, (ufl.BaseForm, slate.TensorBase)):
+            self.J = J
+            self.constant_jacobian = False
+        elif isinstance(J, str) and J == "constant":
+            self.J = None
+            self.constant_jacobian = True
+        else:
+            raise TypeError(
+                "Provided Jacobian is a '%s', and is not valid" % type(J).__name__)
+
         self.linear = not depends_on(self.mu, self.solution)
 
         self.solver_kwargs = kwargs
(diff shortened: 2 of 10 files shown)


demos/3d_spherical/3d_spherical.py
```diff
@@ -103,7 +103,7 @@
 
 energy_solver = EnergySolver(T, u, approximation, delta_t, ImplicitMidpoint, bcs=temp_bcs)
 stokes_solver = StokesSolver(z, T, approximation, bcs=stokes_bcs,
-                             cartesian=False,
+                             cartesian=False, J="constant",
```
stephankramer, line 106: Not a big fan of overloading existing arguments with new meanings depending on type. Providing a custom Jacobian is clearly a separate concern from informing the solver whether the value of the Jacobian is going to be the same in every solve. What happens if we want to provide a custom Jacobian that "is constant"? I would just add a `constant_jacobian` logical argument - using the same name as in Firedrake so people familiar with that in firedrake may recognize it

stephankramer (changes requested):
Great find! Any idea what the tests are failing on? It would be good to see a before and after timings of a complete test run.

demos/3d_spherical/3d_spherical.py
sghelichkhani: Agreed! I have now switched to having `constant_jacobian` for `StokesSolver`.

sghelichkhani:
> Great find! Any idea what the tests are failing on? It would be good to see a before and after timings of a complete test run.

That was actually a silly mistake related to another change. I will comment on that. 

For comparing timings for constant-jacobian and master:
<!DOCTYPE html>
Case | `constant_jacobian=True` | master
-- | -- | --
base_case.py | 65.845703s | 177.611435s
compressible_case_ALA.py | 396.284546s | 710.862610s
compressible_case_TALA.py | 399.185913s | 711.363647s
forward.py | 54.320751s | 140.825516s
2d_cylindrical.py | 1302.710815s | 1383.112183s
3d_cartesian.py | 894.135986s | 1309.180420s
3d_spherical.py | 653.593262s | 710.387878s


So it does lead to a significant speed up. Although some cases are weirdly not that affected. (I assume we do not do a lot of timesteps for them?)

gadopt/stokes_integrators.py
```diff
@@ -130,17 +148,17 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
         self.strong_bcs = []
         for id, bc in bcs.items():
             weak_bc = {}
-            for type, value in bc.items():
-                if type == 'u':
+            for bc_type, value in bc.items():
```
sghelichkhani, line 151: @stephankramer So I changed `type` to `bc_type` to avoid shadowing the built-in function. Mostly because my formatter was refusing to let it go! :-D

gadopt/stokes_integrators.py
```diff
@@ -169,13 +187,26 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
...
-        self.solver = fd.NonlinearVariationalSolver(self.problem,
-                                                    solver_parameters=self.solver_parameters,
-                                                    options_prefix=self.name,
-                                                    appctx={"mu": self.mu},
-                                                    **self.solver_kwargs)
+        if self.constant_jacobian:
+            z_tri = fd.TrialFunction(self.Z)
+            F_stokes_lin = fd.replace(self.F, {self.solution: z_tri})
+            a, L = fd.lhs(F_stokes_lin), fd.rhs(F_stokes_lin)
+            self.problem = fd.LinearVariationalProblem(a, L, self.solution,
```
sghelichkhani, line 194: @stephankramer One of my motivations for overloading `J`, was that in `fd.LinearVariationalProbelm` it is assumed that `J = a` (`a` being the LHS). So if both `J` and `constant_jacobian` is provided, probably means that the user is not aware how this works? 

Additionally, what I still don't understand is the role of `aP` for preconditioning the system, and how this goes with all the other preconditioners we use. Our interface does not allow for providing `aP`.

sghelichkhani (commented):
All tests are passed! So once @stephankramer has a look, I think this is good to go.

gadopt/stokes_integrators.py
stephankramer: Fair enough - I think you previously also had a `raise TypeError(f"Unexpected {type(bla)}")` in the same context which may also have caused issues - but yeah fine with that

gadopt/stokes_integrators.py
stephankramer: Ah yes, good point - it indeed doesn't make sense to set a J for the linear case, only makes sense in the context of a Newton solve - but I  do think the interface is cleaner now.

Hm yeah I can't immediately think of a good use case for Stokes (which is indeed probably why we don't provide an interface for it) - but say you have some weakly coupled ("mixed") linear system, then in `a` you would supply the full equations including the coupling terms, which correspond to off-diagonal blocks, but that system might be hard to solve directly, so you supply a `aP` where you leaves these terms out so you end up with a block-diagonal system and then you could set your preconditioner to `"pc_type": "ksp"` to exactly solve the block-diagonal system which would already be a good approximation as a preconditioner to the full system. So say in the future we want to couple Stokes and Temperature (also makes it nonlinear but let's ignore that) we could approximate it in `aP` with the uncoupled system. The equivalent in the nonlinear case would be Jp btw, so if we do add support for that, we probably just need to add `aP=Jp` as well for `constant_jacobian` case.

gadopt/stokes_integrators.py
```diff
@@ -110,7 +112,23 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
         self.approximation = approximation
         self.mu = ensure_constant(mu)
         self.solver_parameters = solver_parameters
+
+        # Assume Jacobian varies in time unless specified otherwise
+        self.constant_jacobian = False
```
stephankramer, line 117: This line is redundant now

gadopt/stokes_integrators.py
```diff
@@ -110,7 +112,23 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
...
+        if J is not None:
+            if isinstance(J, (ufl.BaseForm, slate.TensorBase)):
+                self.J = J
+            elif isinstance(J, str) and J.lower() == "constant":
+                # Handling constant Jacobian
+                self.constant_jacobian = True
+            else:
+                # Raising error for invalid Jacobian type
+                raise TypeError(f"Invalid Jacobian type: {repr(J)}")
+
```
stephankramer, line 131: I think this can all go now, right?

gadopt/stokes_integrators.py
```diff
@@ -2,6 +2,8 @@
 from .utility import upward_normal, ensure_constant
 from .utility import log_level, INFO, DEBUG, depends_on
 import firedrake as fd
+import ufl
+from firedrake import slate
```
stephankramer, line 6: Also check whether you still need these

stephankramer (changes requested):
Some clean up after interface change - other than that, good to go!

gadopt/stokes_integrators.py
sghelichkhani: Gotcha! As always thanks for the helpful explanation.

gadopt/stokes_integrators.py
sghelichkhani: Removed now!

gadopt/stokes_integrators.py
sghelichkhani: All gone now!

sghelichkhani:
Okido! I cleaned up the unnecessary code and will merge after the regressions tests.

PR #50, opened 2024-01-15
Install updated CMake image when building pyrol wheel for Actions

angus-g:
With some recent changes (including moving our working branch of pyrol to master), the Ubuntu-provided CMake isn't sufficiently new enough (even though none of the new features are actually used :roll_eyes:). This change only affects the case when the pyrol wheel is rebuilt, i.e. when it isn't cached, so it shouldn't hit every invocation of the workflow.


sghelichkhani (approved):
All looks clean! Thanks Angus.

PR #51, opened 2024-01-16
Add free surface code 

w3168:
I have copied over only the free surface code from the old viscoelastic branch and checked that the convergence rates are still correct for the viscous free surface relaxation.

Still a lot of tidying up to do:

- ~~Hooking up github actions and Angus' testing framework (and linting the tests _scripts)~~ _Done_
- ~~I don't like the user adding an additional fields dictionary in the runscript~~ _This is changed now so it is just optional arguments passed to stokes solver_
- ~~I haven't generalised this to work with bottom free surfaces ( or having a top and bottom free surface at the same time)~~ _This now works with bottom and top free surface_
- ~~Add more test cases rayleigh taylor-drunken sailor and also an annulus domain...~~ _Added cylindrical test case which works ok. Going to work on RT drunken sailor with Thomas using his level set code separately._
- ~~Trying this in 3d with iterative solvers and getting the schur complement factorisation right...~~ _Works with iterative solvers in 2D (box and annulus)_


w3168:
I have remembered two more things for the to do list: 
- go back to default mumps for gadopt for 2d
- allow a nonzero density on the surface

angus-g (commented):
Looking pretty good so far. I guess the free surface equation is actually pretty simple!

For the moment, I like the idea of separating the free surface specifics out of `StokesSolver`, so I'll leave that as my major comment for now.

The demos look pretty good, although given the similarity of the implicit and explicit versions, I wonder if it would be instructive to, e.g. base the implicit on the explicit and demonstrate just the changes required? Maybe leave that for the moment, but I'll think about ways through that.

w3168:
Thanks Angus!

demos/free_surface/explicit_viscous_surface.py
```diff
@@ -0,0 +1,140 @@
...
+    eta_timestepper = BackwardEuler(eta_eq, eta, eta_fields, dt, bnd_conditions=eta_bcs, strong_bcs=eta_strong_bcs, solver_parameters=mumps_solver_parameters)
+
+    # analytical function
+    eta_analytical = Function(W, name="eta analytical")
+    eta_analytical.interpolate(exp(-time/tau0)*F0 * cos(kk * X[0]))
+
+    if do_write:
+        # Create output file and select output_frequency:
+        filename = "explicit_viscous_freesurface"
+        output_file = File(filename+"_D"+str(float(D/L0))+"_mu"+str(float(mu))+"_nx"+str(nx)+"_dt"+str(float(dt/tau0))+"tau.pvd")
```
stephankramer, line 107:
```suggestion
        output_file = File(f"{filename}_D{float(D/L0))}_mu{float(mu)}_nx{nx}_dt{float(dt/tau0))}tau.pvd")
```

demos/free_surface/explicit_viscous_surface.py
```diff
@@ -0,0 +1,140 @@
...
+        'snes_type': 'ksponly',
+        'ksp_type': 'preonly',
+        'pc_type': 'lu',
+        'pc_factor_mat_solver_type': 'mumps',
+        'mat_type': 'aij',
+        'snes_max_it': 100,
+        'snes_rtol': 1e-8,
+        'snes_atol': 1e-6,
+        'mat_mumps_icntl_14': 200
+    }
```
stephankramer, line 93: How is this different from the defaults in `StokesSolver` ? Why do you need a Newton solver here - isn't this a linear system? If you do need additional options, in addition to the defaults you can do:
```
  stokes_solver = StokesSolver(...)  # without solver_parameters, sets default options
  stokes_solver.solver_parameters.update({
      # only options that are different from the defaults
   })
```

demos/free_surface/implicit_viscous_surface.py
```diff
@@ -0,0 +1,137 @@
...
+        'snes_type': 'ksponly',
+        'ksp_type': 'preonly',
+        'pc_type': 'lu',
+        'pc_factor_mat_solver_type': 'mumps',
+        'mat_type': 'aij',
+        'snes_max_it': 100,
+        'snes_rtol': 1e-8,
+        'snes_atol': 1e-6,
+        'mat_mumps_icntl_14': 200
+    }
```
stephankramer, line 84: Same question as above

demos/free_surface/implicit_viscous_surface.py
```diff
@@ -0,0 +1,137 @@
...
+        # Write output:
+        if do_write:
+            eta_midpoint.append(eta_.at((L/L0)/2, (D/L0)-0.001/L0))
+
+            if timestep % dump_period == 0:
+                log("timestep", timestep)
+                log("time", float(time))
+                output_file.write(u_, eta_, p_, eta_analytical)
+    if do_write:
+        with open(filename+"_D"+str(float(D/L0))+"_mu"+str(float(mu))+"_nx"+str(nx)+"_dt"+str(float(dt/tau0))+"tau.txt", 'w') as file:
```
stephankramer, line 125: I'd suggest f-strings

demos/free_surface/implicit_viscous_surface.py
```diff
@@ -0,0 +1,137 @@
...
+    eta_analytical = Function(W, name="eta analytical")
+    eta_analytical.interpolate(exp(-time/tau0)*F0 * cos(kk * X[0]))
+
+    # Create output file and select output_frequency:
+    if do_write:
+        # Write output files in VTK format:
+        dump_period = 1
+        log("dump_period ", dump_period)
+        filename = "implicit_viscous_freesurface"
+        output_file = File(filename+"_D"+str(float(D/L0))+"_mu"+str(float(mu))+"_nx"+str(nx)+"_dt"+str(float(dt/tau0))+"tau.pvd")
```
stephankramer, line 101: I'd suggest f-strings

gadopt/equations.py
```diff
@@ -45,14 +45,21 @@ def __init__(self, test_space, trial_space, quad_degree=None, **kwargs):
...
 
     def mass_term(self, test, trial):
         r"""Return the UFL for the mass term \int test * trial * dx typically used in the time term."""
-        return firedrake.inner(test, trial) * self.dx
+
+        if 'free_surface_id' in self.kwargs:
+            log("using surface mass term...")
+            return firedrake.inner(test, trial) * self.ds(self.kwargs['free_surface_id'])
+        else:
+            return firedrake.inner(test, trial) * self.dx
```
stephankramer, line 62: I think a nicer way to do this is to have FreeSurfaceEquation overload `mass_term()` so you don't have to mess with equation specific things in the base equation.

gadopt/momentum_equation.py
```diff
@@ -115,6 +116,11 @@ def residual(self, test, trial, trial_lagged, fields, bcs):
                 # here we need only the third term, because we assume jump_u=0 (u_ext=u)
                 # the provided stress = n.(mu.stress_tensor)
                 F += dot(-phi, bc['stress']) * self.ds(id)
+            if 'free_surface_stress' in bc:
+                # add the free surface stress
+                free_surface_stress = bc['free_surface_stress']
+                F -= dot(-phi, free_surface_stress * n) * self.ds(id)
```
stephankramer, line 122: From the implementation it looks like `'free_surface_stress'` is just a "normal_stress" should we call it that?

Similar to the current checks at the end of this subroutine, you probably want to checks against the combination of `u` or `un` with "normal_stress"/"free_surface_stress". Setting a tangential u component with a free surface _might_ make sense (I vaguely recall having tried that in Fluidity for some case) but we can add it if/when someone needs it.

stephankramer (changes requested):
Very nice. Some additional suggestions/questions...

demos/free_surface/explicit_viscous_surface.py
w3168: done!

demos/free_surface/explicit_viscous_surface.py
w3168: thanks I meant to go back to the defaults ( I had copied these from Thwaites at one point when I was debugging)

demos/free_surface/implicit_viscous_surface.py
w3168: done

demos/free_surface/implicit_viscous_surface.py
w3168: done

gadopt/equations.py
w3168: Thanks I've moved the mass term to FreeSurfaceEquation!

gadopt/momentum_equation.py
w3168: I've renamed "free_surface_stress" to "normal_stress" (I think this is clearer too...!)

w3168: and I've also added some checks at the end for combinations of u/un and normal  stress

demos/free_surface/implicit_viscous_surface.py
w3168: done

demos/free_surface/test_viscous_surface.py
```diff
@@ -0,0 +1,31 @@
...
+    # use the highest resolution analytical solutions as the reference in scaling
+    ref = errors[-1]
+    relative_errors = errors / ref
+    convergence = np.log2(relative_errors[:-1] / relative_errors[1:])
+
+    assert np.allclose(convergence, expected_convergence, rtol=1e-1)
+
+
+if __name__ == "__main__":
+    print("Implicit-freesurface-coupling_")
```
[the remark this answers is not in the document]
w3168: Done!

demos/free_surface/test_viscous_surface.py
```diff
@@ -0,0 +1,31 @@
...
+
+@pytest.fixture
+def expected_errors(coupling):
+    return np.loadtxt(base / f"expected-errors-{coupling}-free-surface-coupling.dat")
+
+
+@pytest.mark.parametrize("coupling,expected_convergence", [("explicit", [1.4, 1.1, 1.0]), ("implicit", [2.0, 2.0, 2.0])])
+def test_scalar_advection_diffusion_DH27(coupling, expected_convergence, expected_errors):
+
+    expected_errors = expected_errors
```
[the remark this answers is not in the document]
w3168: oops... 
removed now thanks!

gadopt/free_surface_equation.py
```diff
@@ -0,0 +1,40 @@
...
+NOTE: for all terms, the residual() method returns the residual as it would be on the RHS of the equation, i.e.:
+
+  dq/dt = \sum term.residual()
+
+This sign-convention is for compatibility with Thetis' timeintegrators. In general, however we like to think about
+the terms as they are on the LHS. Therefore in the residual methods below we assemble in F as it would be on the LHS:
+
+  dq/dt + F(q) = 0
+
+and at the very end "return -F".
```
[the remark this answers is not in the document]
w3168: Thanks for spotting this!

gadopt/stokes_integrators.py
```diff
@@ -97,33 +97,32 @@ def create_stokes_nullspace(Z, closed=True, rotational=False, translations=None)
...
-                 closed=True, rotational=False, J=None, constant_jacobian=False,
-                 **kwargs):
+    def __init__(self, z, T, approximation, bcs=None, mu=1, quad_degree=6,
+                 cartesian=True, solver_parameters=None, closed=True,
+                 rotational=False, J=None, constant_jacobian=False, **kwargs):
+
         self.Z = z.function_space()
         self.mesh = self.Z.mesh()
         self.test = fd.TestFunctions(self.Z)
-        self.equations = StokesEquations(self.Z, self.Z, quad_degree=quad_degree,
```
sghelichkhani, line 107: So based on what I see you can still inherit most of the things from `StokesSolver`. So for this here I would suggest that you keep it as it is.

gadopt/stokes_integrators.py
```diff
@@ -97,33 +97,32 @@ def create_stokes_nullspace(Z, closed=True, rotational=False, translations=None)
...
         self.approximation = approximation
         self.mu = ensure_constant(mu)
+        self.quad_degree = quad_degree
         self.solver_parameters = solver_parameters
         self.J = J
         self.constant_jacobian = constant_jacobian
         self.linear = not depends_on(self.mu, self.solution)
-
         self.solver_kwargs = kwargs
-        u, p = fd.split(self.solution)
```
sghelichkhani, line 118: change this to  `solutions_split = fd.split(self.solution)` instead of dropping it.

gadopt/stokes_integrators.py
```diff
@@ -97,33 +97,32 @@ def create_stokes_nullspace(Z, closed=True, rotational=False, translations=None)
...
         self.k = upward_normal(self.Z.mesh(), cartesian)
+
+        # Note that velocity, pressure and the buoyancy term are added to fields later in initialise().
+        # This allows there to be three trial functions when 'splitting' z for the implicit free surface coupling.
         self.fields = {
-            'velocity': u,
-            'pressure': p,
             'viscosity': self.mu,
```
sghelichkhani, line 123: here you can then change it to 
```
'velocity': solutions_split[0]
'pressure': solutions_split[1]
```

gadopt/stokes_integrators.py
```diff
@@ -193,6 +211,62 @@ def setup_solver(self):
...
+        theta = 0.5
+        eta_theta = (1-theta)*eta_old + theta*eta
+
+        # Add velocity, pressure and buoyancy term to the fields dictionary
+        self.fields['velocity'] = u
+        self.fields['pressure'] = p
+        self.fields['source'] = self.approximation.buoyancy(p, self.T) * self.k
+
+        # Initialise Stokes equations with a free surface
+        self.equations = FreeSurfaceStokesEquations(self.Z, self.Z, quad_degree=self.quad_degree,
```
sghelichkhani, line 254: Your `FreeSurfaceStokesEquations` is actually identical to `StokesEquations`. 
So you can simply write 
```
self.equations = StokesEquation(test_space.sub(2), trial_space.sub(2), quad_degree=quad_degree, **kwargs)
self.equations +=  FreeSurfaceEquation(test_space.sub(2), trial_space.sub(2), quad_degree=quad_degree, **kwargs)
```

gadopt/stokes_integrators.py
```diff
@@ -193,6 +211,62 @@ def setup_solver(self):
...
+
+    def initialise(self):
+        # Solution is made up of 3 trial functions so we have a separate initialise()
+        u, p, eta = fd.split(self.solution)
+        u_old, p_old, eta_old = fd.split(self.solution_old)
+
+        # Theta timestepping scheme for free surface variable
+        # Forward Euler (theta = 0), Crank Nicholson (thetat = 0.5), Backward Euler (theta = 1)
+        theta = 0.5
+        eta_theta = (1-theta)*eta_old + theta*eta
```
sghelichkhani, line 246: So looking at how you use eta, and how you do time loops, I wonder if there is a way to be consistent with how you are doing your time integration. Here you are hard coding `$\theta=0.5$`, but then when time-stepping in your actual time integration you often use `BackwardEuler`. 

Correct me if I am wrong, but the correct way of doing this would be to supply a time-stepping scheme to `FreeSurfaceStokesSolver` that will self-consistently first solve Stokes with the consistent approximation of `eta` in time, and then advance `eta` with the same approximation.

sghelichkhani (changes requested):

gadopt/stokes_integrators.py
```diff
@@ -169,6 +168,25 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
         # so people can overwrite parameters we've setup here
         self._solver_setup = False
 
+    def initialise(self):
```
stephankramer, line 171: Maybe with @sghelichkhani 's suggestion you can avoid the need for a separate code path that only gets executed in the pure (no fs) Stokes case, i.e. you execute all the standard Stokes code and just modify afterwards what needs to be different for the fs solver.
If you do need this separate initializer:
* it should have a more descriptive name `__init__` and `setup_solver` are also initialisation - so I suggest something like `setup_equations`
* do we really need to change the API for this. The reason we defer the creation of the variational solver is that we allow the user to fiddle (add to) the equations and adjust default solver parameters. So a similar justification would be that we want to allow changing stuff that's been setup in `__init__` _before_ the equations are setup. It might well be there are/will be cases where we want to allow this but I'm not sure why we need it now: if you do need a distinct `setup_equations()` you can just call it in `StokesSolver.__init__`. When the latter is being called from `FreeSurfaceStokesSolver.__init__` it will then call your overridden  `setup_equations()` from FreeSurfaceStokesSolver. In that way the equations would already be setup in `__init__` (as it is now) so we don't have to change the API.

gadopt/stokes_integrators.py
```diff
@@ -169,6 +168,25 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
...
+
+        # Add velocity, pressure and buoyancy term to the fields dictionary
+        u, p = fd.split(self.solution)
+        self.fields['velocity'] = u
+        self.fields['pressure'] = p
+        self.fields['source'] = self.approximation.buoyancy(p, self.T) * self.k
```
stephankramer, line 177: Try to avoid unneeded  repetition - this is the same code as in StokesSolver

gadopt/stokes_integrators.py
```diff
@@ -193,6 +211,62 @@ def setup_solver(self):
...
+
+        # Theta timestepping scheme for free surface variable
+        # Forward Euler (theta = 0), Crank Nicholson (thetat = 0.5), Backward Euler (theta = 1)
+        theta = 0.5
+        eta_theta = (1-theta)*eta_old + theta*eta
+
+        # Add velocity, pressure and buoyancy term to the fields dictionary
+        self.fields['velocity'] = u
+        self.fields['pressure'] = p
+        self.fields['source'] = self.approximation.buoyancy(p, self.T) * self.k
```
stephankramer, line 251: Try to avoid unneeded  repetition - this is the same code as in StokesSolver

gadopt/stokes_integrators.py
w3168: Thanks for chatting yesterday and going through this. I agree it should inherit as much as possible!

gadopt/stokes_integrators.py
w3168: I've changed this now to only take the first two elements of split.

gadopt/stokes_integrators.py
w3168: fixed as above.

gadopt/stokes_integrators.py
w3168: I think we discussed this separately yesterday so hopefully it is ok.

One thing is whether the last equation should be multiplied by theta so that the u, p, eta block is symmetric... 

e.g from Stephan's 2012 paper:

image

gadopt/stokes_integrators.py
w3168: Thanks Stephan, as I mentioned on Zulip I've had a go at doing this. I think it makes sense to have these separate methods because I hope this removes repeating too much code. I think this means that we don't have to repeat the looping over the terms to make sure the weak bc for the free surface normal stress gets into F. It also means we can pass the free surface id to the FreeSurfaceEquations.

gadopt/stokes_integrators.py
w3168: I think these new changes have removed the code repetitions

gadopt/stokes_integrators.py
w3168: actually changed it back to how you suggested!

gadopt/stokes_integrators.py
w3168: As I mentioned on Zulip I've changed this again to try and be a bit neater with the inheritance so now FreeSurfaceStokesSolver.setup_equations() calls super().setup_equations and there is also an overloaded FreeSurfaceStokesSolver.solve()

gadopt/stokes_integrators.py
```diff
@@ -101,49 +102,41 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
...
         self.fields = {
-            'velocity': u,
-            'pressure': p,
+            'velocity': self.stokes_vars[0],
+            'pressure': self.stokes_vars[1],
             'viscosity': self.mu,
             'interior_penalty': fd.Constant(2.0),  # allows for some wiggle room in imposition of weak BCs
                                                    # 6.25 matches C_ip=100. in "old" code for Q2Q1 in 2d.
-            'source': self.approximation.buoyancy(p, T) * self.k,
+            'source': self.approximation.buoyancy(self.stokes_vars[1], T) * self.k,
```
Patol75, line 132: Didn't we prefer `self.fields["pressure"]` to make it more explicit?

gadopt/stokes_integrators.py
```diff
@@ -195,4 +211,58 @@ def setup_solver(self):
...
+        # StokesSolver.setup_equations() in StokesSolver.__init__()
+
+        # First setup Stokes equations (without a free surface)
+        super().setup_equations()
+
+        # Define free surface variables for timestepping
+        assert len(self.stokes_vars) == 3
+        self.solution_old = fd.Function(self.solution)
+        self.eta = self.stokes_vars[2]
+        self.eta_old = fd.split(self.solution_old)[2]
```
Patol75, line 245: Do we actually need `solution_old`, or can we just use `eta_old`?

gadopt/stokes_integrators.py
w3168: Thanks for having a look over this thomas. When I actually ran this, the problem with this was that fields doesn't exist yet!

gadopt/stokes_integrators.py
w3168: I think when I tried this I couldn't get the assign to work from eta (part of the mixed space) to eta_old though I might have done something stupid. I agree I think it would maybe be neater with just eta for now as we don't need the old velocity or pressure.

gadopt/stokes_integrators.py
Patol75: Very true! Should be fine as is then :)

gadopt/stokes_integrators.py
Patol75: What if you have `self.eta_old = fd.Function(self.eta)`?

w3168: I think I tried that and it didn't work because of the mixed spaces... But maybe I should try again

gadopt/stokes_integrators.py
w3168: So I have updated the code now so that the bottom equation is multiplied by theta but i think I shouldnt have multiplied the u.n term by delta t...

w3168: So i think now it is better with only theta multiplied by u.n... but should is it now antisymmetric? do i need a minus sign...

gadopt/stokes_integrators.py
```diff
@@ -140,14 +149,58 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
...
+                elif bc_type == 'free_surface':
+                    # Copy free surface information to a new dictionary
+                    free_surface_dict[id] = value
+                    self.free_surface = True
                 else:
                     weak_bc[bc_type] = value
             self.weak_bcs[id] = weak_bc
 
+        if self.free_surface:
+            assert free_surface_dt is not None
```
angus-g, line 161: Just a minor one: I think this should probably be a check with a `raise` if `free_surface_dt` isn't provided, to give a more informative exception.

w3168: Thanks Angus - I've just added a type error in with a short description

demos/free_surface/implicit_top_bottom_buoyancy_free_surface.py
```diff
@@ -0,0 +1,68 @@
...
+    def initialise_temperature(self):
+        self.alpha = 2e-5
+        self.Q_temp_scale = 1
+        self.forcing_depth = 0.5
+        width = (1/self.nx)
+        delta = exp(-pow((self.X[1]-self.forcing_depth)/width, 2)) / (width * sqrt(pi))
+        self.T = Function(self.Q, name="Temperature").interpolate(delta * self.Q_temp_scale * cos(self.kk * self.X[0]))  # Initialise temperature field
+
+    def initialise_approximation(self):
+        Ra = Constant(1)  # Rayleigh number, here we set this to zero as there are no bouyancy terms
```
angus-g, line 31: That's not zero!

demos/free_surface/explicit_free_surface.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+class ExplicitFreeSurfaceModel:
+    # Test case from Section 3.1.1 of `An implicit free surface algorithm
+    # for geodynamical simulations', Kramer et al 2012.
+
+    name = "explicit"
+    bottom_free_surface = False
+
+    def __init__(self, dt_factor, nx=80, do_write=False, cartesian=True):
```
angus-g, line 13: As mentioned in the comment for `test_viscous_surface.py`, maybe this should also take kwargs, for unknowns that are passed to subclasses (e.g. `iterative_2d`):

```suggestion
    def __init__(self, dt_factor, nx=80, do_write=False, cartesian=True, **kwargs):
```

demos/free_surface/implicit_cylindrical_free_surface.py
```diff
@@ -0,0 +1,56 @@
...
+
+class CylindricalImplicitFreeSurfaceModel(ImplicitFreeSurfaceModel):
+    # Free surface relaxation test in a cylindrical domain.
+
+    name = "implicit-cylindrical-iterative"
+    bottom_free_surface = False
+
+    def __init__(self, dt_factor, do_write=True):
+        self.cartesian = False
+        super().__init__(dt_factor, do_write=do_write, cartesian=self.cartesian)
```
angus-g, line 15: Maybe the following pattern is a bit more general (also keeps `do_write` at its false default)

```suggestion
    def __init__(self, dt_factor, **kwargs):
        super().__init__(dt_factor, cartesian=False, **kwargs)
```

demos/free_surface/implicit_free_surface.py
```diff
@@ -0,0 +1,65 @@
...
+
+    name = "implicit"
+    bottom_free_surface = False
+
+    def __init__(self, dt_factor, nx=80, do_write=False, iterative_2d=False, cartesian=True):
+        self.do_write = do_write
+        self.iterative_2d = iterative_2d
+        self.cartesian = cartesian
+
+        super().__init__(dt_factor, nx=nx, do_write=do_write, cartesian=cartesian)
```
angus-g, line 18: Maybe:

```suggestion
    def __init__(self, dt_factor, iterative_2d=False, **kwargs):
        super().__init__(dt_factor, **kwargs)
        self.iterative_2d = iterative_2d
```

demos/free_surface/implicit_top_bottom_buoyancy_free_surface.py
```diff
@@ -0,0 +1,68 @@
...
+class BuoyancyTopBottomImplicitFreeSurfaceModel(TopBottomImplicitFreeSurfaceModel):
+    # Test case from Section 3.1.3 of `An implicit free surface algorithm
+    # for geodynamical simulations', Kramer et al 2012.
+
+    name = "implicit-buoyancy"
+    bottom_free_surface = True
+
+    def __init__(self, dt_factor, nx=320, do_write=False, iterative_2d=False, cartesian=True):
+
+        super().__init__(dt_factor, nx=nx, do_write=do_write, iterative_2d=iterative_2d, cartesian=cartesian)
```
angus-g, line 15: I think you're only overriding the `nx` default here:

```suggestion
    def __init__(self, dt_factor, nx=320, **kwargs):
        super().__init__(dt_factor, nx=nx, **kwargs)
```

demos/free_surface/explicit_free_surface.py
```diff
@@ -0,0 +1,187 @@
...
+        self.stokes_vars[0].rename("Velocity")
+        self.stokes_vars[1].rename("Pressure")
+
+    def initialise_wavenumber(self):
+        lam_dimensional = self.D/2  # wavelength of load in m
+        self.lam = lam_dimensional/self.L0  # dimensionless lambda
+        self.kk = Constant(2 * pi / self.lam)  # wavenumber (dimensionless)
+
+    def initialise_temperature(self):
+        self.T = Function(self.Q, name="Temperature").assign(0)  # Setup a dummy function for temperature
```
angus-g, line 103: Taking this function as an example, I guess I did say this method (of setting a class member rather than returning a value) is fine (and it is!), but the alternative of returning a value is a bit easier to type-check and make sure the inheritance contract is fulfilled.

Having said that, I don't think it's that important in this set of examples where you've already verified the results, and they're all being run (and therefore tested). Just leaving the comment for posterity!

gadopt/free_surface_equation.py
```diff
@@ -0,0 +1,58 @@
...
+"""
+
+
+class FreeSurfaceTerm(BaseTerm):
+    r"""
+    Free Surface term: u \dot n
+    """
+    def residual(self, test, trial, trial_lagged, fields, bcs):
+        assert 'free_surface_id' in self.term_kwargs
```
angus-g, line 24:
```suggestion
```

The next line would raise a `KeyError` anyway.

gadopt/free_surface_equation.py
```diff
@@ -0,0 +1,58 @@
...
+        free_surface_id = self.term_kwargs['free_surface_id']
+
+        # Multiply by constant factor to keep the block system symmetric for the implicit coupling case
+        if 'theta' in self.term_kwargs:
+            prefactor = self.term_kwargs['theta']
+        else:
+            prefactor = 1
```
angus-g, line 31:
```suggestion
        prefactor = self.term_kwargs.get('theta', 1)
```

gadopt/free_surface_equation.py
```diff
@@ -0,0 +1,58 @@
...
+class FreeSurfaceEquation(BaseEquation):
+    """
+    Free Surface Equation.
+    """
+
+    terms = [FreeSurfaceTerm]
+
+    def mass_term(self, test, trial):
+        r"""Return the UFL for the mass term \int test * trial * ds for the free surface time derivative term integrated over the free surface."""
+        assert 'free_surface_id' in self.kwargs
```
angus-g, line 51:
```suggestion
```

gadopt/free_surface_equation.py
```diff
@@ -0,0 +1,58 @@
...
+        free_surface_id = self.kwargs['free_surface_id']
+
+        assert 'k' in self.kwargs
```
angus-g, line 54:
```suggestion
```

gadopt/stokes_integrators.py
```diff
@@ -78,20 +79,26 @@ def create_stokes_nullspace(Z, closed=True, rotational=False, translations=None)
...
     else:
-        p_nullspace = W
+        p_nullspace = stokes_funcspace[1]
 
-    return fd.MixedVectorSpaceBasis(Z, [V_nullspace, p_nullspace])
+    null_space = [V_nullspace, p_nullspace]
+
+    # If free surface unknowns, add dummy free surface nullspace
+    for i in range(len(stokes_funcspace)-2):
+        null_space.append(stokes_funcspace[2+i])
```
angus-g, line 99: Not actually sure if this works, because maybe `subfunctions` doesn't quite return a list?

```suggestion
    # If free surface unknowns, add dummy free surface nullspace
    null_space += stokes_funcspace[2:]
```

Hopefully this idea of extra function spaces doesn't clash with any other physics later on!

gadopt/stokes_integrators.py
```diff
@@ -140,14 +156,67 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
...
+            eta_old = []
+            eta_theta = []
+            self.free_surface_id_list = []
+
+            c = 0  # Counter for free surfaces (N.b. we already have two equations from StokesEquations)
+            for id, value in free_surface_dict.items():
+                if 'exterior_density' in value:
+                    exterior_density = value['exterior_density']
+                else:
+                    exterior_density = 0
```
angus-g, line 182:
```suggestion
                exterior_density = value.get('exterior_density', 0)
```

angus-g (commented):
Nice! I think this is looking pretty good. Maybe just a few tweaks to the demo structure, but otherwise it's pretty close.

demos/free_surface/test_viscous_surface.py
angus-g, line 1: Oh I also wondered if this should be `test_free_surface.py`?

demos/free_surface/test_viscous_surface.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # check that norm(q) is the same as previously run
+    assert np.allclose(errors, expected_errors, rtol=1e-6, atol=1e-16)
+
+    # use the highest resolution analytical solutions as the reference in scaling
+    ref = errors[-1]
+    relative_errors = errors / ref
+    convergence = np.log2(relative_errors[:-1] / relative_errors[1:])
+    print(f"{coupling} convergence:", convergence)
+    print()
```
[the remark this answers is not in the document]
w3168: cool (  I meant to take it out as i was just using it for testing)

demos/free_surface/implicit_top_bottom_free_surface.py
```diff
@@ -0,0 +1,66 @@
...
+
+class TopBottomImplicitFreeSurfaceModel(ImplicitFreeSurfaceModel):
+    # Test case from Section 3.1.2 of `An implicit free surface algorithm
+    # for geodynamical simulations', Kramer et al 2012.
+
+    name = "implicit-both"
+    bottom_free_surface = True
+
+    def __init__(self, dt_factor, nx=80, do_write=True, iterative_2d=False, cartesian=True):
+        self.dt_factor = dt_factor
```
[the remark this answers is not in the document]
w3168: I think I actually need these for the absorption term... but maybe it is not that neat.

demos/free_surface/test_viscous_surface.py
w3168: good idea - I meant to do it before!

gadopt/approximations.py
```diff
@@ -89,6 +89,9 @@ def __init__(self, Ra, kappa=1, g=1, rho=1, alpha=1):
     def buoyancy(self, p, T):
         return self.Ra * self.g * self.alpha * self.rho * T
 
+    def rho_field(self, p, T):
+        return self.rho * (1 - self.alpha * T)
+
```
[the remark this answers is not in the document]
w3168: done

gadopt/__init__.py
```diff
@@ -1,14 +1,15 @@
...
-from .utility import log, ParameterLog, TimestepAdaptor, LayerAveraging, timer_decorator
+from .utility import log, ParameterLog, TimestepAdaptor, LayerAveraging, timer_decorator, InteriorBC
 from .diagnostics import GeodynamicalDiagnostics
 from .momentum_equation import StokesEquations
 from .scalar_equation import EnergyEquation
 from .stokes_integrators import StokesSolver, create_stokes_nullspace
 from .energy_solver import EnergySolver
 from .approximations import BoussinesqApproximation, ExtendedBoussinesqApproximation, AnelasticLiquidApproximation, TruncatedAnelasticLiquidApproximation
 from .preconditioners import SPDAssembledPC, VariableMassInvPC
+from .free_surface_equation import FreeSurfaceEquation
```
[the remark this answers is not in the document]
angus-g: Following our discussion, I hadn't appreciated that you *do* indeed use `FreeSurfaceEquation` directly in the explicit case. Given that, I guess it should go here!

demos/free_surface/explicit_free_surface.py
```diff
@@ -0,0 +1,187 @@
...
+    def setup_nullspaces(self):
+        # Nullspaces and near-nullspaces:
+        self.Z_nullspace = None  # Default: don't add nullspace for now
+        self.Z_near_nullspace = None  # Default: don't add nullspace for now
+
+    def setup_solver(self):
+        # Set up the stokes solver
+        self.stokes_solver = StokesSolver(self.z, self.T, self.approximation, bcs=self.stokes_bcs, mu=self.mu, cartesian=self.cartesian,
+                                          free_surface_dt=self.dt, nullspace=self.Z_nullspace, transpose_nullspace=self.Z_nullspace,
+                                          near_nullspace=self.Z_near_nullspace)
```
angus-g, line 132: I think I confused myself here and thought that `free_surface_dt` meant the free surface is being handled in `StokesSolver`, but that's only the case where there's *also* a `'free_surface'` boundary condition. And in the implicit case, this method is being overridden anyway.

demos/free_surface/implicit_top_bottom_free_surface.py
```diff
@@ -0,0 +1,66 @@
...
+        self.rho_bottom = 2
+        self.zeta_error = 0
+        super().__init__(dt_factor, nx=nx, do_write=do_write, iterative_2d=iterative_2d, cartesian=cartesian)
+
+        if iterative_2d:
+            # Schur complement splitting leads to a nullspace in the velocity block.
+            # Adding a small absorption term bringing the vertical velocity to zero removes this nullspace
+            # and does not effect convergence provided that this term is small compared with the overall numerical error.
+            self.absorption_penalty()
```
angus-g, line 23: Maybe `self.absorption_penalty(dt_factor)`, since it only seems to be called during initialisation anyway?

gadopt/__init__.py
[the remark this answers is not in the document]
w3168: cool ok!

demos/free_surface/implicit_top_bottom_free_surface.py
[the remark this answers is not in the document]
w3168: As we discussed I'll put a comment here because it does look a bit out of place (like it could have been setup earlier...)

demos/free_surface/explicit_free_surface.py
w3168: Thanks I actually didn't mean to put this free_surface_dt here because as you say it doesnt actually do anything if there is no bc! (I've also removed the nullsapce because the implict case will overwrite this anyway

demos/free_surface/implicit_top_bottom_free_surface.py
w3168: sounds good

demos/free_surface/implicit_top_bottom_free_surface.py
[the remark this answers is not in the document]
w3168: (accutally just put dt as an argument as you suggested below)

demos/free_surface/test_viscous_surface.py
```diff
@@ -0,0 +1,94 @@
...
+            errors_iterative.append(simulation.final_error)
+
+            if simulation.bottom_free_surface:
+                errors_zeta_iterative.append(simulation.final_zeta_error)
+
+        if simulation.bottom_free_surface:
+            np.savetxt(f"errors-{model.name}-iterative-top-free-surface-coupling.dat", errors_iterative)
+            np.savetxt(f"errors-{model.name}-iterative-bottom-free-surface-coupling.dat", errors_zeta_iterative)
+        else:
+            np.savetxt(f"errors-{model.name}-iterative-free-surface-coupling.dat", errors_iterative)
```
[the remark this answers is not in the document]
w3168: cool thanks Angus, as discussed I've used this structure and added direct/iterative attributes to all the cases so that it skips the the explicit iterative case and the direct cylindrical case.

gadopt/stokes_integrators.py
w3168: yeah I guess this might not work if other function spaces also require nullspaces but I guess we can change that down the line.

demos/free_surface/explicit_free_surface.py
w3168: done

demos/free_surface/implicit_top_bottom_buoyancy_free_surface.py
w3168: oops thanks - fixed now!

demos/free_surface/implicit_cylindrical_free_surface.py
w3168: ok!

demos/free_surface/implicit_free_surface.py
w3168: sounds good

demos/free_surface/implicit_top_bottom_buoyancy_free_surface.py
w3168: sounds good

demos/free_surface/explicit_free_surface.py
w3168: Cool thanks for explaining that Angus!

gadopt/free_surface_equation.py
w3168: done!

gadopt/stokes_integrators.py
```diff
@@ -60,15 +61,15 @@ def create_stokes_nullspace(Z, closed=True, rotational=False, translations=None)
     """
     X = fd.SpatialCoordinate(Z.mesh())
     dim = len(X)
-    V, W = Z.subfunctions
+    stokes_funcspace = Z.subfunctions
```
Patol75, line 64: Out of curiosity, why `funcspace`? I think this is slightly confusing, given that `subfunctions` returns a list of `Functions` and that Firedrake also defines `FunctionSpace`.

w3168: So I am pretty sure they are returning functionspaces, because Z is the mixed function space. If you look [here](https://www.firedrakeproject.org/firedrake.html#firedrake.functionspaceimpl.FunctionSpace) Functionspace does have an attribute called subfunctions which gives you the component function spaces. I agree it seems like a confusing naming convention

Patol75: My bad. I did not know that was a thing. Thanks, Will.

drhodrid:
From a user perspective, I think this now looks very good and intuitive. The demos are also really nicely put together, with systematically increasing complexity. Well done @w3168. I would like @stephankramer to take a look though, prior to merging, just to make sure the changes align with his previous advice.

drhodrid:
Also a question for Steph - in Fluidity, we have this option for variable_density, relating to whether or not material at the free surface (within the domain) has a variable density. Is that something that will get handled naturally in Will's implementation above, or is it something we need to worry about? Perhaps something to discuss at our next zoom!

w3168:
Thanks for looking at this @drhodrid 

> Also a question for Steph - in Fluidity, we have this option for variable_density, relating to whether or not material at the free surface (within the domain) has a variable density. Is that something that will get handled naturally in Will's implementation above, or is it something we need to worry about? Perhaps something to discuss at our next zoom!

I added a flag for this in this commit (https://github.com/g-adopt/g-adopt/pull/51/commits/d6a5458da91ff719fbc0043ad8eb899d0788f9b5)  so it should be handled now. Originally it was just using 'rho' from approximation for the interior density which is the constant reference density.

stephankramer (changes requested):
I'd say nearly - we should get it in as soon as possible.
There's two little things that are nagging me:
* the `rho_field` thing. Oh actually, can we just combine this with the variable_density logic, i.e. make that argument an argument of the approximation, not of the stokes solver, so something like:
```
class BoussinesqApproximation(...)
   ...
   def __init__(self, ..., variable_free_surface_density=False)
       ....
       self.variable_free_surface_density = variable_free_surface_density
       
   def free_surface_density(self, T, p):
       if self.variable_free_surface_density:
         return self.rho*(1-self.alpha*T)
```
Then add a line to the BaseApproximation doc string that says something along the lines of: "at free surface boundaries, a normal stress of g*(variable_free_surface_density(T, p) - external_density)*eta is applied, where eta is..."
* the prefactor thing in the free surface equation: I don't like that in the f.s. equation module it looks like it's applied only to the u.n term, and not the mass term - and more generally I don't like this mixing of timestepping details with equations. The whole point of treating the mass term and the "other residual terms" separately, is to avoid that - i.e these theta factors, RK weights etc. should be brought in where the timestepping is done, in this case in the Stokes solver, but I can see it's a little awkward. Yeah so either hack the stokes solver, e.g. you could do self.test[c+2] *= Constant(theta)` - or, add some general functionality for a prefactor in any equation, i.e. allow a factor to be specified in `BaseEquation` that gets multiplied into the residual of all terms.

stephankramer:
Oh, final point (for now ;-)): you don't seem to be using a $\tilde\eta=\Delta\rho g\eta$ as in the f.s. paper. Unless I'm missing something that means that your Stokes matrix is *not* symmetric, no? I mean if you look at eqn (23) of the paper: if you don't include the $\Delta\rho$ in the variable you are solving for, you need $\theta H\Delta\rho g$ in the top right corner  (which you get through the normal stress bc), but the bottom left corner remains $\theta H$. Come to think of it, I wonder why I didn't simply multiply the kinematic bc (bottom row) through by $\Delta\rho g$ - that way it would also be symmetric, but you could keep $\eta$ as the solution variable. I'm probably missing something...

Patol75:
@stephankramer I just wanted to quickly point out the implementation in the level set branch: the density method in the approximation relies on the buoyancy method, and that allows to couple seamlessly the free surface and multimaterial aspects of the simulation. That is what we used to replicate the drunken sailor case.

gadopt/stokes_integrators.py
```diff
@@ -100,35 +106,44 @@ class StokesSolver:
...
+
         self.mu = ensure_constant(mu)
         self.solver_parameters = solver_parameters
         self.J = J
         self.constant_jacobian = constant_jacobian
         self.linear = not depends_on(self.mu, self.solution)
 
         self.solver_kwargs = kwargs
-        u, p = fd.split(self.solution)
+        self.stokes_vars = fd.split(self.solution)  # For default stokes this is a tuple of (velocity, pressure)
```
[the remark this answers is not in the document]
Patol75: This way, `eta` becomes a list, right? That would require some additional changes here, but it may actually be quite handy. The logic could only fail if we added another variable I suppose, but I think it is a very nice suggestion indeed!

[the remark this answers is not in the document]
angus-g: Oh I forgot you could do that, I agree that's a clean way to do it!

Patol75:
@w3168 Given your last CI green tick was two months ago and that your recent commits should not affect things much, could it be that some Firedrake changes are responsible for current CI failures?

Patol75:
Just a little note from gh-35 that we could rename `rho_field` to `full_density`, and the associated returned expression should be `self.rho - self.buoyancy(p, T) / self.g`.

demos/analytical_comparisons/smooth_cylindrical_freesurface.py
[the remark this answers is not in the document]
w3168: Thanks Stephan I've added your summary above which is the logic I was going with for this test! As you say it would be good to derive a time dependent analytical free surface case at some point!

demos/analytical_comparisons/smooth_cylindrical_freesurface.py
```diff
@@ -0,0 +1,164 @@
...
+        stokes_solver.solve()
+        time.assign(time + dt)
+
+        # Calculate L2-norm of change in velocity:
+        maxchange = sqrt(assemble((u - u_old)**2 * dx))
+        log("maxchange = ", maxchange)
+
+        u_error = Function(V, name="VelocityError").assign(u_-u_anal)
+        p_error = Function(W, name="PressureError").assign(p_-p_anal)
+        eta_error = Function(W, name="EtaError").assign(eta_-eta_anal)
```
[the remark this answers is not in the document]
w3168: thanks stephan fixed now

demos/free_surface/Makefile
```diff
@@ -0,0 +1,15 @@
+cases := explicit implicit implicit_top_bottom
+parallel := implicit_top_bottom_buoyancy implicit_cylindrical
```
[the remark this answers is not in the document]
w3168: done

demos/free_surface/implicit_cylindrical_free_surface.py
```diff
@@ -0,0 +1,57 @@
...
+            self.bottom_id: {'un': 0}
+        }
+
+    def setup_nullspaces(self):
+        # Nullspaces and near-nullspaces:
+        self.Z_nullspace = create_stokes_nullspace(self.Z, closed=False, rotational=True)
+        self.Z_near_nullspace = create_stokes_nullspace(self.Z, closed=False, rotational=True, translations=[0, 1])
+
+    def update_analytical_free_surfaces(self):
+        self.eta_analytical.interpolate(exp(-self.time/self.tau0)*self.F0 * cos(self.number_of_lam * atan2(self.X[1], self.X[0])))  # Analytical free surface solution
```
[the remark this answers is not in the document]
w3168: This was another guess at an analytical solution for a cylinder. I've added a comment at the top of the run file: 

# N.b. the analytical solution in this test is a guess that the initial free
# surface topography on the cylinder (a cosine) should decay exponentially in
# a similar way to the cartesian cases. This seems to hold true for the set of
# wavenumbers/resolution used here.

gadopt/stokes_integrators.py
Patol75: @w3168 Just checking if this conversation should be resolved.

w3168: @Patol75  not yet! though stephan brought it up again so I think it should be resolved soon!

demos/free_surface/implicit_top_bottom_buoyancy_free_surface.py
```diff
@@ -0,0 +1,70 @@
...
+        self.G0 = 0
+        self.stokes_vars[2].assign(self.F0)  # initial top free surface condition
+        self.stokes_vars[3].assign(self.G0)  # initial bottom free surface condition
+        self.eta_analytical = Function(self.stokes_vars[2], name="eta analytical")
+        self.zeta_analytical = Function(self.stokes_vars[3], name="zeta analytical")
+
+    def update_analytical_free_surfaces(self):
+        # We use the more complicated analytical expression because the wavelength for this test case is the same
+        # as the depth of the domain, in order to minimise the spatial error without having to refine the grid too much.
+        # In earlier cases we had lambda = D/2 so the additional terms in the analytical expression had negligible effect.
```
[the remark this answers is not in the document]
w3168: thanks Stephan I've done that now in explicit_free_surface.py

```
+        # Equation A.4 from Kramer et al., 2012. Here we have a simplified form assuming that the relaxation time scale,
+        # tau = tau0 (see Equation A.7) which is valid for wavelengths << depth (e.g. see Table 2 from Kramer et al 2012).
```

demos/free_surface/implicit_top_bottom_free_surface.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __init__(self, dt_factor, **kwargs):
+        self.rho_bottom = 2
+        self.zeta_error = 0
+        super().__init__(dt_factor, **kwargs)
+
+        if self.iterative_2d:
+            # Schur complement splitting leads to a nullspace in the velocity block.
+            # Adding a small absorption term bringing the vertical velocity to zero removes this nullspace
+            # and does not effect convergence provided that this term is small compared with the overall numerical error.
```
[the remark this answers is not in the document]
w3168: done

demos/free_surface/implicit_top_bottom_buoyancy_free_surface.py
```diff
@@ -0,0 +1,70 @@
...
+    def initialise_free_surfaces(self):
+        self.F0 = 0
+        self.G0 = 0
+        self.stokes_vars[2].assign(self.F0)  # initial top free surface condition
+        self.stokes_vars[3].assign(self.G0)  # initial bottom free surface condition
+        self.eta_analytical = Function(self.stokes_vars[2], name="eta analytical")
+        self.zeta_analytical = Function(self.stokes_vars[3], name="zeta analytical")
+
+    def update_analytical_free_surfaces(self):
+        # We use the more complicated analytical expression because the wavelength for this test case is the same
```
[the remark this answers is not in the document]
w3168: Thanks - I've added equation numbers to the expressions below now!

demos/free_surface/implicit_top_bottom_free_surface.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def initialise_free_surfaces(self):
+        super().initialise_free_surfaces()
+        self.stokes_vars[3].interpolate(self.F0 * cos(self.kk * self.X[0]))  # Initial Bottom free surface condition
+        self.zeta_analytical = Function(self.stokes_vars[3], name="zeta analytical")
+
+    def setup_bcs(self):
+        super().setup_bcs()
+        # N.b. stokes_integrators assumes that the order of the bcs matches the order of the free surfaces defined in the
+        # mixed space, i.e. in this case top_id comes before bottom_id in the dictionary.
```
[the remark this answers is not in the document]
w3168: added your comment to the run file and to stokes integrators so that we might change it in future

gadopt/momentum_equation.py
```diff
@@ -115,6 +115,11 @@ def residual(self, test, trial, trial_lagged, fields, bcs):
                 # here we need only the third term, because we assume jump_u=0 (u_ext=u)
                 # the provided stress = n.(mu.stress_tensor)
                 F += dot(-phi, bc['stress']) * self.ds(id)
+            if 'normal_stress' in bc:
+                # add the free surface stress
```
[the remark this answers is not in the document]
w3168: done

gadopt/preconditioners.py
```diff
@@ -43,14 +44,26 @@ def initialize(self, pc):
         # Handle vector and tensor-valued spaces.
 
         # 1/mu goes into the inner product in case it varies spatially.
-        self.a = fd.inner(1/mu * u, v)*fd.dx
+        dx = context.appctx['dx']
```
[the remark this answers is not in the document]
w3168: changed to FreeSurfaceMassInvPC since the master branch has switched to MassInvPC since this message

gadopt/stokes_integrators.py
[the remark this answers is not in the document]
w3168: I've added this in now!

gadopt/stokes_integrators.py
w3168: changed now!

gadopt/approximations.py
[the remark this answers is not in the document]
w3168: as you said here https://github.com/g-adopt/g-adopt/pull/51#pullrequestreview-1986455985 i've moved the definition of free surface density into approximation, as well as the flag to turn it on/off, so that it is clear that this density is associated with the free surface external stress term

gadopt/stokes_integrators.py
```diff
@@ -140,14 +155,64 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
...
+                    self.free_surface = True
                 else:
                     weak_bc[bc_type] = value
             self.weak_bcs[id] = weak_bc
 
+        if self.free_surface:
+            if free_surface_dt is None:
+                raise TypeError("Please provide a timestep to advance the free surface, currently free_surface_dt=None.")
+
+            self.solution_old = fd.Function(self.solution)
```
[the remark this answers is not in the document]
w3168: Since this comment @Patol75  added a solution old too. We discussed it a little bit yesterday but I'm not 100 % sure why you have added it in Thomas?

gadopt/stokes_integrators.py
```diff
@@ -100,35 +106,44 @@ class StokesSolver:
     def __init__(self, z, T, approximation, bcs=None, mu=1,
                  quad_degree=6, cartesian=True, solver_parameters=None,
                  closed=True, rotational=False, J=None, constant_jacobian=False,
-                 **kwargs):
+                 iterative_2d=False, free_surface_dt=None, free_surface_variable_rho=True,
+                 free_surface_theta=0.5, **kwargs):
+
```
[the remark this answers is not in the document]
w3168: done

gadopt/stokes_integrators.py
```diff
@@ -140,14 +155,64 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
...
+                exterior_density = value.get('exterior_density', 0)
+
+                # Define free surface variables for timestepping
+                eta.append(self.stokes_vars[2+c])
+                eta_old.append(fd.split(self.solution_old)[2+c])
+                eta_theta.append((1-free_surface_theta)*eta_old[c] + free_surface_theta*eta[c])
+
+                # Add free surface equation
+                self.equations.append(FreeSurfaceEquation(self.Z.sub(2+c), self.Z.sub(2+c), quad_degree=quad_degree,
+                                      free_surface_id=id, free_surface_dt=free_surface_dt, theta=free_surface_theta, k=self.k))
```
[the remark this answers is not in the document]
w3168: oops that's right I've removed it now!

gadopt/free_surface_equation.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+class FreeSurfaceTerm(BaseTerm):
+    r"""
+    Free Surface term: u \dot n
+    """
+    def residual(self, test, trial, trial_lagged, fields, bcs):
+        free_surface_id = self.term_kwargs['free_surface_id']
+
+        # Multiply by constant factor to keep the block system symmetric for the implicit coupling case
```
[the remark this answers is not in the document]
w3168: As you mentioned above this is confusing. I've added a general prefactor to base equation so that it is specifed in as few places as possible

gadopt/stokes_integrators.py
```diff
@@ -140,14 +155,64 @@ def __init__(self, z, T, approximation, bcs=None, mu=1,
...
         self.F = 0
         for test, eq, u in zip(self.test, self.equations, fd.split(self.solution)):
             self.F -= eq.residual(test, u, u, self.fields, bcs=self.weak_bcs)
 
+        if self.free_surface:
+            for i in range(len(eta)):
+                # Add free surface time derivative term
+                # Multiply by theta to keep the block system symmetric for the implicit coupling case
+                # (N.b. we already have two equations from StokesEquations)
+                self.F += self.equations[2+i].mass_term(self.test[2+i], free_surface_theta*(eta[i]-eta_old[i])/free_surface_dt)
```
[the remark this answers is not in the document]
w3168: As above i've moved a general prefactor to the base equation (as well as the mass term of free surface) and then just specify the prefactor in stokes_integrators

w3168:
> I'd say nearly there - we should get it in as soon as possible! There's two little things that are nagging me:
> 
>     * the `rho_field` thing. Oh actually, can we just combine this with the variable_density logic, i.e. make that argument an argument of the approximation, not of the stokes solver, so something like:
> 
> 
> ```
> class BoussinesqApproximation(...)
>    ...
>    def __init__(self, ..., variable_free_surface_density=False)
>        ....
>        self.variable_free_surface_density = variable_free_surface_density
>        
>    def free_surface_density(self, T, p):
>        if self.variable_free_surface_density:
>          return self.rho*(1-self.alpha*T)
> ```
> 
> Then add a line to the BaseApproximation doc string that says something along the lines of: "at free surface boundaries, a normal stress of g*(variable_free_surface_density(T, p) - external_density)*eta is applied, where eta is..."
> 
>     * the prefactor thing in the free surface equation: I don't like that in the f.s. equation module it looks like it's applied only to the u.n term, and not the mass term - and more generally I don't like this mixing of timestepping details with equations. The whole point of treating the mass term and the "other residual terms" separately, is to avoid that - i.e these theta factors, RK weights etc. should be brought in where the timestepping is done, in this case in the Stokes solver, but I can see it's a little awkward. Yeah so either hack the stokes solver, e.g. you could do self.test[c+2] *= Constant(theta)`- or, add some general functionality for a prefactor in any equation, i.e. allow a factor to be specified in`BaseEquation` that gets multiplied into the residual of all terms.

Thanks @stephankramer for this review. Sorry I was so slow to get back to this. I think I've made most of the changes now. 

As you said it makes more sense to deal with the variable density logic and implementation in approximations as this way the free surface density is associated with a free surface term. 

I've also added a general prefactor to base equation (as well as redefining the mass term in free surface equation) so that all the terms in the equation can be multipled by the prefactor. 

Re the last comment about the delta rho g factor. I have tried now with -deltarho g theta multiplying the bottom equations and I think this is consistent with Equation 23 (without needing to rescale eta -> eta_tilde...). The iterations counts seem to improve especially for the two free surface case as the iteration counts roughly half. 

I have used delta_rho = approximation.rho - exterior_density (i.e. the constant rho from approximations, not the spatially variant free_surface_density) as I am not too sure what would happen if you try and scale the equations by a function. Perhaps this is the right thing to do though... Do you have any thoughts? 

Also I am a little unsure why @Patol75 introduced a solution_old in the level set branch. I agree that it would be cleaner to just have eta_old but I think there was a problem before about trying to assign eta (from the mixed function space) to a scalar function eta_old... Probably I should be able to get it to work though...

w3168:
I've just changed stokes solver so only eta_old needs updating at the end of the timestep and I've added a fix for the tosi benchmark in multimaterial. Hopefully this is all good to go now!

gadopt/equations.py
```diff
@@ -66,6 +69,13 @@ def __init__(
 
         self.dx = firedrake.dx(domain=self.mesh, degree=quad_degree)
 
+        # General prefactor multiplying all terms in the equation
+        # N.b. setting this to a firedrake constant (i.e. prefactor = Constant(1)) breaks
+        # Drucker-Prager rheology test even though it is being multiplied by 1...
```
stephankramer, line 74: Interesting!

stephankramer (approved):
Very nice! Very nicely integrated with the rest now, and some extensive testing.
One last suggestion (promise!) to subclass FreeSurfaceMassInvPC from MassInvcPC - but ignore if this can't be done easily

drhodrid:
Hey @w3168 - looks like you have the green light from Steph! Congrats! One thing from me, as we've discussed. Could you add a simple demo, showing how to set up a free surface BC on top of the base case, purely highlighting the changes required? I think this would necessitate another branch propagating from "base" in the diagram -- perhaps entitled "boundary conditions".

gadopt/stokes_integrators.py
```diff
@@ -224,19 +303,39 @@ def __init__(
             else:
                 self.solver_parameters.update(iterative_stokes_solver_parameters)
 
+            if self.solver_parameters.get("pc_type") == "fieldsplit":
+                # extra options for iterative solvers
```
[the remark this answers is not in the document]
Patol75: Yup, we figured it did not make sense not to have it consistent across all cases. :)

gadopt/equations.py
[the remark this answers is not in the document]
Patol75: Will made me notice that I had a similar issue with Drucker-Prager when I changed `buoyancy` in the level-set PR, and the quick fix was the same.

gadopt/stokes_integrators.py
```diff
@@ -224,19 +303,39 @@ def __init__(
...
         self._solver_setup = False
 
     def setup_solver(self):
         """Sets up the solver."""
         # mu used in MassInvPC:
-        mu_over_rho = self.mu / self.approximation.rho_continuity()
+        appctx = {"mu": self.mu / self.approximation.rho_continuity()}
+
+        if self.free_surface:
+            appctx["dx"] = self.equations[0].dx
```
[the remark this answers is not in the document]
w3168: thanks I've removed this now!

w3168:
Thansks @stephankramer for the review again! I've simplified the FreeSurfaceMassInvPC so that it is a subclass of MassInvPC overloading the form() method. Hopefully this is all good now! 

@drhodrid I'm going to create a demo now on another branch building off the base case.

Patol75:
After two days of intense efforts, the adventurers of the free surface ark are hesitant to propose a new demo. Feedback appreciated. Discussions needed. Rest required.

stephankramer (approved):
> I've simplified the FreeSurfaceMassInvPC so that it is a subclass of MassInvPC overloading the form() method. Hopefully this is all good now!

Nice, much cleaner

demos/free_surface/free_surface.py
```diff
@@ -0,0 +1,355 @@
...
+    # Centre camera on xy plane and write frame
+    plotter.camera_position = "xy"
+    plotter.write_frame()
+    plotter.clear()
+
+# Closes and finalizes movie
+plotter.close()
+
+
+# -
```
Patol75, line 271: Should this section have the `"active-ipynb"` tag? It is currently generating a CI execution error, as `pyvista` and `matplotlib` are not imported.

Patol75 (changes requested):

drhodrid (approved):
I think aside from @Patol75's comment on the demo, this is now probably good to go.

demos/free_surface/free_surface.py
```diff
@@ -0,0 +1,355 @@
...
+# \end{equation}
+# where $\star$ indicates a nondimensional variable. Rearranging, our expected dynamic topography at steady state would be
+# \begin{equation}
+#      \eta_{\star} = \dfrac{-p_{\star}'  + 2 \frac{\partial v_{\star}}{\partial y_{\star} }}{ Ra T_{\star}- Ra_{FS} }.
+# \end{equation}
+
+# First let's load the final state from the base case and calculate the expected dynamic topography with the code below.
+
+# + tags=["active-ipynb"]
+# # load velocity and pressure from final state of the base case run (we have saved earlier it earlier!).
```
angus-g, line 304:
```suggestion
# # load velocity and pressure from final state of the base case run (we have saved it earlier!).
```

Does this also mean we don't need the `Final_State_base.h5` file committed?

demos/free_surface/free_surface.py
```diff
@@ -0,0 +1,355 @@
...
+
+with CheckpointFile("Final_State.h5", "w") as final_checkpoint:
+    final_checkpoint.save_mesh(mesh)
+    final_checkpoint.save_function(T, name="Temperature")
+    final_checkpoint.save_function(z, name="Stokes")
+# -
+
+# Let's use the python package **PyVista** to plot the temperature field through time. We will use the calculated free surface height to artifically scale the mesh in the vertical direction. We have exaggerated the vertical stretching by a factor of 1.5, **BUT...** it is important to remember this is just for ease of visualisation - the mesh is not moving in reality!
+
+# +
```
angus-g, line 229:
```suggestion
# + tags=["active-ipynb"]
```

demos/free_surface/free_surface.py
w3168: Thanks for picking this up @Patol75 also as you mentioned offline I've added imageio to the demos dependency list for gadopt in the pytoml.

demos/free_surface/free_surface.py
Patol75: I think that is a yes. I do not have it on my branch, and that does not cause an issue.

w3168: I think we do need the file because this is specifically for plotting a previous base case run to compare with the free surface run at steady state.

demos/free_surface/free_surface.py
w3168: thanks done that now!

demos/free_surface/free_surface.py
Patol75: Oh, then that may explain why I keep getting a run failure despite the test success! I will try adding it in to see.

w3168:
I've updated some last plotting bits, including adding the gif file of temperature and free surface evolution. Is it the right place to put the file? It's 1.6Mb


@angus-g  can we check if this works ok on the website? I guess it would make sense to put it in geometry and boundary condition block

angus-g:
> Is it the right place to put the file? It's 1.6M

No, the notebook should generate and use the gif itself, instead of embedding in the repository. I can help with this.


> @angus-g  can we check if this works ok on the website? I guess it would make sense to put it in geometry and boundary condition block

I'll put the bits in place for this

demos/free_surface/free_surface.py
Patol75: Can confirm everything ran successfully.

Patol75 (approved):
Happy to have this in :)

angus-g:
I’ll just make sure that all the new demo stuff is working as expected and a quick proofread, then we should be good to go!

PR #52, opened 2024-02-05
Defer to PyPI for pyroltrilinos package

angus-g:
Now that this is being built for most platforms, we should use the packaged wheel to make sure that's working, and remove the need for a fairly lengthy build in the testing workflow. This should also make things simpler for upstream installs (i.e. `pip install gadopt[optimisation]` should do what you want).

I've also split the gadopt specification: installing bare gadopt **won't** include pyrol, or assess/pandas. These packages are pulled in by the *optimisation* and *testing* variants, respectively. So in the installation for testing, we now install `gadopt[testing,optimisation]`. Is this a sensible way to split things up?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,33 +56,10 @@ jobs:
           tsp -S 48
       - uses: actions/checkout@v4
 
-      - name: Load pyrol wheel from cache
-        uses: actions/cache/restore@v3
-        id: cache-pyrol
-        with:
-          path: /tmp/wheels
-          key: pyrol
-
-      - name: Build pyrol wheel
-        if: steps.cache-pyrol.outputs.cache-hit != 'true'
-        run: |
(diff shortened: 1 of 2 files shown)


drhodrid:
Hi @angus-g - thanks for this. I will defer to somebody else to review the coding aspects, but your split (i.e. default, testing and optimisation) makes perfect sense to me. That's exactly how I would have done it. 

I'm wondering if this is also a model we should eventually use for pygplates... but something we can discuss at the dev meeting this week.

sghelichkhani (approved):
Looks good to me! 
The split makes perfect sense. We probably don't need to install pandas for our actual applications. 
@angus-g can correct me, but unless pygplates provides a PyPI package in the near future (which does not seem likely based on how the package is released right now) this will not be how we install pygplates. For our pygplates demos, it is safe to assume that we are dealing with a user that knows how to install pygplates on an HPC system in a virtual environment.

PR #53, opened 2024-02-05
Changing Interpolators to be consistent with future Firedrake changes

sghelichkhani:
This PR changes the way interpolation is used for LayerAveraging and Adaptive time-stepping to be consistent with future Firedrake changes and avoid raising long weird warnings.


angus-g (approved):
Looks alright, although I do wonder how we'll know to switch back from importing from `__future__`...

sghelichkhani:
That's a good point. I assume that will be another lengthy warning that we should deal with at some point? :-D

PR #57, opened 2024-02-22
Format import style 

Patol75:
I am trying to see if there is some merit to defining a consistent way in which packages are imported in G-ADOPT.

The current situation is quite heterogeneous, and there are some intricate relationships. For example,  `demos/optimisation_checkpointing/helmholtz.py` calls `Control`, which is imported using `from gadopt.inverse import *`, in which one can find `from firedrake.adjoint import *`. This is obscure at best.

The way Firedrake classes and functions are imported into G-ADOPT seems to vary from module to module. Having a defined style would help to tidy the code. I can see some constraints to reorganising imports:
- `from module import *` is not considered good practice in Python, given potential namespace clashes and the ambiguous origin of objects used in the code
- `from module import Class, function, object` is fine, but it forces the programmer to always scroll up and insert the new object they need
- the beauty of UFL lies in the fact that Python code can be read as if it were mathematics; using `fd.grad(fd.dot(a, b))` loses this magic, and I can already hear people complaining about it

With this in mind, I have tried the following approach:
- `import firedrake as fd` and use `fd.` to refer to Classes
- `from firedrake import functions` to refer to any object using lowercase naming

What are your thoughts on this matter?

### DIFF
--- a/gadopt/__init__.py
+++ b/gadopt/__init__.py
@@ -1,14 +1,20 @@
 from firedrake import *
-from .time_stepper import CrankNicolsonRK, ImplicitMidpoint
-from .limiter import VertexBasedP1DGLimiter
-from .utility import log, ParameterLog, TimestepAdaptor, LayerAveraging, timer_decorator
+
+from .approximations import (
+    BoussinesqApproximation,
+    ExtendedBoussinesqApproximation,
+    AnelasticLiquidApproximation,
+    TruncatedAnelasticLiquidApproximation,
+)
 from .diagnostics import GeodynamicalDiagnostics
+from .energy_solver import EnergySolver
+from .limiter import VertexBasedP1DGLimiter
 from .momentum_equation import StokesEquations
+from .preconditioners import SPDAssembledPC, VariableMassInvPC
 from .scalar_equation import EnergyEquation
 from .stokes_integrators import StokesSolver, create_stokes_nullspace
-from .energy_solver import EnergySolver
-from .approximations import BoussinesqApproximation, ExtendedBoussinesqApproximation, AnelasticLiquidApproximation, TruncatedAnelasticLiquidApproximation
-from .preconditioners import SPDAssembledPC, VariableMassInvPC
+from .time_stepper import CrankNicolsonRK, ImplicitMidpoint
+from .utility import log, ParameterLog, TimestepAdaptor, LayerAveraging, timer_decorator
 
-from firedrake.petsc import PETSc
 PETSc.Sys.popErrorHandler()
+
--- a/gadopt/approximations.py
+++ b/gadopt/approximations.py
@@ -1,5 +1,8 @@
 import abc
-from firedrake import sym, grad, inner, div, Identity
+
+import firedrake as fd
+from firedrake import div, grad, inner, sym
+
 from .utility import ensure_constant, vertical_component
 
 
@@ -133,7 +136,7 @@ class ExtendedBoussinesqApproximation(BoussinesqApproximation):
     def viscous_dissipation(self, u):
         stress = 2 * self.mu * sym(grad(u))
         if self.compressible:  # (used in AnelasticLiquidApproximations below)
-            stress -= 2/3 * self.mu * div(u) * Identity(u.ufl_shape[0])
+            stress -= 2/3 * self.mu * div(u) * fd.Identity(u.ufl_shape[0])
         phi = inner(stress, grad(u))
         return phi * self.Di / self.Ra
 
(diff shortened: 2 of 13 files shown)


sghelichkhani:
Thanks for the discussion @Patol75. While I understand your issue with wildcard importing, and all the merits associated with clear definition of where each class/function is imported from, we should note a couple of issues: First, doing adjoints, many operations, such as taping, are global operations. So as far as my understanding goes, firedrake.inverse and  our gadopt.inverse functionalities use global parameters that require a wildcard import.

Second issue, which is just more justifying why wildcard import is OK (not necessarily better), is that firedrake is not just a module that we use in conjunction with others, the same way we use numpy/scipy. It is an environment that is finely tweaked so that the whole machinery, including petsc, and mpi are working finely together. So it is safe to assume that any user  in this environment is aware of this, and their primary goal is to use firedrake. In the same way, gadopt is a firedrake application.  So all the issues with wildcard importing are less severe than usual. 

That said, other than the issue with adjoint, I don't see any problem with your proposition. Keen to see if you have a way around that.

Patol75:
First issue: do you mean that some objects need to be imported even though they are not explicitly referred to in the code?

Second issue: does that mean that we could have `from firedrake import *` in all `gadopt` modules and then have `from gadopt import *` in all scripts that use G-ADOPT (i.e. any simulation)?

sghelichkhani:
I think the answer the both is yes! But I need to double check with a min reproducer once I am back in the office. 👶

Patol75:
Having just checked what Firedrake does in their tests and Thetis in their sources, it seems that `import *` is what everybody does. In that case, would we be happy to just go ahead with that? Additionally, we could define `__all__` in every `gadopt` module to be explicit about what is imported.

stephankramer:
Just to summarize my thoughts:
* as I mentioned in the meeting I think we should make a distinction between what happens in gadopt modules (everything under gadopt/) and demos and tests
* the reason (for better or for worse) firedrake goes for the "from firedrake import *" is that they indeed want to provide a fully functional FEM platform with all whistles and bells as a user experience. As @sghelichkhani says we can expect a large part of our users to be familiar with firedrake who would probably mostly still use "from firedrake import *" even if we never did so in any of our demos. So in our demos which should reflect how we intend people to use gadopt, but also how we reasonably expect people to use it I think sticking with that model is probably the best. The model we've been following so far is the same as Thetis indeed, where `from gadopt import *` also exposes all of `from firedrake import *` - this is nice in the sense that it gives a kind of unified gadopt-as-a-platform feel, where users don't have to worry about which bits comes from firedrake or from gadopt.
* tests in thetis and firedrake mostly test the user interface, so there it makes sense to use the same paradigm as the user would. There are also unit tests (the dividing line isn't always that clear) that tests bits that are not directly exposed to the user, so you'll find some "from firedrake.submodule import X, Y" as well. gadopt doesn't currently have many "pure" tests (i.e. not written for demonstration purposes) instead we mostly rely on closely checking changes in demo outputs.
* the adjoint is a bit of a historic artefact. It used to be the case indeed that you had to do "from firedrake import *" *followed* by "from firedrake_adjoint import *" where the latter would expose wrapped versions of Function, Constant, etc. and also automatically switch on the tape. Now however these firedrake classes are always wrapped, the wrapper just doesn't do anything when the tape is not on. It's still the case that "from firedrake_adjoint import *" switches on the tape, but it produces a warning and you're now encouraged to do `from firedrake.adjoint import *` and explicitly switch on the tape yourself with `continue_annotation`. `firdrake.adjoint` exports additional bits from pyadjoint that are useful for inversion: `Control, ReducedFunctional` etc. `gadopt.inverse` mimicks this. But if you don't want to use `from X import *` there's nothing in the adjoint stopping you doing that anymore

What we do inside gadopt/ is I think a separate choice and there we have been strict to not use `from X import *` except with the explicit purpose to export firedrake symbols.  But there's still a choice indeed between `import firedrake as fd` and `from firedrake import X, Y, Z`. Unfortunately the split you @Patol75 propose between functions and classes is not as clear cut. Most lower case UFL "functions" are indeed python functions, but they are really just thin wrappers that create a UFL term of the corresponding class, i.e. `grad(u)` produces an object of type `Grad`, etc. And the capitalisation is really not consistent (this is all historic backward compatibility stuff), e.g. `dx` is actually directly a class (it's just an alias for `Measure`), reversely `Mesh` is actually a function that returns a `MeshGeometry` object, `And` is a python function that returns an `AndCondition` object, etc. etc. So if you want to be consistent I think you just have to make choice between `import as fd` or `import X, Y, Z` - they both have their drawbacks - but to be honest don't really see a strong advantage of being consistent there.

Patol75:
Thanks, Steph. I feel like I have not been used to working in this kind of environment, but now I understand better the logic and the approach chosen. I will close this PR accordingly.

PR #60, opened 2024-04-03
Remove VariableMassInvPC

stephankramer:
The main reason to have our own version of MassInvPC was that MassInvPC did not update for varying (between solves) viscosities. This has been fixed in https://github.com/firedrakeproject/firedrake/pull/3372 and since then the options structure for the solvers beneath it has changed: need to use pcksp with an additional ksp_ layer now. Thus VariableMassInvPC is no longer a drop in-replacement for MassInvPC as it was intended. Better to stay consistent with Firedrake therefore.

### DIFF
--- a/demos/Drucker-Prager_rheology/spiegelman.py
+++ b/demos/Drucker-Prager_rheology/spiegelman.py
@@ -91,9 +91,10 @@ def spiegelman(U0, mu1, nx, ny, picard_iterations, stabilisation=False):
             "ksp_type": "preonly",
             "ksp_converged_reason": None,
             "pc_type": "python",
-            "pc_python_type": "gadopt.VariableMassInvPC",
-            "Mp_ksp_type": "cg",
-            "Mp_pc_type": "sor",
+            "pc_python_type": "gadopt.MassInvPC",
+            "Mp_pc_type": "ksp",
+            "Mp_ksp_ksp_type": "cg",
+            "Mp_ksp_pc_type": "sor",
         }
     }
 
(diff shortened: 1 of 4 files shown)


stephankramer:
Turns out there was another we reason we have separate version of MassInvPC, see
https://github.com/g-adopt/g-adopt/pull/43

Wondering whether to ask this to be changed in firedrake...

drhodrid:
I think it's best to get this changed in Firedrake indeed - or at least ask... That way we minimise code duplication, and you've always told me that's a good thing :-)

stephankramer:
See https://github.com/firedrakeproject/firedrake/pull/3488

angus-g:
Now that the upstream PR has landed, is this ready to go?

demos/Drucker-Prager_rheology/spiegelman.py
```diff
@@ -91,9 +91,10 @@ def spiegelman(U0, mu1, nx, ny, picard_iterations, stabilisation=False):
             "ksp_type": "preonly",
             "ksp_converged_reason": None,
             "pc_type": "python",
-            "pc_python_type": "gadopt.VariableMassInvPC",
-            "Mp_ksp_type": "cg",
-            "Mp_pc_type": "sor",
+            "pc_python_type": "gadopt.MassInvPC",
```
stephankramer, line 94:
```suggestion
            "pc_python_type": "firedrake.MassInvPC",
```

stephankramer:
Ah yes - and I see it's passing now as well, nice
Just spotted one minor thing I've just fixed - but when it goes green again, yes ready for review.
This was basically also my attempt to address the doc strings in gadopt/preconditioners.py  -there's not much left now, but will probably grow again soon

angus-g (approved):

PR #66, opened 2024-04-25
ENH: Simplify time stepper classes

Patol75:
Except if I am missing something, there should not be a reason to have each time stepper class inherit from an intermediate abstract class. This PR simplifies current class definitions and also includes small housekeeping items (sorry, I cannot help it).

### DIFF
--- a/gadopt/time_stepper.py
+++ b/gadopt/time_stepper.py
@@ -1,6 +1,6 @@
 """Timestepper implementation, mostly copied from Thetis."""
 import operator
-from abc import ABC, abstractmethod, abstractproperty
+from abc import ABC, abstractmethod
 from numbers import Number
 from typing import Any, Optional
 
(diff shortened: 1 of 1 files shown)


angus-g (approved):
Seems sensible, although I would like to make sure that @stephankramer didn't do it for some particular reason?

drhodrid:
From what I remember, a lot of our time-stepping was motivated by what's in thetis or thwaites. Perhaps @w3168 may also have an idea of whether or not there was a reason for the previous configuration.

stephankramer:
The timestepper code is mostly just copied from Thetis, and written by Thomas Karna. I believe the current structure is to allow combining different implementations of either DIRK or ERK. Have a look at https://github.com/thetisproject/thetis/blob/master/thetis/rungekutta.py. There's a DIRKGeneric and a DIRKGenericUForm and these are combined withe the same DIRK33Abstract to form DIRK33 and DIRK33UForm. Similarly there's a ERKGeneric and a ERKGenericShuOsher. So these are mathematically the same timestepping scheme, but algorithmically follow a different approach - where I presume there's efficiency differences for different cases/eqns/solver strategies - tbh I'm a little vague on the details here.

My quick-and-dirty approach so far has been to to just copy what we needed and try to not let the two (Thetis and gadopt's version of it) diverge too much. Ideally you'd split that timestepping code out as a separate abstraction layer, but then this is also what I believe IRKSome offers, so if that could do everything we need that would be better. Either way would require some effort that's probably better warranted when we have a better picture of variations we might need.

Patol75:
Fair, I will close for now. From what I remember (not clear if reliable haha) when I read about time-integration schemes, the Shu-Osher form was mentioned as more convenient mathematically to express SSP schemes, but I do not have a memory of any computational-efficiency discussion.

PR #68, opened 2024-05-02
Pressure nullspace

sghelichkhani:
Adding the right nullspace case for anelastic liquid approximation.


demos/2d_compressible_ALA/ALA_p_nullspace.py
```diff
@@ -0,0 +1,33 @@
+from gadopt import *
```
stephankramer, line 1: Could you add a comment on top to say what this script does. I mean, I can see it, and it looks useful, but just for people looking at a demo, and quickly want to understand which python file does what... Actually maybe name it 'visualise_ALA_p_nullspace.py' ?

stephankramer (changes requested):
This looks great. Did you see any changes in solver behaviour. Oh, I guess this is just using a direct solve, and I'm not entirely sure that actually uses the right nullspace - hm yeah, so we should probably test whether our ideas about what the nullspace should be are correct. So maybe try with iterative solvers and add
  "mat_null_space_test_view": None,
  "ksp_test_null_space": None
inside the fieldsplit_1 dictionary. This makes petsc test the nullspace and print out whether it thinks indeed the provided nullspace is a nullspace of the operator (in this case the Schur complement).

stephankramer:
I'm also thinking, if we want to make this future proof for other approximations with a pressure dependent buoyancy term, we could change ALA to something like:

   def dbuoyancydp(self, p, T):
       """Pressure derivative of buoyancy term"""
       return    -self.Di * self.cp0 / self.cv0 / self.gamma0 * self.g * self.rho * self.chi

    def buoyancy(self, p, T):
        pressure_part = self.dbuoyancydp(self, p, T)  * p
        temperature_part = super().buoyancy(p, T)
        return pressure_part + temperature_part

and use dbuoyancydp() in the nullspace instead of having that `-self.Di * self.cp0 / self.cv0 / self.gamma0 * self.g * self.rho * self.chi` defined in two places.

demos/2d_compressible_ALA/ALA_p_nullspace.py
sghelichkhani: I changed it to a demo completely just in case it is useful for anyone, and have copied the math I used as a discussion there.

sghelichkhani:
> I'm also thinking, if we want to make this future proof for other approximations with a pressure dependent buoyancy term, [...]
> and use dbuoyancydp() in the nullspace instead of having that. 

Done! Good suggestion.

sghelichkhani:
Thanks @stephankramer. I passed in the flags for nullspace and it seems to work beautifully. Using the flags:
 ```iterative_stokes_solver_parameters["mat_null_space_test_view"] = None
iterative_stokes_solver_parameters["ksp_test_null_space"] = None
iterative_stokes_solver_parameters["fieldsplit_0"]["ksp_rtol"] = 1e-9
iterative_stokes_solver_parameters["fieldsplit_0"]["ksp_converged_reason"] = None
iterative_stokes_solver_parameters["fieldsplit_1"]["ksp_converged_reason"] = None
iterative_stokes_solver_parameters["fieldsplit_1"]["ksp_rtol"] = 1e-8
```
I get the `Null vector 0 is likely null vector || A * v[0] || = 2.72263e-08`. So we are calculating the nullspace correctly. 

In terms of improvement in convergence, we do have minor improvements. Often reducing the number of velocity solves by 1, and sometimes reducing the number of pressure iterations by 1. So not significant, but still visible.

**Comment by sghelichkhani** [comment]
@angus-g I have added the demo for this to the Makefile. Is it all one needs to do?

**Comment by angus-g** [comment]
> @angus-g I have added the demo for this to the Makefile. Is it all one needs to do?

No, the current setup expects the demos to have the format `demo_name/demo_name.py`, so it won't work for this case since you've put it in the `2d_compressible_ALA` directory. I'll fix that up

File: gadopt/stokes_integrators.py [hunk 2]
```diff
@@ -71,11 +73,16 @@ def create_stokes_nullspace(
       closed: Whether to include a constant pressure null space
       rotational: Whether to include all rotational modes
       translations: List of translations to include
+      ala_approximation: AnelasticLiquidApproximation for calculating right nullspace
 
     Returns:
       A Firedrake mixed vector space basis incorporating the null space components
 
     """
+    # ala_approximation and top_subdomaion_id are both needed when calculating rigt nullspace for ala
```
drhodrid, line 82: rig**h**t

drhodrid (commented):
Very nice. Only s minor typo from me. 

The changes to the demo are fine I think and I'm ok to merge. However, post merge, I think we need to change the order of TALA and ALA in the demos (i.e. base --> TALA --> ALA rather than base --> ALA --> TALA ) because this nullspace adds a complexity that is not in TALA and requires explanation. I'm happy to do this after this one goes in. I'd also appreciate @stephankramer having a quick look at this PR if possible, to ensure his suggestions have been addressed.

stephankramer (changes requested):

stephankramer:
As discussed off-line, I've addressed (my own :-)) PR comments - as @sghelichkhani might not have time in the near future, and fixed some of the typesetting in the demo. @drhodrid - could you read re-review - also to make sure I haven't messed anything up in the rebase, as I had quite a lot of conflicts.

@angus-g - could you indeed have a look to see how to best deal with the 2d_compressible_ALA/visualise_ALA_p_nullspace demo in the Makefile (or even just move it into its own dir if you prefer). Btw is there a way we can reference (link) other demos in a demo?

angus-g:
> Btw is there a way we can reference (link) other demos in a demo?

I think a link to, e.g. `/tutorials/2d_compressible_TALA` is sufficient.

gadopt/stokes_integrators.py
```diff
@@ -265,3 +277,38 @@ def solve(self):
...
+            $$ - \nabla p + g * Di * \rho * \chi * cp/(cv * \gamma) * \hat{k} * p = 0 $$
+        Taking the divergence:
+            $$ - \nabla \cdot(\nabla p) + \nabla \cdot(g * Di * \rho * \chi * cp/(cv * \gamma) * \hat{k} * p) $$
+        And then testing it with q:
+            $$ - q * \nabla \cdot(\nabla p) * dx + q * \nabla \cdot(g * Di * \rho * \chi * cp/(cv * \gamma) * \hat{k} * p) * dx $$
+        Doing integration by parts:
+            $$ - \nabla \cdot( q * \nabla(p)) * dx + \nabla(q). \nabla(p) * dx + \nabla \cdot(q * g * Di * \rho * \chi * cp/(cv * \gamma) * \hat{k} * p) * dx - \nabla(q) . (g * Di * \rho * \chi * cp/(cv * \gamma) * \hat{k} * p) * dx $$
+
+        Dropping the boundary condition terms:
+        $$ \nabla(q) \cdot \nabla(p) * dx - \nabla q \cdot (g * Di * \rho * \chi * cp/(cv * \gamma) * \hat{k} * p) * dx $$
```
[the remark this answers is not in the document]
angus-g: I've changed this to asciimath (consistent with our other docstrings), so hopefully it's a bit more readable in plaintext, but also renders to math.

drhodrid (approved):
This looks good to go. As mentioned, I will reverse order of, and update, TALA and ALA demos in a subsequent merge, hopefully next week.

PR #69, opened 2024-05-06
Smoothing of a scalar function using DiffusiveSmoothingSolver

sghelichkhani:
This PR adds DiffusiveSmoothingSolver: it solves the diffusion equation using a backward Euler method. This can be used for comparisons with data, and potentially having smoother fields.


demos/smoothing/demo.py
```diff
@@ -0,0 +1,69 @@
...
+# In isotropic smoothing, we assume that the diffusion coefficient is the same in all directions.
+# This simplifies the diffusion tensor to a scalar value, promoting uniform smoothing across all spatial directions.
+smooth_solution_isotropic = Function(T.function_space(), name="Smooth Temperature - Isotropic")
+smoother_isotropic = DiffusiveSmoothingSolver(
+    function_space=T.function_space(),
+    wavelength=0.1,  # Smoothing duration
+    bcs=temp_bcs)
+
+smooth_solution_isotropic.assign(smoother_isotropic.action(T))
+VTKFile("isotropic_smoothing.pvd").write(T.assign(T - T_avg), smooth_solution_isotropic.assign(smooth_solution_isotropic - T_avg))
```
angus-g, line 39: Does the `T.assign(T - T_avg)` here affect `T` for the following anisotropic section?

gadopt/utility.py
```diff
@@ -302,6 +307,113 @@ def get_functionspace(mesh, h_family, h_degree, v_family=None, v_degree=None,
...
+            mesh (firedrake.Mesh): The mesh to solve the problem on.
+            delta_t (Number): The wavelength or time duration for diffusion.
+            bcs (Optional[dict[int, dict[str, int | float]]]): Boundary conditions to impose on the solution.
+            quad_degree (int): Quadrature degree for integrations.
+            solver_parameters (Optional[dict[str, Union[str, float]]]): Solver parameters for the solver.
+            K (Union[firedrake.Function, Number]): Diffusion tensor.
+            solver_kwargs (dict): Additional keyword arguments to pass to the solver.
+            _solver_is_setup (bool): Flag indicating if the solver is set up.
+
+        """
```
angus-g, line 347: This docstring should go under the class itself (not in `__init__`), and the `Args` section doesn't need to replicate the types: you've already annotated them in the function definition itself!

If you don't expect the attributes to be directly accessed (which I don't think you do), I don't think they need to be in the docstring like this either.

gadopt/utility.py
```diff
@@ -302,6 +307,113 @@ def get_functionspace(mesh, h_family, h_degree, v_family=None, v_degree=None,
...
+            firedrake.Function: The smoothed field.
+
+        """
+        if not self._solver_is_setup:
+            self._setup_smoothing_solver()
+
+        # TODO: make sure you change it to interpolate
+        self.rhs.assign(T)
+        self.solver.solve()
+
```
angus-g, line 376: Does this TODO need to be done?

gadopt/utility.py
```diff
@@ -302,6 +307,113 @@ def get_functionspace(mesh, h_family, h_degree, v_family=None, v_degree=None,
...
+        self.solver_parameters = solver_parameters  # solver parameters to pass on to solve
+        self.K = K  # diffusion tensor, default is 1
+        self.solver_kwargs = kwargs  # optional parameters to pass to solver
+        self._solver_is_setup = False
+
+    def action(self, T: fd.Function):
+        """Apply smoothing action.
+
+        Args:
+            T (firedrake.Function): The input field to be smoothed.
```
angus-g, line 364:
```suggestion
            T: The input field to be smoothed.
```

gadopt/utility.py
```diff
@@ -302,6 +307,113 @@ def get_functionspace(mesh, h_family, h_degree, v_family=None, v_degree=None,
...
+                return fd.ds(subdomain_id)
+        bcs = []
+        # for boundary conditions set, we have flux at the boundaries
+        for subdomain_id, bc in self.bcs.items():
+            for bc_type, value in bc.items():
+                if bc_type == 'T':
+                    bcs.append(fd.DirichletBC(self.function_space, value, subdomain_id))
+                    a -= self.delta_t * test * fd.inner(fd.grad(trial), fd.FacetNormal(self.mesh)) * ds(subdomain_id)
+                else:
+                    raise ValueError("Boundary conditions other that Dirichlet are not supported.")
```
angus-g, line 407:
```suggestion
                    raise ValueError("Boundary conditions other than Dirichlet are not supported.")
```

demos/smoothing/demo.py
Patol75: I think you are right, this should impact line 66.

gadopt/utility.py
```diff
@@ -302,6 +307,113 @@ def get_functionspace(mesh, h_family, h_degree, v_family=None, v_degree=None,
     return constructor(mesh, elt, **kwargs)
 
 
+class DiffusiveSmoothingSolver:
+    def __init__(
+            self,
+            function_space: fd.FunctionSpace,
+            wavelength: Number,
+            bcs: Optional[dict[int, dict[str, int | float]]] = None,
```
Patol75, line 315: `int | float` can be replaced by `Number`. I think I introduced the former as part of documentation changes, but Angus rightly pointed the existence of `Number`.

tests/test_smoothing.py
```diff
@@ -0,0 +1,75 @@
...
+
+    smoother = DiffusiveSmoothingSolver(
+        function_space=T.function_space(),
+        wavelength=.1,
+        bcs=temp_bcs)
+
+    smooth_solution.assign(smoother.action(T))
+    expected_value = 0.06243541054145882
+    result = assemble((smooth_solution - T) ** 2 * dx)
+    assert np.isclose(result, expected_value, rtol=1e-5), f"Expected {expected_value}, got {result}"
```
Patol75, line 45: You may want to use [assert_allclose](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_allclose.html).

Patol75:
For future reference: the proposed functionality seems to be compatible with [`GenericTransportSolver`](https://github.com/g-adopt/g-adopt/blob/fc31b3fffe5d8b14c97c299bb9f4559b1aef768f/gadopt/transport_solver.py#L251) and should use it if that is the case.

drhodrid:
@Patol75 and @sghelichkhani to chat about this one in a few weeks, to determine if it is best to integrate with generic transport solver, or whether to go ahead as a standalone.

drhodrid:
Interesting to see how it runs with GenericTransportSolver.

That said, committing directly to someone else’s pull request isn’t usually best practice unless agreed in advance. It’s generally better to suggest changes via a comment or a separate branch, so the original author can review and incorporate them as they see fit.

We should definitely discuss the benefits and drawbacks of both solver approaches further, but as noted, let’s park that for a few weeks until we’ve worked through the priorities from yesterday’s dev meeting.

sghelichkhani:
@Patol75 Could you please have a go at this? I rewrote it to accommodate for your suggestion of using transport_solver.py

demos/makefile.cases.inc
```diff
@@ -17,11 +17,11 @@ gia_demos := $(addprefix $(gia_dir)/,$(gia_demo_names))
 gia_cases := $(addprefix $(gia_dir)/,$(gia_test_names))
 
 # other cases not tied to specific physics/domains
-extra_cases := PDE_constrained_optimisation
+extra_cases := PDE_constrained_optimisation smoothing
```
Patol75: Should we create a directory called `extras` to group these demos not tied to the main G-ADOPT categories?

demos/smoothing/smoothing.py
```diff
@@ -0,0 +1,255 @@
...
+# 2. **Anisotropic smoothing**: directionally-dependent diffusion
+#
+# The checkpoint file used in this tutorial contains a cylindrical temperature field
+# from a 2D adjoint simulation. We use the same checkpoint file as an adjoint simulation
+# to demonstrate smoothing techniques on a realistic geodynamic field.
+# To download the checkpoint file if it doesn't already exist, execute the following command:
+
+# + tags=["active-ipynb"]
+# ![ ! -f adjoint-demo-checkpoint-state.h5 ] && wget https://data.gadopt.org/demos/adjoint-demo-checkpoint-state.h5
+# -
```
Patol75: Does this block match the fact that we are downloading a file called `smoothing-example.h5`?

demos/smoothing/smoothing.py
```diff
@@ -0,0 +1,255 @@
...
+
+# How to do smoothing  in G-ADOPT
```
Patol75: Extra white space

demos/smoothing/smoothing.py
```diff
@@ -0,0 +1,255 @@
...
+# This tutorial has demonstrated how to:
+# - Load temperature fields from checkpoint files
+# - Apply isotropic smoothing with uniform diffusion
+# - Apply anisotropic smoothing with directional diffusion control
+# - Compare the effects of different smoothing approaches
+# - Visualise and quantify smoothing results
+#
+# The key difference between isotropic and anisotropic smoothing is evident:
+# - Isotropic smoothing reduces variations uniformly in all directions
+# - Anisotropic smoothing preserves radial structure while smoothing tangentially
```
Patol75: Perhaps make this statement general and not specific to the current demo?

gadopt/transport_solver.py
```diff
@@ -17,12 +17,13 @@
 from .approximations import BaseApproximation
 from .equations import Equation
 from .solver_options_manager import SolverConfigurationMixin, ConfigType
-from .time_stepper import RungeKuttaTimeIntegrator
-from .utility import DEBUG, INFO, absv, is_continuous, log, log_level
+from .time_stepper import RungeKuttaTimeIntegrator, BackwardEuler
+from .utility import DEBUG, INFO, absv, is_continuous, log, log_level, ensure_constant
```
Patol75: Let's try to preserve the alphabetical order.

gadopt/transport_solver.py
```diff
@@ -390,3 +391,112 @@ def set_equation(self) -> None:
...
+
+        for subdomain_id, bc in bcs.items():
+            for bc_type, value in bc.items():
+                if bc_type == 'T':
+                    # GenericTransportSolver expects "g" for Dirichlet boundary conditions
+                    transport_bcs[subdomain_id] = {"g": value}
+                else:
+                    raise ValueError("Boundary conditions other than Dirichlet are not supported.")
+
+        return transport_bcs
```
Patol75: Do we need this method? Smoothing should be applicable to any field, and it is not clear to me why the strong bcs tag should be `T`. If `g` (for general) is not suitable, we could use `s` (for smoothing). Additionally, weak bcs are allowed in `diffusion_term`, so if people, for some reason, want to use them, shouldn't they be allowed to?

demos/smoothing/smoothing.py
sghelichkhani: It does. File is from the adjoint tests and experiments. But since we are having the checkpointfile on data.gadopt.org just for this sake, I named it that way so it would be easier for us to know why we have that file there. The comment explains that this is the same file as in there, we are just using it for demonstration there.

demos/smoothing/smoothing.py
sghelichkhani: Not really clear what you mean by this. If it is meant to describe the isotropic versus anisotropic properties, this is not really the place. I have used this just to reiterate what we have done here.

demos/smoothing/smoothing.py
```diff
@@ -0,0 +1,255 @@
...
+
+# +
+smooth_solution_isotropic = Function(T.function_space(), name="Smooth_Temperature_Isotropic")
+smoother_isotropic = DiffusiveSmoothingSolver(
+    function_space=T.function_space(),
+    wavelength=0.2,  # Smoothing wavelength parameter
+    bcs=temp_bcs)
+
+# Apply isotropic smoothing
+smooth_solution_isotropic.assign(smoother_isotropic.action(T))
```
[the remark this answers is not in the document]
sghelichkhani: https://github.com/g-adopt/g-adopt/discussions/315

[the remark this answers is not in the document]
sghelichkhani: And again beyond the scope of here.

gadopt/transport_solver.py
```diff
@@ -390,3 +391,112 @@ def set_equation(self) -> None:
...
+            if not isinstance(p, int):  # Tensor-product element
+                p = max(p)
+            integration_quad_degree = 2 * p + 1
+
+        # For anisotropic diffusion, use average diffusivity
+        if hasattr(K, 'ufl_shape') and len(K.ufl_shape) > 0:
+            # Tensor diffusivity
+            K_avg = (
+                assemble(sqrt(inner(K, K)) * dx(mesh, degree=integration_quad_degree)) /
+                assemble(Constant(1) * dx(mesh, degree=integration_quad_degree))
```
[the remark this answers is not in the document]
sghelichkhani: It is actually! You need either an expression or a form for an integration.

sghelichkhani (commented):
Thanks @Patol75. Some good points there. Some I have left a comment in response only.

demos/smoothing/smoothing.py
w3168: @sghelichkhani maybe I misunderstood but I ran the demo as is and this doesn't work...? In your makefile you get the right data file so I guess it all runs with the CI but I think in a jupyter notebook we also need to download `smoothing-example.h5` ?

demos/smoothing/smoothing.py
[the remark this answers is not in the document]
sghelichkhani: Thanks both! Happy to go with the team! Let's change to space_function, and if you are OK, let's keep it as `action` as it clarifies what is being smoothened. Thanks agin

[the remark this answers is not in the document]
sghelichkhani: Please check the recent change and spare an "approved" if you are happy

demos/smoothing/smoothing.py
sghelichkhani: Is changing all filenames to `smoothing-example.h5` not enough? That files is not the correct file for this. So doesn't make sense to have that check for thi.s

gadopt/transport_solver.py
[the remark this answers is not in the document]
sghelichkhani: Agreed off-topically! Way less error-prone.

demos/smoothing/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+all: params.log
+
+params.log : category := smoothing
+params.log: smoothing.py smoothing-example.h5
+	$(run-python)
+
+smoothing-example.h5:
+	@echo "Downloading checkpoint file..."
+	@[ ! -f smoothing-example.h5 ] && wget -q https://data.gadopt.org/demos/smoothing-example.h5 || true
```
angus-g: What is the provenance of this? How do we regenerate it?

dsroberts: Isn't this the kind of thing we use `git-lfs` for?

sghelichkhani:
Should be good to go in!

gadopt/transport_solver.py
```diff
@@ -390,3 +391,93 @@ def set_equation(self) -> None:
...
+            )
+        else:
+            # Scalar diffusivity
+            K_avg = K
+
+        return Constant(wavelength**2 / (4 * K_avg))
+
+    def action(self, T: Function) -> None:
```
Patol75: Should we change `T` here to something like `field`?

Patol75 (changes requested):
Thank you, Sia! The PR is now in great shape! I was going to approve it, but there is one outstanding issue in the `Makefile` that needs to be addressed. I also added three super minor suggestions; your call. Once the `Makefile` is fixed, if everyone else is happy, then feel free to merge.

demos/smoothing/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+all: params.log
+
+params.log : category := smoothing
+params.log: smoothing.py smoothing-example.h5
+	$(run-python)
+
+smoothing-example.h5:
+	@echo "Copying checkpoint file from tests directory..."
+	@[ ! -f smoothing-example.h5 ] && cp ../../tests/adjoint_2d_cylindrical/Checkpoint230.h5 smoothing-example.h5 || true
```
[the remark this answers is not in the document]
angus-g: No problem! Yep, that's the way to do it :)

PR #71, opened 2024-05-08
Use string to request default solver parameters

Patol75:
Default iterative and direct Stokes solver parameters can now be specified using eponymous strings.

### DIFF
--- a/tests/test_stokes_solver_configuration.py
+++ b/tests/test_stokes_solver_configuration.py
@@ -1,5 +1,5 @@
 import firedrake as fd
-import numpy as np
+import pytest
 
 from gadopt.approximations import BoussinesqApproximation
 from gadopt.stokes_integrators import (
@@ -81,13 +81,7 @@ def test_solver_parameters_argument():
             solver_parameters=solver_parameters,
         )
 
-        np.testing.assert_equal(stokes_solver.solver_parameters, expected_value)
+        assert stokes_solver.solver_parameters == expected_value
 
-    np.testing.assert_raises(
-        ValueError,
-        StokesSolver,
-        stokes_function,
-        temperature,
-        approximation,
-        solver_parameters="",
-    )
+    with pytest.raises(ValueError):
+        StokesSolver(stokes_function, temperature, approximation, solver_parameters="")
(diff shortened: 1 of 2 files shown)


drhodrid:
Going to tag @stephankramer and @angus-g for this one - just in case I miss something subtle. However, it looks simple and good to me.

drhodrid:
Could you add a short test of the new functionality?

angus-g:
> Could you add a short test of the new functionality?

Probably a good candidate for just a unit test here.

I think the proposed idea is good and sensibly-implemented. I guess the `match/case` construct is Python 3.10+, which we kind of already require because of the type annotations anyway. I think it's fair to say that Python 3.10 is easily available, but it is indeed not always the _default_, which can cause issues.

drhodrid:
Thomas please add a test, and then this is good to go.

Patol75:
Thank you for requesting a test; my implementation did not work! Not sure how that happened. Anyway, I fixed it and part of the logic in `stokes_integrator.py`; let me know if everything looks reasonable to you. I added a test, which passes locally. Hopefully, the CI will confirm. One small issue I spotted is that `fd.interpolate` is still used in some of the tests despite its deprecation status.

tests/test_stokes_solver_configuration.py
```diff
@@ -0,0 +1,93 @@
...
+        np.testing.assert_equal(stokes_solver.solver_parameters, expected_value)
+
+    np.testing.assert_raises(
+        ValueError,
+        StokesSolver,
+        stokes_function,
+        temperature,
+        approximation,
+        solver_parameters="",
+    )
```
angus-g, line 93: I guess `assert_equal` is fine enough here (although it doesn't need any of the np-specific handling), but I think it would be clearer to use a regular `with pytest.raises`. And in that case, might as well not use np in this file at all?

angus-g (commented):
Test looks good, this is just about ready to go

tests/test_stokes_solver_configuration.py
Patol75: Thank you, Angus. I did not realise that `pytest` is very capable to handle verbosely plain `assert`. Should be good to go now.

angus-g (approved):
Looks good, thanks!

PR #74, opened 2024-05-30
Require `cartesian` argument when instantiating approximation

Patol75:
`cartesian` is currently supplied to `StokesSolver` and any approximation inheriting from `ExtendedBoussinesqApproximation`. In both cases, it is an optional argument that defaults to `True`. The current situation may lead to confusion. When setting up a non-Cartesian simulation, one may well omit to supply `cartesian=False` to either of the approximation and Stokes solver without any complaint from G-ADOPT. Moreover, one could assume that `cartesian` does not apply to `BoussinesqApproximation` when, in fact, it indirectly does through `StokesSolver`.

The present PR enforces `cartesian` as a required argument for all approximations and removes it from `StokesSolver`'s argument list.

@drhodrid and @sghelichkhani, please feel free to add more information.

### DIFF
--- a/demos/2d_compressible_ALA/compressible_case_ALA.py
+++ b/demos/2d_compressible_ALA/compressible_case_ALA.py
@@ -44,7 +44,9 @@ alphabar = Function(Q, name="IsobaricThermalExpansivity").assign(1.0)
 cpbar = Function(Q, name="IsobaricSpecificHeatCapacity").assign(1.0)
 chibar = Function(Q, name="IsothermalBulkModulus").assign(1.0)
 
-approximation = AnelasticLiquidApproximation(Ra, Di, rho=rhobar, Tbar=Tbar, alpha=alphabar, chi=chibar, cp=cpbar)
+approximation = AnelasticLiquidApproximation(
+    Ra, Di, cartesian=True, rho=rhobar, Tbar=Tbar, alpha=alphabar, chi=chibar, cp=cpbar
+)
 
 time = 0.0
 steady_state_tolerance = 1e-9
@@ -92,9 +94,14 @@ stokes_bcs = {
 }
 
 energy_solver = EnergySolver(T, u, approximation, delta_t, ImplicitMidpoint, bcs=temp_bcs)
-stokes_solver = StokesSolver(z, T, approximation, bcs=stokes_bcs,
-                             cartesian=True, constant_jacobian=True,
-                             transpose_nullspace=Z_nullspace)
+stokes_solver = StokesSolver(
+    z,
+    T,
+    approximation,
+    bcs=stokes_bcs,
+    constant_jacobian=True,
+    transpose_nullspace=Z_nullspace,
+)
 
 checkpoint_file = CheckpointFile("Checkpoint_State.h5", "w")
 checkpoint_file.save_mesh(mesh)
--- a/demos/2d_compressible_TALA/compressible_case_TALA.py
+++ b/demos/2d_compressible_TALA/compressible_case_TALA.py
@@ -41,7 +41,9 @@ alphabar = Function(Q, name="IsobaricThermalExpansivity").assign(1.0)
 cpbar = Function(Q, name="IsobaricSpecificHeatCapacity").assign(1.0)
 chibar = Function(Q, name="IsothermalBulkModulus").assign(1.0)
 
-approximation = TruncatedAnelasticLiquidApproximation(Ra, Di, rho=rhobar, Tbar=Tbar, alpha=alphabar, cp=cpbar)
+approximation = TruncatedAnelasticLiquidApproximation(
+    Ra, Di, cartesian=True, rho=rhobar, Tbar=Tbar, alpha=alphabar, cp=cpbar
+)
 
 time = 0.0
 steady_state_tolerance = 1e-9
@@ -93,9 +95,14 @@ stokes_bcs = {
 }
 
 energy_solver = EnergySolver(T, u, approximation, delta_t, ImplicitMidpoint, bcs=temp_bcs)
-stokes_solver = StokesSolver(z, T, approximation, bcs=stokes_bcs,
-                             cartesian=True, constant_jacobian=True,
-                             transpose_nullspace=Z_nullspace)
+stokes_solver = StokesSolver(
+    z,
+    T,
+    approximation,
+    bcs=stokes_bcs,
+    constant_jacobian=True,
+    transpose_nullspace=Z_nullspace,
+)
 
 checkpoint_file = CheckpointFile("Checkpoint_State.h5", "w")
 checkpoint_file.save_mesh(mesh)
--- a/demos/2d_cylindrical/2d_cylindrical.py
+++ b/demos/2d_cylindrical/2d_cylindrical.py
@@ -37,7 +37,7 @@ r = sqrt(X[0]**2 + X[1]**2)
 T.interpolate(rmax - r + 0.02*cos(4*atan2(X[1], X[0])) * sin((r - rmin) * pi))
 
 Ra = Constant(1e5)  # Rayleigh number
-approximation = BoussinesqApproximation(Ra)
+approximation = BoussinesqApproximation(Ra, cartesian=False)
 
 delta_t = Constant(1e-7)  # Initial time-step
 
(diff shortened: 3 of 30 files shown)


angus-g:
I agree that this is a source of confusion, but I'm not sure I like the approach of pushing the flag backwards into the approximation. It seems like the approximation shouldn't really be restricted to the geometry in that way.

My suggestion would be to have the solver indicate that to the approximation, e.g. by injecting the `cartesian` flag. I'd be interested in hearing the others' opinions though!

sghelichkhani:
@angus-g has a point. Knowing that approximations are used only by StokesSolver to get the additional terms, one can use  StokesSolver itself  to set e.g., `ExtendedBoussinesqApproximation.cartesian`. In a way we don't  need to know cartesian False or True before setting up the solvers anyway!  

Another way could be by providing the flag as an argument to those places where we call work_against_gravity` in approximations, but that might involve more changes to the code, and maybe unnecessary?

drhodrid:
Thanks for putting this PR together Thomas!

For background, the issue is that I (probably stupidly) assumed that if you specified cartesian = False under Stokes Solver, the EOS approximation would be made aware of this. This is not currently the case, which means I’ve recently been running spherical models, where one of the additional terms in the EBA approximation (which underpins TALA and ALA) has been using a Cartesian coordinate direction for “work done against gravity”. Obviously, these simulations are now useless - but the error was very subtle and difficult to spot. The same is true for some of Sia’s computationally expensive simulations. This type of issue is something we really need to avoid.

When talking to Thomas and looking through the code, it seems that we are in a position where we need to specify Cartesian = False in many places (off the top of my head, in approximation, Stokes Solver, and even layer average). As a user, this is not ideal. Ideally it would be specified in one place, and carried through.

How this is best done, I think is best handled by you devs - I can see merit to having it under Stokes solver. But we really need to be able to avoid having some aspects of the model assuming we’re in Cartesian and others assuming we’re in spherical.



On 31 May 2024, at 7:58 AM, Sia Ghelichkhan ***@***.***> wrote:



@angus-g<https://github.com/angus-g> has a point. Knowing that approximations are used only by StokesSolver to get the additional terms, one can use StokesSolver itself to set e.g., ExtendedBoussinesqApproximation.cartesian. In a way we don't need to know cartesian False or True before setting up the solvers anyway!

Another way could be by providing the flag as an argument to those places where we call work_against_gravity` in approximations, but that might involve more changes to the code, and maybe unnecessary?

angus-g:
Having looked at it a bit more, I think the _"best"_ place to set the property would be on the mesh itself, e.g. by setting something like `mesh.cartesian = True`:
- `work_against_gravity` doesn't need to know the specifics, it just gets it from `utility.vertical_component` which then looks at the mesh underpinning the specified function
- similarly, the `LayerAveraging` class can work out what is the depth direction by querying the mesh it's provided
- `StokesSolver` uses `utility.upward_normal`
- and the only other place is for determining the default solver, which already looks at the mesh topological dimension, so it could just look at whether it's cartesian too

The only issue here is that it's difficult to enforce... We could add it as a parameter on `StokesSolver` that then sets a property on the mesh, but if we use e.g. `LayerAveraging` before that, it won't work. We could do some kind of extra wrapping with a very thin g-adopt specific extension to `Mesh` that essentially only tracks whether it's cartesian or not. In that case, `LayerAveraging`/`StokesSolver` can assume some default if given a regular Firedrake mesh, and complain. Something like the following:

```python
# g-adopt specific cartesian mesh
mesh = CartesianMesh(UnitSquareMesh(nx, ny))
...

# g-adopt specific non-cartesian mesh
mesh1d = NonCartesianMesh(CircleManifoldMesh(ncells, radius=rmin, degree=2))
mesh = ExtrudedMesh(mesh1d, layers=nlayers, extrusion_type='radial')
...

# regular Firedrake mesh
mesh = UnitSquareMesh(nx, ny)
...
stokes_solver = StokesSolver(z, T, ...)

# prints:
"warning: assuming cartesian mesh"
```

Anyway, this is all just spitballing on a potential API design!

Patol75:
I like the idea of attaching the information to the mesh. Looking at the code, I believe that all we need to define is the `upwards_direction`. I assumed there was an explicit coordinate system difference in Firedrake, but I do not think that is the case. In other words, it seems to me that Firedrake always uses Cartesian coordinates (please correct me if I am wrong). Coming back to `upwards_direction`, I am thinking of the following syntax:
```python
ga.upwards_direction(mesh, direction="radial")
```
Under the hood, the function creates `mesh.upwards_direction` as a Firedrake `Function`. `direction` should be a string and values could be `x`, `y`, `z`, or `radial`. If the user omits the function call, then we can proceed as @angus-g suggested.

sghelichkhani:
Thanks all for the brainstorming. 

The decorator approach, i.e., adding the feature to mesh, might still lead to issues as we cannot enforce it! The same issue above in Rhodri's comment could have easily happened with the user forgetting to use the decorator. So whatever solution, we need to be able to enforce it.

 I am note sure I am understanding the change @Patol75. We do already have `upward_normal` in  https://github.com/g-adopt/g-adopt/blob/f9f7a38a8adfe9264286b4f9bee503da188e7f55/gadopt/utility.py#L86-L87 which does the same job (there we only assume upwards directions in cartesian is n-1). What am I missing?
 
 I still think letting `StokesSolver` decide on `upwards_normal` everywhere might work best.

angus-g:
> ... which does the same job (there we only assume upwards directions in cartesian is n-1). What am I missing?

The proposal is to create a mesh, then define its upward normal. Then whenever it is required, it is retrieved from the mesh itself. i.e.
```python
mesh = Mesh(...)
ga.upward_normal(mesh, direction="radial")
la = LayerAveraging(mesh, ...) # knows the direction is radial from the mesh itself
solver = StokesSolver(z, ...) # doesn't need a cartesian argument; knows from the mesh itself
```

> I still think letting StokesSolver decide on upwards_normal everywhere might work best.

This doesn't solve the issue when you need to know whether the mesh is cartesian or not _before_ the StokesSolver is created, e.g. https://github.com/g-adopt/g-adopt/blob/f9f7a38a8adfe9264286b4f9bee503da188e7f55/demos/3d_spherical/3d_spherical.py#L66

sghelichkhani:
I see! I would still be worried about the fact we are not enforcing it. But I agree that it is the best solution thus far.

Patol75:
```python
from functools import wraps

import firedrake as fd


def require_upwards_direction(firedrake_mesh):
    @wraps(firedrake_mesh)
    def with_upwards_direction(*args, upwards_direction, **kwargs):
        mesh = firedrake_mesh(*args, **kwargs)

        cart_upw_dir = [0] * mesh.geometric_dimension()
        mesh.upwards_direction = fd.Function(fd.VectorFunctionSpace(mesh, "CG", 1))
        match upwards_direction:
            case "x":
                cart_upw_dir[0] = 1
                mesh.upwards_direction.interpolate(fd.as_vector(cart_upw_dir))
            case "y":
                cart_upw_dir[1] = 1
                mesh.upwards_direction.interpolate(fd.as_vector(cart_upw_dir))
            case "z":
                cart_upw_dir[2] = 1
                mesh.upwards_direction.interpolate(fd.as_vector(cart_upw_dir))
            case "radial":
                mesh_coords = fd.SpatialCoordinate(mesh)
                mesh.upwards_direction.interpolate(
                    mesh_coords / fd.sqrt(fd.dot(mesh_coords, mesh_coords))
                )

        return mesh

    return with_upwards_direction


UnitSquareMesh = require_upwards_direction(fd.UnitSquareMesh)

mesh = UnitSquareMesh(2, 4, upwards_direction="y")
print(mesh.upwards_direction.dat.data)
mesh = UnitSquareMesh(2, 4)
# TypeError: UnitSquareMesh() missing 1 required keyword-only argument: 'upwards_direction'

```

stephankramer:
Interesting discussion!

Adding my own 0.02 AUD:
First of all I agree that we should really be specifying this in one place. At first glance I was happy with what this PR was proposing, but reading @angus-g objection above I agree that it's not necessarily the most natural place to define this: it would be nice to have the approximation be independent of this choice. As pointed out above the only reason it needs it is to work out the gravity direction and so maybe indeed that should be done in a central location. If I'm being honest I'm a little bit nervous about monkey-patch-attaching it to the firedrake Mesh object. Purely seen from a gadopt user interface perspective, I can see that it's quite nice having it defined there - it being associated with the geometry object. What I don't like is that it kind of breaks the picture of our user interface to a large extent existing of pure Firedrake objects for which the user can use their firedrake knowledge and firedrake documentation at heart to do pre and post-processing etc. and then we provide some additional tools and functionality to form the heart of the model. I realize this is a bit of a subjective objection, but what I specifically worry about is coupling to other firedrake-based utilities, e.g. a mesh adaptivity layer - it replaces the mesh does it need to know about this enrichment with an upward normal? - or what about checkpointing in firedrake, it's probably going to barf about unexpected attributes. What about geoscientific problem that don't involved gravity? In any case, any example that I can come up with is something that can be easily fixed - it's just a premonition that I have that it might be not optimal.

What I would suggest is that in the approximation, and anywhere else that needs it, the upward_normal is just made an explicity argument to the method/function. Say in `work_against_gravity()`, all it needs is some UFL expression, and we can provide it from where it's being called (from inside StokesSolver) just like we provide the u and the p. The exact expression btw might in the future be a choice that may depend on numerical considerations, e.g. an interpolated gravity normal as in the code above - but then what should the polynomial degree be? - or a pure UFL expression: `mesh_coords / fd.sqrt(fd.dot(mesh_coords, mesh_coords))`.

To force the user (i.e. @drhodrid ) to make a consistent choice, say when it's using both StokesSolver and LayerAveraging, you could just have it define that expression and explicitly pass it as a required argument to these two:
`
  khat  = ...  # UFL expression
  solver = StokesSolver(..., khat, ...)
  averager = LayerAverager(..., khat, ...)
`
where the RHS of `upward_normal` could just be the current `upward_normal(mesh, cartesian=...)`
StokesSolver would then in turn pass it on when calling any method of approximation that needs it.
I'll admit not as pretty, but a bit more flexible.

Patol75:
Superseeded by gh-99.

PR #75, opened 2024-06-03
Convert GeodynamicalDiagnostics to accept a mixed function

angus-g:
Instead of splitting our mixed function `z` and then passing `u` and `p` around (e.g. to `GeodynamicalDiagnostics`), it seems to make more sense to provide it with `z` directly. We have precedent for this already in `StokesSolver`, especially with the free surface fields coming in #51.

Additionally, instead of exposing some of the gory MPI internals in our demos, I've pushed that into a `ux_max` diagnostic, which can be given a surface ID. I think related to the discussion in #74, this would be better off knowing what the "lateral" direction is, as it just makes the cartesian assumption at the moment.

### DIFF
--- a/demos/2d_compressible_ALA/compressible_case_ALA.py
+++ b/demos/2d_compressible_ALA/compressible_case_ALA.py
@@ -1,5 +1,4 @@
 from gadopt import *
-from mpi4py import MPI
 
 # Set up geometry:
 nx, ny = 40, 40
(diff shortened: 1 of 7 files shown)


Patol75 (commented):
That looks very sensible to me.

drhodrid (approved):
Clean and glorious. Looks fab. Pending green CI, this is good to go.

PR #76, opened 2024-06-03
Convert demos to notebook-compatible format

drhodrid:
In this branch, I have updated the following tutorials, as planned:
1. Base case.
2. ALA.
3. TALA.
4. Viscoplastic.
5. 3-D Cartesian.

I feel now is a good time to get it through the testing environment and have somebody else's eyes over it - ideally @angus-g . I will next work on another branch to finalise:
1. 2-D cylindrical.
2. 3-D spherical. 

And after that, the DP Spiegelman case. 

From my end, aside from the adjoint and level set tutorials that Sia and Thomas is working on, everything else can be shifted across to a longtest.


angus-g (approved):
Happy with this for the moment, and we can put the progress logging into the notebooks later.

PR #77, opened 2024-06-03
Spherical notebooks

drhodrid:
This includes the transition of 2D cylindrical and 3D spherical case across to a format that can be used for notebooks. To be merged after #76.

Note for Angus - I have corrected definitions of Nu such that numbers for Nu_bottom and Nu_top are strictly positive (but otherwise should be identical to CI check). The test values will need to be updated to reflect this, before CI goes green. 

All that's outstanding for me is Drucker-Praeger, but need some of Angus' wisdom on how best to transfer that across to a Notebook (and will subsequently need Steph to fill in the gaps!).


angus-g:
Is this the next one to go in?

drhodrid:
It is - I’m just hoping to figure out the dx stuff today… sorry yesterday was a bit of a write off.

I guess alternatively it can go in as is, if you’re happy, and I can sort the dx stuff with a later commit?

On 6 Jun 2024, at 8:59 AM, Angus Gibson ***@***.***> wrote:



Is this the next one to go in?

angus-g:
That's alright, happy to wait on the dx stuff.

drhodrid:
OK this seems to be green. If you're happy @angus-g I'm now happy too.

angus-g:
This still has the issue of bringing in pyvista as a requirement... I guess there's overlap between the package set required for testing and the demos, so maybe we can rename that and use it for both. Will need addition to install instructions, i.e. `pip install gadopt[testing]` (or whatever we call it, maybe `demos`). And a change https://gadopt.org/install/...

PR #78, opened 2024-06-04
Convert Drucker-Prager demo to a tutorial

drhodrid:
Adding a pull request now, such that I can use the testing suite as I make changes to bring this test case into line with other g-adopt examples. I will ultimately need help from the gurus on some aspects.


drhodrid:
I think this one can be closed, since @stephankramer was going to write a separate tutorial and the test has been moved across to the new folder.

PR #79, opened 2024-06-04
Add multi-material tutorials

Patol75:
This PR brings multi-material tutorials to the G-ADOPT framework. Happy to take comments and make necessary changes.


demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,281 @@
...
+#
+# $$\psi(\mathbf{x}, t) = \frac{1}{2} \left[
+# \mathrm{tanh} \left( \frac{\phi(\mathbf{x}, t)}{2\epsilon} \right) + 1
+# \right]$$
+#
+# Throughout the simulation, the level-set field is advected with the flow:
+#
+# $$\frac{\partial \psi}{\partial t} + \nabla \cdot \left( \mathbf{u}\psi \right) = 0$$
+#
+# Advection of the level set deteriorates the shape of the initial profile. To maintain
```
drhodrid, line 34: I find "deteriorates the shape of the initial profile" a little vague. Can you write explicitly what you mean here?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,281 @@
...
+plog.close()
+
+with CheckpointFile("Final_State.h5", "w") as final_checkpoint:
+    final_checkpoint.save_mesh(mesh)
+    final_checkpoint.save_function(T, name="Temperature")
+    final_checkpoint.save_function(z, name="Stokes")
+    final_checkpoint.save_function(psi, name="Level set")
+# -
+
+# We can visualise the final temperature field using Firedrake's built-in plotting
```
drhodrid, line 273: temperature --> level set?

demos/multi_material/thermochemical_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,313 @@
+# Rayleigh-Taylor instability
+# ===========================
+#
+# Rationale
+# ---------
+#
+# One may wish to simulate a geodynamical flow involving multiple physical phases. A
```
drhodrid, line 7: For introduction to this tutorial, I think you can be less verbose, building on the previous example.

demos/multi_material/thermochemical_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,313 @@
...
+plog.close()
+
+with CheckpointFile("Final_State.h5", "w") as final_checkpoint:
+    final_checkpoint.save_mesh(mesh)
+    final_checkpoint.save_function(T, name="Temperature")
+    final_checkpoint.save_function(z, name="Stokes")
+    final_checkpoint.save_function(psi, name="Level set")
+# -
+
+# We can visualise the final temperature field using Firedrake's built-in plotting
```
drhodrid, line 305: Visualise T and phi here?

drhodrid (commented):
At a glance this looks good. I have made some minor comments and will try running shortly. For my info - this can now be run on master - given everything has been merged?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,281 @@
...
+# the initial time, initial time step $\Delta t$, and output frequency (in time units).
+
+# +
+time_now = 0  # Initial time
+delta_t = Function(R).assign(1)  # Initial time step
+output_frequency = 10  # Frequency (based on simulation time) at which to output
+t_adapt = TimestepAdaptor(
+    delta_t, u, V, target_cfl=0.6, maximum_timestep=output_frequency
+)
+# -
```
angus-g, line 165:
```suggestion
time_now = 0  # Initial time
delta_t = Function(R).assign(1)  # Initial time step
output_frequency = 10  # Frequency (based on simulation time) at which to output
t_adapt = TimestepAdaptor(
    delta_t, u, V, target_cfl=0.6, maximum_timestep=output_frequency
)
```

Blocks like this without any blank lines don't need explicit delimiters -- maybe a little easier to read without them?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,281 @@
...
+
+# This problem has a constant pressure nullspace, handled identically to our previous
+# tutorials.
+
+# +
+Z_nullspace = create_stokes_nullspace(Z)
+# -
```
angus-g, line 172:
```suggestion
Z_nullspace = create_stokes_nullspace(Z)
```

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,281 @@
...
+# on the left and ride sides.
+
+# +
+stokes_bcs = {
+    bottom_id: {"ux": 0, "uy": 0},
+    top_id: {"ux": 0, "uy": 0},
+    left_id: {"ux": 0},
+    right_id: {"ux": 0},
+}
+# -
```
angus-g, line 184:
```suggestion
stokes_bcs = {
    bottom_id: {"ux": 0, "uy": 0},
    top_id: {"ux": 0, "uy": 0},
    left_id: {"ux": 0},
    right_id: {"ux": 0},
}
```

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,281 @@
...
+# +
+Z_nullspace = create_stokes_nullspace(Z)
+# -
+
+# Boundary conditions are specified next: no slip at the top and bottom and free slip
+# on the left and ride sides.
+
+# +
+stokes_bcs = {
+    bottom_id: {"ux": 0, "uy": 0},
```
drhodrid, line 179: For zero slip, in 3-D example, we set u = 0, so we don't have to set the components individually. Just noting that for consistency.

Patol75:
Thank you for the suggestions; I am working on them.

@drhodrid Yes, everything should run on `master`.

drhodrid:
Running them now and will have a few more comments later this eve. For now, all I'd add to previous reviews is:
1. Lovely to see this setup. It actually demonstrates how elegant what you have done is. Well done.
2. Could you call the python files something that makes them easier to distinguish (they currently are both vankekek1997.py).
3. Given that these tutorials are also tested, you'll need to provide some diagnostic values at the end of the run so that @angus-g can add these to our testing platform. The cases run quickly so perhaps increase resolution (x2) so that the tutorials look better resolved and the diagnostic outputs are more meaningful? 
4. In tutorial 1, remove temperature from VTK output, as it is confusing (even though it is zero).

Patol75:
1. I am glad you like it.
2. Given I put them in different subdirectories, I thought that was enough. Or do you prefer I get rid of subdirectories and change file names?
3. Will do. Keep in mind the entrainment diagnostic for the thermochemical case will look nothing like the published benchmark.
4. Well, there will always be a confusion, given that Stokes solver requires a temperature field in the first place.

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,276 @@
...
+nx, ny = 40, 40  # Number of cells in x and y directions
+lx, ly = 0.9142, 1  # Domain dimensions in x and y directions
+# Rectangle mesh generated via Firedrake
+mesh = RectangleMesh(nx, ny, lx, ly, quadrilateral=True)
+left_id, right_id, bottom_id, top_id = 1, 2, 3, 4  # Boundary IDs
+
+V = VectorFunctionSpace(mesh, "CG", 2)  # Velocity function space (vector)
+W = FunctionSpace(mesh, "CG", 1)  # Pressure function space (scalar)
+Z = MixedFunctionSpace([V, W])  # Stokes function space (mixed)
+Q = FunctionSpace(mesh, "CG", 2)  # Temperature function space (scalar)
```
drhodrid, line 72: Q not used or needed?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,276 @@
...
+K = FunctionSpace(mesh, "DQ", 2)  # Level-set function space (scalar, discontinuous)
+R = FunctionSpace(mesh, "R", 0)  # Real space for time step
+
+z = Function(Z)  # A field over the mixed function space Z
+u, p = split(z)  # Symbolic UFL expressions for velocity and pressure
+z.subfunctions[0].rename("Velocity")  # Associated Firedrake velocity function
+z.subfunctions[1].rename("Pressure")  # Associated Firedrake pressure function
+T = Function(Q, name="Temperature")  # Firedrake function for temperature
```
drhodrid, line 80: Does this need to be Q2? Could we not put it on CG 1 (W) function space, thus reducing the need for Q above (given that T is not really necessary for this simulation, aside from G-ADOPT requirement!)

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,276 @@
...
+# -
+
+# We now provide initial conditions for the level set field. To this end, we use the
+# shapely library to represent the initial location of the material interface and
+# derive the signed-distance function. Finally, we apply the transformation to obtain a
+# smooth step function profile.
+
+# +
+import numpy as np
+import shapely as sl
```
drhodrid, line 91: Just wanted to confirm that shapely was now included as a g-adopt dependency? When I installed g-adopt using pip install -e earlier (after the merge of level sets), I still needed to install shapely via pip. Is this the expected behavior or should it now come in via g-adopt?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+interface_deflection = 0.02  # Amplitude of the cosine function marking the interface
+interface_wavelength = 2 * lx  # Wavelength of the cosine function
+material_interface_y = 0.2  # Vertical shift of the interface along the y axis
+# Group parameters defining the cosine profile
+isd_params = (interface_deflection, interface_wavelength, material_interface_y)
+
+# Shapely LineString representation of the material interface
+interface_x = np.linspace(0, lx, 1000)
```
drhodrid, line 106: Out of interest, are results at all sensitivity to the choice of 1000 here? Obviously irrelevant for this tutorial, but perhaps when we think about publishing this approach. 

Probably a comment here too about why you chose 1000?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,276 @@
...
+
+# +
+buoyant_material = Material(RaB=-1)  # Vertical direction is flipped in the benchmark
+dense_material = Material(RaB=0)
+materials = [buoyant_material, dense_material]
+
+Ra = 0  # Thermal Rayleigh number
+
+RaB = field_interface(
+    [psi], [material.RaB for material in materials], method="arithmetic"
```
drhodrid, line 150: Worth adding a comment here about the "arithmetic" averaging approach used at the interface (I assume that's what this is?) alongside the equation? Do you perhaps want to quickly note the other options available (geometric, harmonic) and those equations too?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,276 @@
...
+delta_t = Function(R).assign(1)  # Initial time step
+output_frequency = 10  # Frequency (based on simulation time) at which to output
+t_adapt = TimestepAdaptor(
+    delta_t, u, V, target_cfl=0.6, maximum_timestep=output_frequency
+)
+
+# This problem has a constant pressure nullspace, handled identically to our previous
+# tutorials.
+
+Z_nullspace = create_stokes_nullspace(Z)
```
drhodrid, line 170: I realise it does the same thing, but for consistency with previous tutorials, do you want to say closed = True, rotational = False?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,276 @@
...
+
+# Boundary conditions are specified next: no slip at the top and bottom and free slip
```
drhodrid, line 172: Is it worth adding a comment here that since all domain boundaries are closed, there is no boundary condition for the level set required?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,276 @@
...
+# instantiate G-ADOPT geodynamical diagnostic utility.
+
+# +
+output_file = VTKFile("output.pvd")
+
+plog = ParameterLog("params.log", mesh)
+plog.log_str("step time dt u_rms entrainment")
+
+gd = GeodynamicalDiagnostics(z, T, bottom_id, top_id)
+material_area = material_interface_y * lx  # Quantity of tracked material
```
drhodrid, line 193: Instead of quantity - I would say areal fraction, or volume fraction? It looks like that's what this is?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,276 @@
...
+entrainment_height = 0.2  # Height above which entrainment diagnostic is calculated
```
drhodrid, line 194: Say a little more about what this diagnostic is (i.e. the integral of a certain material above the specified height)?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,276 @@
...
+stokes_solver = StokesSolver(
+    z,
+    T,
+    approximation,
+    bcs=stokes_bcs,
+    nullspace=Z_nullspace,
+    transpose_nullspace=Z_nullspace,
+)
+
+subcycles = 1  # Number of advection solves to perform within one time step
```
drhodrid, line 211: Provide some context for this? You use 1, but is this the recommendation for users too? What are the other options available and what are the trade-offs?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,276 @@
...
+level_set_solver = LevelSetSolver(psi, u, delta_t, eSSPRKs10p3, subcycles, epsilon)
+# Increase the reinitialisation time step to make up for the coarseness of the mesh
```
drhodrid, line 213: Make it cleaar that this is increased relative to the default values. If you increase the resolution of this case, will you need to revisit this parameter choice? How should a user pick the value?

demos/multi_material/thermochemical_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,277 @@
...
+# essential to modelling Earth's mantle evolution.
+#
+# This example
+# ------------
+#
+# Here, we consider the entrainment of a thin, compositionally dense layer by thermal
+# convection presented in van Keken et al. (1997). Inside a 2-D domain heated from
+# below, a denser material sits at the bottom boundary beneath a lighter material.
+# Whilst the compositional stratification is stable, heat transfer from the boundary
+# generates positive buoyancy in the denser material, allowing thin layers of material
```
drhodrid, line 19: Instead of thing layers, perhaps use thin "tendrils" - that's the lingo I seem to remember.

drhodrid: I would add a comment at the end of this paragraph that the goal of this tutorial is to illustrate how to set up such a problem. To make the tutorial tractable, it must be run at a coarser resolution than is required, and hence results differ to those in the benchmark. Nonetheless, you have verified, by simulating this case at higher resolution, that things do converge towards those benchmark values.

demos/multi_material/thermochemical_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,277 @@
...
+
+stokes_bcs = {
+    bottom_id: {"uy": 0},
+    top_id: {"uy": 0},
+    left_id: {"ux": 0},
+    right_id: {"ux": 0},
+}
+temp_bcs = {bottom_id: {"T": 1}, top_id: {"T": 0}}
+
+# We move on to initialising the temperature field.
```
drhodrid, line 150: For consistency with other tutorials, could the temperature field be initialised before boundary conditions are set for T? I realise it doesn't make a difference, but the more consistent we are for a user, the better.

drhodrid (commented):
A few more specific comments. I really enjoyed going through this and learning. Goes to show the benefit of a nice tutorial! Well done!

demos/multi_material/thermochemical_buoyancy/van_keken_1997.py
```diff
@@ -0,0 +1,277 @@
...
+level_set_solver.reini_params["tstep"] *= 20
+# -
+
+# Finally, we initiate the time loop, which runs until the simulation end time is
+# attained.
+
+# +
+step = 0  # A counter to keep track of looping
+output_counter = 0  # A counter to keep track of outputting
+time_end = 0.05
```
drhodrid, line 219: Although the benchmark goes for this time, given we are under resolved, perhaps we only need to run the tutorial to 0.02 or similar (and we can flag that the true benchmark runs for longer with a comment)?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
Patol75: That will not change the fact that `T` is there. Defining `Q` aligns with other tutorials.

demos/multi_material/compositional_buoyancy/van_keken_1997.py
Patol75: What @angus-g said. :)

demos/multi_material/compositional_buoyancy/van_keken_1997.py
Patol75: This number defines the amount of points used to discretise the initial interface. You just need a number that is sufficiently large to capture the wiggles, if any, of the initial interface. If you put something too low, your initial state will not look like what you expected on ParaView.

demos/multi_material/compositional_buoyancy/van_keken_1997.py
Patol75: I edited the comment above.

demos/multi_material/compositional_buoyancy/van_keken_1997.py
Patol75: It is not a fraction though. It is the area occupied by the material in axis units. It would be a volume in 3-D.

demos/multi_material/compositional_buoyancy/van_keken_1997.py
drhodrid: True - but it's not really used here, hence confusing. Happy to leave as is for now and come back to this down the track if we find a way to not require T for simple multi-material Stokes solves.

demos/multi_material/compositional_buoyancy/van_keken_1997.py
drhodrid: Does does the discretised interface ultimately smear to some extent, and this 1000 is just to ensure the best possible representation of that initial condition with our current discretisation?

demos/multi_material/compositional_buoyancy/van_keken_1997.py
drhodrid: Just say area then? "Quantity" is a bit vague.

demos/multi_material/compositional_buoyancy/van_keken_1997.py
Patol75: This is used when the CFL for the energy system differs from that of the level-set advection. The latter should not exceed 0.6, but @stephankramer mentioned that the one for the energy system is usually higher. Subcycling allows multiple level-set advection steps without updating Stokes and energy in the meantime.

demos/multi_material/compositional_buoyancy/van_keken_1997.py
Patol75: I think it is pretty clear. With increased resolution, I would expect this parameter not to require as much of an increase. I do not really have clever advice on how to choose ideal reinitialisation parameters. It is a little bit of trial and error.

demos/multi_material/thermochemical_buoyancy/van_keken_1997.py
Patol75: I need to impose the boundary conditions on the initial state; that is why I swapped the order.

Patol75:
@drhodrid I have addressed a few comments and added responses and reactions to others. More from me tomorrow.

drhodrid:
Roger that. It’s looking good!

On 4 Jun 2024, at 8:27 PM, Thomas Duvernay ***@***.***> wrote:



@drhodrid<https://github.com/drhodrid> I have addressed a few comments and added responses and reactions to others. More from me tomorrow.

Patol75:
Today's update:

- **Diagnostics** Can I just provide the RMS velocity and entrainment values at the end of each simulation?
- I have addressed all comments.
- I am keen on seeing the rendered version. That will allow me to make further targeted changes.

angus-g:
>     * I am keen on seeing the rendered version. That will allow me to make further targeted changes.

https://78ff9a48.g-adopt.pages.dev/tutorials/compositional_buoyancy/ and https://78ff9a48.g-adopt.pages.dev/tutorials/thermochemical_buoyancy/ -- I guess first comment is to give them different titles!

Patol75:
Thank you, Angus. New changes pushed.

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -0,0 +1,291 @@
...
+z = Function(Z)  # A field over the mixed function space Z
+u, p = split(z)  # Symbolic UFL expressions for velocity and pressure
+z.subfunctions[0].rename("Velocity")  # Associated Firedrake velocity function
+z.subfunctions[1].rename("Pressure")  # Associated Firedrake pressure function
+T = Function(Q, name="Temperature")  # Firedrake function for temperature
+psi = Function(K, name="Level set")  # Firedrake function for level set
+# -
+
+# We now provide initial conditions for the level-set field. To this end, we use the
+# `shapely`` library to represent the initial location of the material interface and
```
angus-g, line 87: Typo?

```suggestion
# `shapely` library to represent the initial location of the material interface and
```

Patol75: Good catch. I blame that one on VSCode.

angus-g (approved):
Looks good from my end. Will let @drhodrid chime in with any extra comments when he gets a chance to take a look.

drhodrid:
I'd say these look really great and are pretty much good to go. I only have two minor comments:
1. For both cases, explain that resolving the material interface typically requires resolution beyond that provided in these tutorials. The goal of the tutorials is merely to showcase how things are set up.
2. For the thermo-chemical tutorial, where you now use gmsh (a nice addition which we don't have in our other tutorials!), I think that gmsh code block should be separated from the function spaces etc that follow.... I would also add a few more comments explaining what the key lines are doing for the gmsh generation.

Patol75:
@drhodrid For your first point, I do not think it applies to the compositional benchmark — the result is pretty convincing if you ask me — and I believe I already mentioned it in the introduction to the thermochemical benchmark. For your second point, I separated the GMSH block from the rest, but I have not added comments to the part where the GMSH API is used. Two reasons: the GMSH API is pretty self-explanatory (when you choose appropriate names for variables, functions, methods, and classes, usually you do not need many comments) and this should not be a GMSH tutorial.

drhodrid:
Ahhh I hadn't seen that change to the second tutorial. Nicely done. 

And yes ok - furry muff regarding gmsh. This is now good to go IMO. Did you want to merge?

Patol75:
@angus-g Do I need the regression test here? I cancelled it...

PR #80, opened 2024-06-04
Jupytext demo for adjoint

sghelichkhani:
This PR introduces the demo for adjoint. It also changes the forward case in our adjoint box case to serve as a demo if possible. Note that the inverse.py and taylor_test.py are as before, they only now work with the new filename for the state variables from the forward run.


demos/adjoint/taylor_test.py
```diff
@@ -25,13 +25,14 @@ def rectangle_taylor_test(case):
         case (str): name of the objective functional term
             either of "damping", "smooothing", "Tobs", "uobs"
     """
+    checkpoint_file = Path(__file__).resolve().parent / "adjoint-demo-checkpoint-state.h5"
```
angus-g, line 28: Missing `Path` import in this file?

sghelichkhani:
I think this is good to go! Def. needs a squash!

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,342 @@
...
+with CheckpointFile(checkpoint_filename, mode="r") as fi:
+    Tic.project(fi.load_function(mesh, "Temperature", idx=timesteps - 1))
+# -
+
+# We finally introduce our control problem to pyadjoint and continue by integrating the solutions
+# each time-step. Notice that we accumulatively compute the misfit term with respect to the
+# surface observable of plate velocities.
+
+# +
+# control
```
angus-g, line 159:
```suggestion
```

drhodrid:
I plan on taking a look at this after lunch.

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,342 @@
...
+    bottom_id: {"T": 1.0},
+    top_id: {"T": 0.0},
+}
+# Energy and Stokes solver
+energy_solver = EnergySolver(T, u, approximation, delta_t, ImplicitMidpoint, bcs=temp_bcs)
+stokes_solver = StokesSolver(z, T, approximation, bcs=stokes_bcs, nullspace=Z_nullspace, transpose_nullspace=Z_nullspace, constant_jacobian=True)
+# -
+
+# Define the Control Space
+# ========================
```
angus-g, line 138: I think the header levels need to be properly hierarchical, or it will break the table of contents in the notebook rendering. That is, this should be level 2, because the document title is level 1.

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,342 @@
...
+# ===============================
+#
+# In optimisation terminology, a reduced functional is a functional that takes a given value for the control and outputs the value of the objective functional defined for it. It does this without explicitly depending on all the intermediary state variables, hence the name "reduced".
+#
+# To define the reduced functional, we provide the class with an objective (which is an overloaded UFL object) and the control. Both of these are essential for formulating the reduced functional.
+
+# +
+# Define the object for pyadjoint
+reduced_functional = ReducedFunctional(objective, control)
+# -
```
angus-g, line 259:
```suggestion
reduced_functional = ReducedFunctional(objective, control)
```

I don't think comments like this are necessary with the big description above. Plus, it's clearer not to explicitly mark small blocks like this as code.

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,342 @@
...
+# ==================
+#
+# At this point, we have completed annotating the tape with the necessary information from running the forward simulation. To prevent further annotations during subsequent operations, we stop the annotation process. This ensures that no additional solves are unnecessarily recorded, keeping the tape focused only on the essential steps.
+#
+# We can then print the contents of the tape to verify what has been recorded.
+
+# +
+# Pause the annotation to stop recording further operations
+pause_annotation()
+# -
```
angus-g, line 271:
```suggestion
pause_annotation()
```

demos/adjoint/forward.py
```diff
@@ -1,71 +1,114 @@
...
 
-checkpoint_file = CheckpointFile("Checkpoint_State.h5", "w")
+# +
+# Save the initial temperature and average temperature to a checkpoint file.
+checkpoint_file = CheckpointFile("adjoint-demo-checkpoint-state.h5", "w")
 checkpoint_file.save_mesh(mesh)
 checkpoint_file.save_function(Taverage, name="Average Temperature", idx=0)
 checkpoint_file.save_function(T, name="Temperature", idx=0)
+# -
+
```
angus-g, line 87:
```suggestion
```

demos/adjoint/forward.py
```diff
@@ -90,22 +140,24 @@
...
+# +
+# Perform the time loop to solve the forward problem.
+for timestep in range(0, timesteps):
     stokes_solver.solve()
     energy_solver.solve()
 
-    # Storing velocity to be used in the objective F
-    checkpoint_file.save_function(u, name="Velocity", idx=timestep)
+    # Store the velocity field for use in the adjoint problem.
+    checkpoint_file.save_function(u, name="Velocity", idx=timestep, timestepping_info={"index": float(timestep), "delta_t": float(delta_t)})
```
angus-g, line 156: Do we have to convert the index to a float? That's weird!

demos/adjoint/inverse.py
```diff
@@ -76,7 +77,7 @@ def inverse(alpha_u, alpha_d, alpha_s):
     Tic = Function(Q1, name="Initial Temperature")
     Taverage = Function(Q1, name="Average Temperature")
 
-    checkpoint_file = CheckpointFile("Checkpoint_State.h5", "r")
```
angus-g, line 79: I guess this means you want to delete `Checkpoint_State.h5` from git and out of the requirements for the testing pipeline?

angus-g (changes requested):

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,342 @@
...
+# +
+# get the working tape
+tape = get_working_tape()
+# clear the tape from any previous annotation
+tape.clear_tape()
+# -
+# + tags=[active-ipynb]
+# # print all the blocks
+# print(tape.get_blocks())
+# -
```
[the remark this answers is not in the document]
sghelichkhani: Yeah the idea is that after clearing the tape, it should be empty. I will put this in the doc.

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,342 @@
...
+# Define the object for pyadjoint
+reduced_functional = ReducedFunctional(objective, control)
+# -
+
+# Pausing Annotation
+# ==================
+#
+# At this point, we have completed annotating the tape with the necessary information from running the forward simulation. To prevent further annotations during subsequent operations, we stop the annotation process. This ensures that no additional solves are unnecessarily recorded, keeping the tape focused only on the essential steps.
+#
+# We can then print the contents of the tape to verify what has been recorded.
```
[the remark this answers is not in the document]
sghelichkhani: I was thinking plotting the tape instead of just printing the blocks. But it requires a lot of dependencies first, so I changed my mind. I might put back a print though.

demos/adjoint/forward.py
sghelichkhani: Yeah we do! Because firedrake enforces everything to be float for attributes.

demos/adjoint/inverse.py
sghelichkhani: We didn't have `Checkpoint_State.h5` on git, did we? I think the one `.h5` file we had was for the cylindrical case. This case starts from a UFL prescribed initial condition. The only thing I have added is that for the tutorials, the user can download it instead of running it from `forward.py`.

angus-g: Yeah, my mistake, sorry!

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,342 @@
...
+#
+# ```wget https://data.gadopt.org/demos/adjoint-demo-checkpoint-state.h5```
+
+# The fields from the reference simulation are stored under the names "Temperature" and "Velocity". To retrieve the timestepping information, we can use the following code:
+
+# +
+# Importing gadopt
+from gadopt import *
+from gadopt.inverse import *
+checkpoint_filename = "adjoint-demo-checkpoint-state.h5"
```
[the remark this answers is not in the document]
sghelichkhani: Yes! That's one way of doing this. Or the one suggested for t he tutorial would be downloading the pre-run `.h5` file. and then run it. Note that for the demo I am not checkpointing to disk to avoid mesh issues (ie when number of cores for `forward.py` are different). But `inverse.py` and `taylort_test.py` are exactly as before, and checkpoint to disk.

demos/adjoint/adjoint_demo.py
sghelichkhani: Done! Thanks!

demos/adjoint/adjoint_demo.py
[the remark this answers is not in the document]
sghelichkhani: Did put back a print.

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,342 @@
+# Adjoint Inverse Demo
+# ================================
+#
+# Introduction
+# ------------
+# In this tutorial, we will demonstrate how to perform an inversion to recover the initial temperature field of the mantle using G-ADOPT.
+#
+# This demonstration involves a *twin experiment*, where we assess the performance of the inversion scheme by inverting the initial state of a synthetic reference simulation, known as the "*Reference Twin*". To create this reference twin, we run a forward mantle convection simulation and record all relevant fields (velocity and temperature) at each time step.
+#
+# We have pre-run this simulation and stored the checkpoint file on our servers. These fields serve as benchmarks for evaluating our inverse problem's performance. To download the reference benchmark checkpoint file, click
```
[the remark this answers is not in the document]
sghelichkhani: Clarified now. Thanks!

demos/adjoint/adjoint_demo.py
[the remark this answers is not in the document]
sghelichkhani: So I don't think we need to get pooch involved here necessarily.

demos/adjoint/forward.py
sghelichkhani: I agree it's a weird thing though! I think this was an oversight at the time by me, because Koki and I were just discussing having the time-step as timestepping history, not noting that this could be an integer.

sghelichkhani:
Thanks again Angus! Good for another look now.

drhodrid:
So I'll make minor changes to wording directly, but will also add some comments to this as I work through. One immediate thing, which we've seen in the past. On a mac, having installed ROL with:

`python -m pip install -v git+https://github.com/angus-g/pyrol`

When I try to import ROL I get the following error:

```
ImportError: dlopen(/Users/rhodrid/Firedrake/firedrake/lib/python3.11/site-packages/_ROL.cpython-311-darwin.so, 0x0002): Library not loaded: @rpath/librol.15.dylib
  Referenced from: <51929512-99D4-33D3-B42D-D46678CEFD9C> /Users/rhodrid/Firedrake/firedrake/lib/python3.11/site-packages/_ROL.cpython-311-darwin.so
  Reason: tried: '$ORIGIN/librol.15.dylib' (no such file), '$ORIGIN/librol.15.dylib' (no such file), '/Users/rhodrid/.pyenv/versions/3.11.9/lib/librol.15.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/rhodrid/.pyenv/versions/3.11.9/lib/librol.15.dylib' (no such file), '/opt/homebrew/lib/librol.15.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/lib/librol.15.dylib' (no such file), '/Users/rhodrid/.pyenv/versions/3.11.9/lib/librol.15.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/Users/rhodrid/.pyenv/versions/3.11.9/lib/librol.15.dylib' (no such file), '/opt/homebrew/lib/librol.15.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/lib/librol.15.dylib' (no such file)
```

The following seems to fix it:
`export DYLD_LIBRARY_PATH=/Users/rhodrid/Firedrake/firedrake/lib/python3.11/site-packages/`

... but ideally we wouldn't need the user to do this. @angus-g @sghelichkhani is this an easy fix?

angus-g:
I think you should install via pip normally... `pip install pyroltrilinos`, which uses the binary packages I've already made for Mac. If they're broken, I'll look into fixing them. Installing directly from git should be considered "developer only", imo.

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,348 @@
...
+# ================================
+#
+# Introduction
+# ------------
+# In this tutorial, we will demonstrate how to perform an inversion to recover the initial temperature field of an idealised mantle convection simulation using G-ADOPT.
+#
+# The tutorial involves a *twin experiment*, where we assess the performance of the inversion scheme by inverting the initial state of a synthetic reference simulation, known as the "*Reference Twin*". To create this reference twin, we run a forward mantle convection simulation and record all relevant fields (velocity and temperature) at each time step.
+#
+# We have pre-run this simulation by running `forward.py`, and stored model output as a checkpoint file on our servers. These fields serve as benchmarks for evaluating our inverse problem's performance. To download the reference benchmark checkpoint file, click
+# [here](https://data.gadopt.org/demos/adjoint-demo-checkpoint-state.h5"), or alternatively, execute the following command in a terminal:
```
drhodrid, line 11: Note clicking doesn't work for me. I get:
<Error>
<Code>NoSuchKey</Code>
<Message/>
<BucketName>gadopt</BucketName>
<RequestId>tx000003f30271967ee1dfb-00666157b6-dbcff29-syd1a</RequestId>
<HostId>dbcff29-syd1a-syd1-zg01</HostId>
</Error>

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,348 @@
...
+# +
+control = Control(Tic)
+
+u_misfit = 0.0
+
+# Take our initial guess and project from Q1 to Q2, simultaneously imposing temperature boundary conditions.
+T.project(Tic, bcs=energy_solver.strong_bcs)
+
+# Populate the tape by running the forward simulation.
+# For learning+tweaking purposes, feel free to change e.g., 0 -> timesteps - 3
```
drhodrid, line 170: I found this hard to underrstand. Maybe provide some more context. For learning and tweaking what?

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,348 @@
...
+# # Define the perturbation in the initial temperature field
+# import numpy as np
+# Delta_temp = Function(Tic.function_space(), name="Delta_Temperature")
+# Delta_temp.dat.data[:] = np.random.random(Delta_temp.dat.data.shape)
+#
+# # Perform the Taylor test to verify the gradients
+# minconv = taylor_test(reduced_functional, Tic, Delta_temp)
+# -
+
+# The `taylor_test` function computes the Taylor remainder and verifies that the convergence rate is close to the theoretical value of $O(2.0)$. This ensures that our gradients are accurate and reliable for optimisation.
```
drhodrid, line 301: Is it worth providing the user with some guidance here on what this test spits out - and maybe even a quick plot of the the remainders or something, just to make it more interactive?

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,348 @@
...
+optimiser = LinMoreOptimiser(
+    minimisation_problem,
+    minimisation_parameters,
+    checkpoint_dir="optimisation_checkpoint",
+)
+
+# Restore from last possible checkpoint
+# optimiser.restore()
+# Run the optimisation
+optimiser.run()
```
drhodrid, line 348: It feels like an abrupt end to a REALLY great tutorial. Why not plot the initial guess as a function of iteration count? Why not also tell the user how to access the "misfit" terms, and plot those misfits (objective functional and initial state) as a function of iteration?

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,348 @@
+# Adjoint Inverse Demo
+# ================================
+#
+# Introduction
+# ------------
+# In this tutorial, we will demonstrate how to perform an inversion to recover the initial temperature field of an idealised mantle convection simulation using G-ADOPT.
```
drhodrid, line 6: I would flag here that the total runtime for this tutorial is XX hours. And a user should have a cup of tea ready!

demos/adjoint/adjoint_demo.py
drhodrid: Having now run it, I think it's worth making it clear to the user that this Taylor test takes time. It took over 20 minutes for me, so important they are aware of this so that they don't feel that they have done something wrong. Also I think flagging what the output at the bottom means would be good (computed residuals, computed convergence). Yes it's obvious to us - but it may not be to somebody coming at this for the first time.

demos/adjoint/adjoint_demo.py
drhodrid: How long does this last step take to run for you? I had to stop it running at my end, as I need to get home. Again, perhaps worth telling the user roughly how long it is expected to take. And definitley need something to close the tutorial, as I suggested above. Otherwise - they have an optimised solution but they don't know what it looks like and how to plot various things associated with that optimisation procedure.

drhodrid (commented):
A few more comments from me. Again - a wonderful turtorial. 

One general point - given we are not running the forward model in the tutorial (the user is grabbing the checkpoint), it's probably worth including a figure or similar (can we add a movie), showing how that model starts and ends up? This is important for context.

I also think figures ouf our recovered initial condition post optimisation, and some sort of misfit, is really important. 

I learnt a lot from this! Thank you for putting it together!

drhodrid:
Last comment for now. The final run took several hours and many optimisation iterations. I think it's worth limiting the number of iterations there to perhaps 10 or so, with a comment that connvergence towards the optimised solution takes many more.

demos/adjoint/adjoint_demo.py
sghelichkhani: I am now thinking to make this tutorial for only 5 timesteps, and explain to the user that for replicating the study in Ghelichkhan et al you should change the `initial_timestep` time-step to 0 instead of `timesteps - 5`.

demos/adjoint/adjoint_demo.py
sghelichkhani: A typing issue! Corrected. now

demos/adjoint/adjoint_demo.py
```diff
@@ -0,0 +1,348 @@
...
+# Load the reference initial state - i.e. the state that we wish to recover
+Tic_ref = checkpoint_file.load_function(mesh, "Temperature", idx=int(temperature_timestepping_info["index"][0]))
+Tic_ref.rename("Reference Initial Temperature")
+checkpoint_file.close()
+
+# and subsequently visualising using standard VTK software, such as Paraview or pyvista:
+
+# + tags=["active-ipynb"]
+# import pyvista as pv
+# dataset = pv.read('./visualisation_vtk.pvd')
```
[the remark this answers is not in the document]
sghelichkhani: Done

demos/adjoint/adjoint_demo.py
sghelichkhani: Sorry I should have been more clear here. This is just to say for the purpose of learning, the user can do this exercise for only 5 timesteps. Everything would run fine for that. But in our experiment in Ghelichkhan et al 2024 we had this as 0.

demos/adjoint/adjoint_demo.py
sghelichkhani: The issue here is that we cannot easily retrieve the residuals + perturbations from taylor test to automatically plot them, it all gets dumped to stdout. Unless I redirect `sys.stdout` to a variable, which I think would make the whole thing confusing!

drhodrid:
Hey @angus-g  - this is looking now. I am going to make a few minor text edits -- to help me better explain things on the day -- but a few minor things for you in the meantime:
1. Could you get the wget setup as you think is required? I think it perhaps needs to be a part of the test anyway, because ultimate this will be tested through our regression suite? Maybe I'm wrong!
2. The Trame error on visualisation.

sghelichkhani:
> 1. Could you get the wget setup as you think is required? I think it perhaps needs to be a part of the test anyway, because ultimate this will be tested through our regression suite? Maybe I'm wrong!
> 

The regression tests for adjoint still rely on previous scripts on the repository, e.g., 'inverse.py'( remember the discussion about the adjoint tests  being a function getting called, and that a function can turn into a single jupyter notebook cell). 

So I suggest we introduce a unit test that downloads the mesh filr, instead. That should be very straightforward.

angus-g:
I think this is just about good to go now, but I did see a random failure in `tosi_2015`:
```
 run_benchmark.py benchmarks.tosi_2015 failed:
          ^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/variational_solver.py", line 329, in solve
          ^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/variational_solver.py", line 329, in solve
  File "/home/firedrake/firedrake/src/firedrake/firedrake/variational_solver.py", line 329, in solve
  File "/home/firedrake/firedrake/src/firedrake/firedrake/variational_solver.py", line 329, in solve
          ^^^^^^^^^^^^^^^^^^^^^
  File "/home/firedrake/firedrake/src/firedrake/firedrake/variational_solver.py", line 329, in solve
    solving_utils.check_snes_convergence(self.snes)
  File "/home/firedrake/firedrake/src/firedrake/firedrake/solving_utils.py", line 124, in check_snes_convergence
```

I've retriggered a run just to be sure (definitely need to tweak the workflow to give proper outputs for these runtime failures that still result in a passing test...)

angus-g (approved):

PR #81, opened 2024-06-04
Use default dx in analytical cases

drhodrid:
Just for testing purposes. Are smooth analytical benchmark comparisons sensitive to dx?

### DIFF
--- a/demos/analytical_comparisons/delta_cylindrical_freeslip.py
+++ b/demos/analytical_comparisons/delta_cylindrical_freeslip.py
@@ -2,9 +2,6 @@ from gadopt import *
 import numpy
 import assess
 
-# Quadrature degree:
-_dx = dx(degree=6)
-
 # Projection solver parameters for nullspaces:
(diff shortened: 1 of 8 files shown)


drhodrid:
Just committed changes to the delta function cases too. Looking promising. Can we also get these tested?

On 4 Jun 2024, at 9:09 PM, Angus Gibson ***@***.***> wrote:



See https://github.com/g-adopt/g-adopt/actions/runs/9365997227

angus-g:
> Just committed changes to the delta function cases too. Looking promising. Can we also get these tested?

Looks like those changes passed too.

drhodrid:
OK - in that case let’s merge this bad boy - and I’ll look next at the tutorials!!

On 6 Jun 2024, at 2:50 PM, Angus Gibson ***@***.***> wrote:



Just committed changes to the delta function cases too. Looking promising. Can we also get these tested?

Looks like those changes passed too.

angus-g (approved):

PR #82, opened 2024-06-06
Remove unused _dx from parallel scaling.

drhodrid:
As it says above. Want to verify this doesn't impact our testing, but can't see how it would. @angus-g - probably a good time to update test values for parallel scaling too?


angus-g:
Yeah, I guess I'll use the last long test run (https://github.com/g-adopt/g-adopt/actions/runs/9365997227) as our "baseline" and then see how this compares.

drhodrid:
Sounds good to me! Thank you.

angus-g (approved):
Parallel scaling still passed with the dx change, and we're back to green on long tests with the adjustment to iterations/solve times to match current expectations (https://github.com/g-adopt/g-adopt/actions/runs/9395970182).

PR #83, opened 2024-06-06
Include rho_continuity in MassInvPC

stephankramer:
For compressible cases where rho_continuity/=1 this should be taken into account the scaled mass preconditioner for the Schur complement:

$$\nabla\bar\rho\cdot \circ \left(\nabla \mu \nabla \cdot\right)^{-1} \circ \nabla \approx \bar\rho/\mu \nabla^2$$

Slightly improves n/o outer iterations in (some?) compressible cases.


drhodrid (approved):
A nice easy change that appears to improve things by around 25% for the compressible cases I've run. Good to go pending a green test suite.

sghelichkhani (approved):

PR #84, opened 2024-06-11
Add diagram showing relationship of demos

angus-g:
I've already put a `.pages` file in the demos directory that is used for the table of contents that shows up on the documentation site. I think that, and this diagram make more sense here because they can be updated alongside the demos.

Diagram is rendered here: https://10f9be33.g-adopt.pages.dev/tutorials/

Also has links to the tutorials that show up in the hierarchy, plus adjoint on the side (without a link until #80 goes in).

### DIFF
--- a/demos/.diagram.mermaid
+++ b/demos/.diagram.mermaid
@@ -1,0 +1,38 @@
+graph TD
+  base[Base]
+
+  subgraph physical[Physical]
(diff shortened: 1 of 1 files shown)


drhodrid (approved):
Looks blooming glorious!

w3168:
very nice!

PR #85, opened 2024-06-18
Assigning 1D profiles from files to functions

sghelichkhani:
Adding `assign_1d_profile` to gadopt. This is practical as many of our parameters in real Earth simulations are derived from 1d profiles obtained from various disciplines. Examples are 1d viscosity profile, thermal expansion, gravity. 

There are also two tests, one for the assignment of 1d average, one for the 1d averaging.

### DIFF
--- a/gadopt/utility.py
+++ b/gadopt/utility.py
@@ -16,6 +16,7 @@ import logging
 from logging import DEBUG, INFO, WARNING, ERROR, CRITICAL  # NOQA
 import os
 from scipy.linalg import solveh_banded
+from scipy.interpolate import interp1d
 
 # TBD: do we want our own set_log_level and use logging module with handlers?
 log_level = logging.getLevelName(os.environ.get("GADOPT_LOGLEVEL", "INFO").upper())
@@ -45,6 +46,7 @@ class ParameterLog:
 class TimestepAdaptor:
     """
     Computes timestep based on CFL condition for provided velocity field"""
+
     def __init__(self, dt_const, u, V, target_cfl=1.0, increase_tolerance=1.5, maximum_timestep=None):
         """
         :arg dt_const:      Constant whose value will be updated by the timestep adaptor
@@ -475,3 +477,56 @@ def node_coordinates(function):
     return [
         Function(func_space).interpolate(coords).dat.data for coords in mesh_coords
     ]
+
+
+def assign_1d_profile(function: Function, one_d_filename: str, cartesian: bool, r1d=None):
+    """
+    Assign a one-dimensional profile to a Function `q` from a file.
+
+    The function reads a one-dimensional radial viscosity profile together with the
+    radius input for the profile, from a file, broadcasts the read array to all
+    processes, and then interpolates this array onto the function space of `q`.
+
+    Args:
+        q (Function): The function onto which the 1D profile will be assigned.
+        one_d_filename (str): The path to the file containing the 1D radial viscosity profile.
+        cartesian (bool): Is the upward direction along z/y(cartesian=True) or radial (cartesian=False)
+
+    Returns:
+        None: It directly assigns the input function `function`.
+
+    Note:
+        - Note the cartesian flag
+        - This is designed to read a file with one process and distribute in parallel with MPI.
+        - The input file should contain an array of radius and an array of viscosity values, seperated by comma.
+        - Make sure radius in the file is consistent with the mesh.
+    """
+    # find the mesh
+    mesh = extract_unique_domain(function)
+
+    visc = None
+    rshl = None
+
+    # read the input file
+    if mesh.comm.rank == 0:
+        # The root process reads the file
+        rshl, visc = np.loadtxt(one_d_filename, unpack=True, delimiter=",")
+
+    # Broadcast the entire 'visc' array to all processes
+    visc = mesh.comm.bcast(visc, root=0)
+    # Similarly, broadcast 'rshl' if needed (assuming all processes need it)
+    rshl = mesh.comm.bcast(rshl, root=0)
+
+    # Coordinates
+    X = SpatialCoordinate(mesh)
+
+    if cartesian:
+        # if in a two-dimensional problem: y otherwise z is the direction
+        upward_coord = X[1] if len(X) == 2 else X[2]
+    else:
+        upward_coord = sqrt(X**2)
+
+    rad = Function(function.function_space()).interpolate(upward_coord)
+
+    averager = LayerAveraging(mesh, rshl if mesh.layers is None else None, cartesian=cartesian)
+    averager.extrapolate_layer_average(function, interp1d(rshl, visc, fill_value="extrapolate")(averager.get_layer_average(rad)))
--- a/tests/test_one_d_assigner.py
+++ b/tests/test_one_d_assigner.py
@@ -1,0 +1,44 @@
+
+
+from gadopt import *
+
+
+def test_oned_average_assignment():
+    # a 10 x 10 square mesh
+    n = 10
+    mesh = UnitSquareMesh(n, n)
+    # because this is not an extruded mesh we have to provide r1d for layer averaging
+    y_disc = np.linspace(0.0, 1.0, n+1)
+    # spatial coordinates and computing vertical coordinate
+    X = SpatialCoordinate(mesh)
+    Q = FunctionSpace(mesh, "CG", 1)
+    q = Function(Q).interpolate(X[1])
+    # computing the average
+    averager = LayerAveraging(mesh, y_disc if mesh.layers is None else None, cartesian=True)
+    q_profile = averager.get_layer_average(q)
+    # write the average out
+    output_fi = ParameterLog("one_d_test.output", mesh)
+    output_fi.log_str("\n".join([f"{y}, {val}" for y, val in zip(y_disc, q_profile)]))
+    output_fi.close()
+    # assign the average to a new function
+    p = Function(Q)
+    assign_1d_profile(p, "one_d_test.output", cartesian=True)
+
+    assert assemble((p-q) ** 2 * dx) < 1e-10
+
+
+def test_layer_averaging():
+    # a 10 x 10 square mesh
+    n = 10
+    mesh = UnitSquareMesh(n, n)
+    # because this is not an extruded mesh we have to provide r1d for layer averaging
+    y_disc = np.linspace(0.0, 1.0, n+1)
+    # spatial coordinates and computing vertical coordinate
+    X = SpatialCoordinate(mesh)
+    Q = FunctionSpace(mesh, "CG", 2)
+    q = Function(Q).interpolate(X[1])
+    # computing the average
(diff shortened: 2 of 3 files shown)


gadopt/utility.py
```diff
@@ -475,3 +477,56 @@ def node_coordinates(function):
...
+    if cartesian:
+        # if in a two-dimensional problem: y otherwise z is the direction
+        upward_coord = X[1] if len(X) == 2 else X[2]
+    else:
+        upward_coord = sqrt(X**2)
+
+    rad = Function(function.function_space()).interpolate(upward_coord)
+
+    averager = LayerAveraging(mesh, rshl if mesh.layers is None else None, cartesian=cartesian)
+    averager.extrapolate_layer_average(function, interp1d(rshl, visc, fill_value="extrapolate")(averager.get_layer_average(rad)))
```
Patol75, line 532: It would be more readable if the interpolator object was defined outside of the method call. Otherwise, you could add `.__call__`.

Patol75: [SciPy's doc](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html#scipy.interpolate.interp1d) states 

> This class is considered legacy and will no longer receive updates. This could also mean it will be removed in future SciPy versions.

Suggestions for replacement can be found [here](https://docs.scipy.org/doc/scipy/tutorial/interpolate/1D.html#tutorial-interpolate-1dsection).

gadopt/utility.py
```diff
@@ -475,3 +477,56 @@ def node_coordinates(function):
...
+    # find the mesh
+    mesh = extract_unique_domain(function)
+
+    visc = None
+    rshl = None
+
+    # read the input file
+    if mesh.comm.rank == 0:
+        # The root process reads the file
+        rshl, visc = np.loadtxt(one_d_filename, unpack=True, delimiter=",")
```
Patol75, line 513: The two assignments at the top usually come as part of an `else` block. It adds to readability but is not mandatory, so it is up to you.

Patol75 (changes requested):
That should be a useful addition, thank you. A few first comments from my side.

gadopt/utility.py
```diff
@@ -475,3 +477,56 @@ def node_coordinates(function):
...
+    Assign a one-dimensional profile to a Function `q` from a file.
+
+    The function reads a one-dimensional radial viscosity profile together with the
+    radius input for the profile, from a file, broadcasts the read array to all
+    processes, and then interpolates this array onto the function space of `q`.
+
+    Args:
+        q (Function): The function onto which the 1D profile will be assigned.
+        one_d_filename (str): The path to the file containing the 1D radial viscosity profile.
+        cartesian (bool): Is the upward direction along z/y(cartesian=True) or radial (cartesian=False)
```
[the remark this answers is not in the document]
sghelichkhani: `r1d` is actually redundant now. It works well with rsh.

gadopt/utility.py
```diff
@@ -475,3 +477,56 @@ def node_coordinates(function):
...
+
+    Returns:
+        None: It directly assigns the input function `function`.
+
+    Note:
+        - Note the cartesian flag
+        - This is designed to read a file with one process and distribute in parallel with MPI.
+        - The input file should contain an array of radius and an array of viscosity values, seperated by comma.
+        - Make sure radius in the file is consistent with the mesh.
```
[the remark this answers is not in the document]
sghelichkhani: Ah! What I mean hear is that because interp1d (now np.interp )does not raise an error for out of bound values, one should be careful about the dimensions. `LayerAveraging` works with whatever value comes out of `interp`

gadopt/utility.py
sghelichkhani: Changing to `np.interp`. Thanks.

gadopt/utility.py
```diff
@@ -475,3 +477,56 @@ def node_coordinates(function):
     return [
         Function(func_space).interpolate(coords).dat.data for coords in mesh_coords
     ]
+
+
+def assign_1d_profile(function: Function, one_d_filename: str, cartesian: bool, r1d=None):
```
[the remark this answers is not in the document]
sghelichkhani: Changed to `interpolate_1d_profile`

sghelichkhani:
Thanks @Patol75! Very useful.

gadopt/utility.py
```diff
@@ -475,3 +476,56 @@ def node_coordinates(function):
...
         Function(func_space).interpolate(coords).dat.data for coords in mesh_coords
     ]
+
+
+def interpolate_1d_profile(function: Function, one_d_filename: str, cartesian: bool):
+    """
+    Assign a one-dimensional profile to a Function `function` from a file.
+
+    The function reads a one-dimensional profile (e.g., viscosity) together with the
+    radius input for the profile, from a file, broadcasts the two arrays to all
```
drhodrid, line 486: depth / radius (rather than just radius)

drhodrid (commented):
I think this is fine, but perhaps @angus-g could quickly have a look through one you are done. One comment to bear in mind for PyGplates merge and the associated tutorial: it would be nice if for that example perhaps we use a simple 1-D viscosity profile, which is read from a file - then we could test the resulting viscosity profile, ensuring ongoing functionality on a sphere (noting taht the tests you've added here are Cartesian).

gadopt/utility.py
```diff
@@ -475,3 +476,56 @@ def node_coordinates(function):
...
+    processes, and then interpolates this array onto the function space of `function`.
+
+    Args:
+        function (Function): The function onto which the 1D profile will be assigned.
+        one_d_filename (str): The path to the file containing the 1D radial profile.
+        cartesian (bool): Is the upward direction along z/y(cartesian=True) or radial (cartesian=False)
```
angus-g, line 492:
```suggestion
        function: The function onto which the 1D profile will be assigned
        one_d_filename: The path to the file containing the 1D radial profile
        cartesian: Whether the upward direction is along z/y (True) or radial (False)
```

gadopt/utility.py
```diff
@@ -475,3 +476,56 @@ def node_coordinates(function):
...
+
+    Returns:
+        None: It directly interpolates the input function `function`.
+
```
angus-g, line 496:
```suggestion
```

gadopt/utility.py
```diff
@@ -475,3 +476,56 @@ def node_coordinates(function):
...
+    Note:
+        - Note the cartesian flag
+        - This is designed to read a file with one process and distribute in parallel with MPI.
+        - The input file should contain an array of radius and an array of values, seperated by comma.
```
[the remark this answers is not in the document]
angus-g: sep**a**rated

gadopt/utility.py
```diff
@@ -475,3 +476,56 @@ def node_coordinates(function):
...
+    rshl = mesh.comm.bcast(rshl, root=0)
+
+    # Coordinates
+    X = SpatialCoordinate(mesh)
+
+    if cartesian:
+        # if in a two-dimensional problem: y otherwise z is the direction
+        upward_coord = X[1] if len(X) == 2 else X[2]
+    else:
+        upward_coord = sqrt(X**2)
```
angus-g, line 525: Don't we have `vertical_component` to do this already?

tests/test_one_d_assigner.py
```diff
@@ -0,0 +1,44 @@
...
+    y_disc = np.linspace(0.0, 1.0, n+1)
+    # spatial coordinates and computing vertical coordinate
+    X = SpatialCoordinate(mesh)
+    Q = FunctionSpace(mesh, "CG", 2)
+    q = Function(Q).interpolate(X[1])
+    # computing the average
+    averager = LayerAveraging(mesh, y_disc if mesh.layers is None else None, cartesian=True)
+    q_profile = averager.get_layer_average(q)
+
+    assert np.allclose(q_profile, y_disc, atol=1e-10)
```
angus-g, line 44: Given that this is in this file, I don't think `test_one_d_assigner.py` is an accurate description.

tests/test_one_d_assigner.py
```diff
@@ -0,0 +1,44 @@
...
+    # computing the average
+    averager = LayerAveraging(mesh, y_disc if mesh.layers is None else None, cartesian=True)
+    q_profile = averager.get_layer_average(q)
+    # write the average out
+    output_fi = ParameterLog("one_d_test.output", mesh)
+    output_fi.log_str("\n".join([f"{y}, {val}" for y, val in zip(y_disc, q_profile)]))
+    output_fi.close()
+    # assign the average to a new function
+    p = Function(Q)
+    interpolate_1d_profile(p, "one_d_test.output", cartesian=True)
```
angus-g, line 25: Should use `tmp_path` fixture and write the file there so the test doesn't leave any artifacts.

angus-g (changes requested):

gadopt/utility.py
```diff
@@ -475,3 +476,56 @@ def node_coordinates(function):
...
+        - This is designed to read a file with one process and distribute in parallel with MPI.
+        - The input file should contain an array of radius and an array of values, seperated by comma.
+    """
+    # find the mesh
+    mesh = extract_unique_domain(function)
+
+    # read the input file
+    if mesh.comm.rank == 0:
+        # The root process reads the file
+        rshl, visc = np.loadtxt(one_d_filename, unpack=True, delimiter=",")
```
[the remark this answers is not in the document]
sghelichkhani: Done

gadopt/utility.py
sghelichkhani: Correct.

gadopt/utility.py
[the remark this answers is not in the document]
sghelichkhani: Honestly, I think seperated should be the correct form. But whatever!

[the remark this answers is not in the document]
sghelichkhani: I don't think you can make it general. So this should be radius only. It can be in any order (decreasing or increasing) but it can be only radius.

gadopt/utility.py
sghelichkhani: Just Radius.

sghelichkhani (commented):
Ready for another look.

Patol75 (commented):
Generally looking good now, but I think there are still a couple of comments from Angus that need to be addressed.

tests/test_one_d_assigner.py
angus-g: And to follow on from this, this should now go in the `tests/unit` directory.

sghelichkhani: Seperated now.

sghelichkhani: Separated!

sghelichkhani (commented):
Good for a final look

drhodrid (commented):
I'm happy if the gurus are happy!

angus-g (approved):
Looks good to me, thanks!

Patol75 (approved):

drhodrid (approved):

PR #87, opened 2024-06-25
Add min/max temperature diagnostics

drhodrid:
Add a minimum and maximum temperature diagnostic to diagnostics.py. This is useful for DG simulations, in particular, where our BCs are imposed weakly and min and max T's are therefore not guaranteed to be 0 and 1. 

I have updated 2d cylindrical demo to include a test of these. @angus-g could you update the pickle file for that demo - values for T_min should be 0 and values for T_max should be 1.


drhodrid:
Note that changes to docstrings are all covered in another branch (e.g. quad_degree) and will be part of a later pull request.

Patol75 (approved):
Looking good.

angus-g:
> could you update the pickle file for that demo - values for T_min should be 0 and values for T_max should be 1

i.e. you want me to add this to the existing testing?

drhodrid:
Just needs an update to the pickle file then @angus-g and this should be good to go.

angus-g:
It doesn't need an update to pass the test...

drhodrid:
Yes please. Just ensure that T_min at the end is 0 and T_max is 1.

PR #89, opened 2024-07-07
Allowing for variable gravity

sghelichkhani:
In (truncated) anelastic liquid approximation, the reference gravitational acceleration is assumed to be a function of depth and can vary. In our codebase for these approximations (for reasons that are unclear to me) we do not allow g to be user-defined. This PR simply removes this constraint.


stephankramer (changes requested):
Yeah can't remember why I put in that assert. My guess is that I hadn't really thought through whether the non-dimensionalisation was consistent in all places. On a quick check I do see, that indeed the dissipation term in the energy equation is missing a gbar (i.e. the `linearized_energy_sink()` method) other than that I think it should be fine (comparing with Schubert). So if you just add that in, I think it's good to go

As an aside, in the docstring I think our labelling of "supported entries in **kwargs" as "keyword arguments" a bit misleading, as any argument with a default value are also commonly referred to as such, and for the user there shouldn't really be any difference between the two. I think technically, the definition of "keyword argument" is in fact based on whether there is a keyword on the calling side (the "argument" is really what's been provided at the calling side, not the name of the "parameter" on the function call side). I see that sphinx somehow does do the right thing. I can see the need for splitting explicitly named parameters from the ones that are gobbled up by **kwargs - basically because the user will see the **kwargs - but when I'm reading the docstring in text and I see the parameters split in "arguments" and "keyword arguments" that makes me think that I should be using "arguments" as positional arguments. Could we maybe use "Other keyword arguments:" instead?

drhodrid:
I believe all comments have now been addressed, excluding those in relation to the docstrings. These will be discussed at next week's dev meeting and updated in a later pull request. I think this is good to go now, but would appreciate another quick check/confirmation.

Patol75 (approved):
This should be good to go, provided that CI returns green.

angus-g:
On the Python side, the parameters should make a bit more sense (they're labelled "other arguments" for those that are inherited). I think the intention is to also make a more friendly table that clarifies which parameters belong to which approximation, and their defaults. I'll push this in for now, and we can put documentation updates in our next iteration as we go.

PR #90, opened 2024-07-09
Update citations

w3168:
updated citations in readme file since sia and co's paper got published https://gmd.copernicus.org/articles/17/5057/2024/gmd-17-5057-2024.html

I've also updated the bibase list separately which automatically updates on the website


drhodrid (approved):
Brilliant. Thank you.

PR #93, opened 2024-07-11
Ensure strictly increasing data for interp

Patol75:
`interpolate_1d_profile` relies on `numpy.interp` for one-dimensional interpolation. However, it currently does not check if the user-provided data is strictly monotonous and if it is given in increasing order. These missing checks can produce silently meaningless results.


drhodrid:
Confirming that this fixes the issue I've seen. Thanks Thomas.

angus-g (approved):
Looks sensible. Thanks!

PR #95, opened 2024-07-13
Force SuperLU_DIST on compressible demos

angus-g:
There's some kind of parallel ordering bug with the MUMPS+METIS combo for direct solves that seems to pop up in particular on these test cases. Using a different solver (e.g. SuperLU_DIST) avoids this issue. The other option would be PaStiX, but it segfaults on the compressible demos. We can't enable superlu globally, because that affects convergence on one of our tests, causing the CI to time out (see https://github.com/g-adopt/g-adopt/actions/runs/9900787219/job/27352140174). To avoid modifying the demo directly, we perform the solver override only when running the demo in its test configuration.

### DIFF
--- a/demos/2d_compressible_ALA/Makefile
+++ b/demos/2d_compressible_ALA/Makefile
@@ -4,7 +4,7 @@ ncpus := 4
 
 params.log: 2d_compressible_ALA.py
 	echo "running $<" >&2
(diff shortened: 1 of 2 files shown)


drhodrid (approved):
Looks good to me, and hopefully leads to more stability in our tests. Is it worth also reporting this behaviour to MUMPS/PETSc devs, just so that they are aware?

PR #98, opened 2024-07-18
Test fix to solution of DG energy equation, under compressible approximations

drhodrid:
Putting up a pull request mainly to determine if this breaks anything else (although admittedly we don't currently test for DG extensively, so suspect all will be ok!).

### DIFF
--- a/gadopt/scalar_equation.py
+++ b/gadopt/scalar_equation.py
@@ -160,7 +160,7 @@ class ScalarDiffusionTerm(BaseTerm):
                 # this corresponds to the same 3 terms as the dS integrals for DG above:
                 F += 2*sigma_ext*phi*inner(n, dot(diff_tensor, n))*jump_q*self.ds(id)
                 F += -inner(dot(diff_tensor, grad(phi)), n)*jump_q*self.ds(id)
-                F += -inner(phi*n, dot(diff_tensor, grad(trial))) * self.ds(id)
+                F += -inner(phi*n, dot(diff_tensor, grad(q))) * self.ds(id)
                 if 'flux' in bc:
                     raise ValueError("Cannot apply both `q` and `flux` bc on same boundary")
             elif 'flux' in bc:
--- a/gadopt/scalar_equation.py
+++ b/gadopt/scalar_equation.py
@@ -149,12 +149,12 @@ class ScalarDiffusionTerm(BaseTerm):
 
         for id, bc in bcs.items():
             if 'q' in bc:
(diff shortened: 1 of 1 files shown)


gadopt/scalar_equation.py
```diff
@@ -149,12 +149,12 @@ def residual(
...
         for id, bc in bcs.items():
             if 'q' in bc:
-                jump_q = q-bc['q']
+                jump_q = trial-bc['q']
                 sigma_ext = sigma * fd.FacetArea(self.mesh)/fd.CellVolume(self.mesh)
                 # this corresponds to the same 3 terms as the dS integrals for DG above:
                 F += 2*sigma_ext*phi*inner(n, dot(diff_tensor, n))*jump_q*self.ds(id)
                 F += -inner(dot(diff_tensor, grad(phi)), n)*jump_q*self.ds(id)
-                F += -inner(phi*n, dot(diff_tensor, grad(q))) * self.ds(id)
+                F += -inner(phi*n, dot(diff_tensor, grad(trial))) * self.ds(id)
```
Patol75, line 157: Just pasting my thought from Zulip that this change should not be necessary.

stephankramer: Yes, I think I agree. Basically we've written down a DG diffusion term for q  = Tbar + T' (where Tbar='reference_for_diffusion' and T'=trial) but then the boundary value we impose should also be in terms of q, so q_bc = Tbar_bc + T'_bc. However the value in `bc['q']` is being interpreted as T' to be consistent with the strong bc case for CG (where the DirichletBC value is always the value for the field we are solving for, so T') - but also we want to keep it that way for the advection term which is formulated purely in terms of T'. But the way to think of jump_q then is: `jump_q = q - q_bc = (Tbar + T') - (Tbar_bc + T'_bc) = T' - T'_bc = trial - bc['q']`. This line 157 is really just the term that comes from integration by parts of `phi*div(kappa*grad(q))` - so we should keep it like that - and the other terms are really the symmetrisation and penalty term that penalize a jump in q=Tbar + T' except the jump in Tbar is zero because it's prescribed both in the interior and on the boundary.

drhodrid:
Thanks @Patol75 and @stephankramer for comments. I believe I've address them with my latest change. If testing goes green, this requires an approval and we should be good to go. I will add a demo that tests this functionality early next week (I currently have a draft PR: https://github.com/g-adopt/g-adopt/pull/101 -- which needs a little more work).

Patol75 (approved):

PR #99, opened 2024-07-19
Make `cartesian` a mesh attribute

angus-g:
There are several places where we would like to know whether the domain is Cartesian or not, and forgetting to specify this can cause silently confusing behaviour. It seems to make sense that the mesh itself should know whether it represents a Cartesian domain, and indeed a mesh is present in all places where previously a `cartesian` kwarg would have been required.

The usage of this approach is fairly simple:
```python
mesh = SomeMeshFunction(...)
mesh.cartesian = True # or False
```
From here on, the `vertical_component` or `upward_normal` functions can query the mesh directly, and it doesn't have to be passed around as a flag. In the case when the `cartesian` attribute is *not* present on the mesh, I think this will fail with an `AttributeError`. Perhaps this is better than silently making an assumption about the domain, but it could certainly do with a more informative error message in this case.

Supersedes #74.

### DIFF
--- a/demos/2d_cylindrical/2d_cylindrical.py
+++ b/demos/2d_cylindrical/2d_cylindrical.py
@@ -33,11 +33,14 @@ from gadopt import *
 # using the optional keyword argument `extrusion_type`, forming 32 layers. To better represent the
 # curvature of the domain and ensure accuracy of our quadratic representation of velocity, we
 # approximate the curved cylindrical shell domain quadratically, using the optional keyword argument `degree`$=2$.
+# Because this problem is not formulated in a Cartesian geometry, we set the `mesh.cartesian`
+# attribute to False. This ensures the correct configuration of a radially inward vertical direction.
 
 # +
 rmin, rmax, ncells, nlayers = 1.22, 2.22, 128, 32
 mesh1d = CircleManifoldMesh(ncells, radius=rmin, degree=2)  # construct a circle mesh
 mesh = ExtrudedMesh(mesh1d, layers=nlayers, extrusion_type='radial')  # extrude into a cylinder
+mesh.cartesian = False
 bottom_id, top_id = "bottom", "top"
 
 V = VectorFunctionSpace(mesh, "CG", 2)  # Velocity function space (vector)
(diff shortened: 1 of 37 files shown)


demos/2d_cylindrical/2d_cylindrical.py
```diff
@@ -38,6 +38,7 @@
 rmin, rmax, ncells, nlayers = 1.22, 2.22, 128, 32
 mesh1d = CircleManifoldMesh(ncells, radius=rmin, degree=2)  # construct a circle mesh
 mesh = ExtrudedMesh(mesh1d, layers=nlayers, extrusion_type='radial')  # extrude into a cylinder
+mesh.cartesian = False
```
angus-g, line 41: Since the explanation about cartesian is removed later in this file, I guess there should be something in the cell above about this change.

demos/3d_spherical/3d_spherical.py
```diff
@@ -29,6 +29,7 @@
 
 mesh2d = CubedSphereMesh(rmin, refinement_level=ref_level, degree=2)
 mesh = ExtrudedMesh(mesh2d, layers=nlayers, extrusion_type='radial')
+mesh.cartesian = False
```
angus-g, line 32: Same comment as `2d_cylindrical`: maybe this should be explained in the cell above.

demos/2d_cylindrical/2d_cylindrical.py
drhodrid: Agreed. Perhaps update the text above to include:

# We specify that the problem is not executed in a Cartesian geometry, by setting the mesh.cartesian argument to false. This ensures that boundary conditions and specification of the vertical/radial direction are configured appropriately.

demos/3d_spherical/3d_spherical.py
drhodrid: See previous comment. I would also suggest in the base case, where this flag is specified for the first time, that we flag it is a requirement and why.

drhodrid: Perhaps saying something like:
      cartesian:
        - True: gravity is assumed to point in the negative z-direction
        - False: gravity is assumed to point radially inward

drhodrid (commented):
A very nice change, that will ensure I (and many others!) don't run simulations with inconsistent Cartesian definitions!! Minor comments from me - but perhaps worth somebody more technical approving.

Patol75 (approved):
Looks great, thanks @angus-g.

PR #107, opened 2024-09-04
Fixing a minor issue with adjoint demo

sghelichkhani:
In the adjoint demo, we intend to use the observed temperature field at the final time-step (i.e. final temperature field of the forward model) as our initial guess for the initial condition, i.e. `Tic`. When loading `Tic` we use  `idx=timesteps-1`, which does not exist, and oddly leaves `Tic` as is instead of raising an error. This is problematic since we enforce zero sensitivity for our derivatives at top and bottom boundaries by using a Dirichlet boundary condition. 

To reproduce the issue:
The above issue is not visible when having damping and smoothing terms in the objective functional. To see the effect, simply set all the other terms except for `t_misfit` in. One will see the bottom boundary temperature to be zero and a strange halo around the hot anomaly.


drhodrid (commented):
The tutorial session still ran well, but an important fix! Does this change the required values for testing too? Or are we not checking that for this specific demo? If not, perhaps we should?

Patol75:
Should this be fixed in [inverse.py](https://github.com/g-adopt/g-adopt/blob/281e38e97816b2ea01e9cce28276fdf7f5e63b07/tests/adjoint/inverse.py#L84) and [taylor_test.py](https://github.com/g-adopt/g-adopt/blob/281e38e97816b2ea01e9cce28276fdf7f5e63b07/tests/adjoint/taylor_test.py#L70) too?

Patol75:
Also, should [this](https://github.com/g-adopt/g-adopt/blob/281e38e97816b2ea01e9cce28276fdf7f5e63b07/demos/adjoint/adjoint.py#L129) be `timesteps = int(temperature_timestepping_info["index"][-1]) + 1`? The current value is 79, and I wonder if it should be 80 instead.

sghelichkhani:
> The tutorial session still ran well, but an important fix! Does this change the required values for testing too? Or are we not checking that for this specific demo? If not, perhaps we should?

The demo we are running replicates the best performing case in our paper, and the regression test we are testing for. The reason why we did not see the differences is that the demo runs only for 5 timesteps only, just make things feasible for a tutorial session. 

So my answer:  I believe our current adjoint testing covers all bases. So adding this would be redundant.

sghelichkhani:
> Should this be fixed in [inverse.py](https://github.com/g-adopt/g-adopt/blob/281e38e97816b2ea01e9cce28276fdf7f5e63b07/tests/adjoint/inverse.py#L84) and [taylor_test.py](https://github.com/g-adopt/g-adopt/blob/281e38e97816b2ea01e9cce28276fdf7f5e63b07/tests/adjoint/taylor_test.py#L70) too?

No! There we have timesteps fixed as 80, so everything stays consistent. This is perhaps relevant to the other issue you raised, which is the correct way to go.

sghelichkhani:
> Also, should [this](https://github.com/g-adopt/g-adopt/blob/281e38e97816b2ea01e9cce28276fdf7f5e63b07/demos/adjoint/adjoint.py#L129) be `timesteps = int(temperature_timestepping_info["index"][-1]) + 1`? The current value is 79, and I wonder if it should be 80 instead. If that is the case, then the proposed change to `idx` is not required and the above two files should be fine as-is.

You are correct about this. I will change the definition of time-steps and keep everything else the same.

Issue #105, opened 2024-08-28
Warning in gplates demo


angus-g:
As mentioned in #104:
```
/home/firedrake/firedrake/lib/python3.12/site-packages/gadopt/gplates/gplates.py:407: RuntimeWarning: Issues finding plate ids for some of the seeds.
  warnings.warn("Issues finding plate ids for some of the seeds.", category=RuntimeWarning)
```

Patol75:
Given gh-109, can we close this one?

PR #109, opened 2024-09-12
Allow for more generic configuration of plate reconstruction datasets

sghelichkhani:
Obvious from the title, this will provide access to different versions of the plate reconstruction model Muller et al 2022.

### DIFF
--- a/demos/gplates_global/gplates_global.py
+++ b/demos/gplates_global/gplates_global.py
@@ -141,7 +141,7 @@ Z_near_nullspace = create_stokes_nullspace(Z, closed=False, rotational=True, tra
 # directory. Below, we verify the required paths in this directory and
 # ensure they exist:
 
-muller_2022_files = obtain_Muller_2022_SE(".")
+muller_2022_files = obtain_Muller_2022_SE_v1_2(".")
 
 # + tags=["active-ipynb"]
 # # These are the files that needs to be passed on to pyGPlates
--- a/demos/gplates_global/gplates_global.py
+++ b/demos/gplates_global/gplates_global.py
@@ -137,11 +137,11 @@ Z_near_nullspace = create_stokes_nullspace(Z, closed=False, rotational=True, tra
 # tutorial, we will use the study published by Muller et al., 2022.
 # The files can be downloaded from EarthByte's server at:
 # https://earthbyte.org/webdav/ftp/Data_Collections/Muller_etal_2022_SE/Muller_etal_2022_SE_1Ga_Opt_PlateMotionModel_v1.2.zip
-# Download and unzip this file into the `./gplates_files`
+# Download and unzip this file into the current
 # directory. Below, we verify the required paths in this directory and
 # ensure they exist:
 
-muller_2022_files = obtain_Muller_2022_SE_v1_2(".")
+muller_2022_files = ensure_reconstruction("Muller 2022 SE v1.2", ".")
 
 # + tags=["active-ipynb"]
 # # These are the files that needs to be passed on to pyGPlates
--- a/gadopt/gplates/gplates.py
+++ b/gadopt/gplates/gplates.py
@@ -4,6 +4,7 @@ import numpy as np
 from firedrake.ufl_expr import extract_unique_domain
 from pyadjoint.tape import annotate_tape
 from scipy.spatial import cKDTree
+from ..utility import log
 
 import pygplates
 
@@ -26,6 +27,9 @@ class GPlatesFunctionalityMixin:
             ndtime (float): The model time for which to update the plate
                 velocities. This should be non-dimensionalised time.
         """
+
+        # Print ndtime translated to geological age
+        log(f"pyGplates: Updating surface velocities for {self.gplates_connector.ndtime2age(ndtime):.2f} Ma.")
         # Check if we need to update plate velocities at all
         if self.gplates_connector.reconstruction_age is not None:
             if abs(self.gplates_connector.ndtime2age(ndtime) - self.gplates_connector.reconstruction_age) < self.gplates_connector.delta_t:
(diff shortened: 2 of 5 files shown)


gadopt/gplates/gplatefiles.py
drhodrid, line 1: Accidental commit?

demos/gplates_global/gplates_global.py
```diff
@@ -141,7 +141,7 @@
 # directory. Below, we verify the required paths in this directory and
 # ensure they exist:
 
-muller_2022_files = obtain_Muller_2022_SE(".")
+muller_2022_files = obtain_Muller_2022_SE_v1_2(".")
```
drhodrid, line 144: Given a few lines above, we tell the user to:
# Download and unzip this file into the `./gplates_files`
Should "." here be replaced by "./gplates_files" for consistency?

drhodrid (commented):
Hi @sghelichkhani  - this looks good. I made 2 minor comments for you to address when you get a chance. 

Also just to confirm with @angus-g - is this demo tested? V_1_2 is what we used to use, so results should change with this pull request, but I wanted to be sure things are tested moving forward.

angus-g:
> Also just to confirm with @angus-g - is this demo tested? V_1_2 is what we used to use, so results shouldn't change with this pull request, but I wanted to be sure things are tested moving forward.

Yes, it is. I've refactored this entire change to be a bit more general now. We aren't testing the v1.2.4 reconstruction anywhere though.

drhodrid:
No issue that we're not testing 1.2.4 -- the format of the files at the earthbyte end is the same, so provided that 1.2 is tested, that's all good.

I like your refactoring - thanks for doing that.

Finally - Sia is going to add one line on log output (logging where we are in geological time!), then this should be good to go.

gadopt/gplates/gplatesfiles.py
```diff
@@ -1,33 +1,47 @@
...
-        ]
-    }
-
-    # This is the path to download and extract files
-    base_path = Path(base_path)
-    base_path = base_path / "Muller_etal_2022_SE_1Ga_Opt_PlateMotionModel_v1.2"
-
-    # Check if all files are already present
-    all_files_present = all((base_path / filename).exists() for filenames in plate_reconstruction_files.values() for filename in filenames)
+_default_plate_files = {
```
sghelichkhani, line 3: Since this only works for Muller et al, I suggest changing the name for this:
```suggestion
_default_muller2018_plate_files = {

**Review by sghelichkhani — COMMENTED** [review]
@angus-g Thanks Angus for the neat solution. I'll add a few things and this should be good to go.

File: gadopt/gplates/gplates.py [hunk 3]
```diff
@@ -404,7 +408,7 @@ def _calc_velocities(self, velocity_domain_features, topology_features, rotation
                         # add it to the list
                         all_velocities.extend(velocity_vectors)
                     else:
-                        warnings.warn("Issues finding plate ids for some of the seeds.", category=RuntimeWarning)
+                        warnings.warn("pygplates couldn't assign plate IDs to some seeds due to irregularities in the reconstruction model. G-ADOPT will interpolate the nearest values.", category=RuntimeWarning)
```
sghelichkhani, line 411: This is an attempt to address the issue raised in https://github.com/g-adopt/g-adopt/issues/105.

The issue is solely sue to some issues in the specific plate reconstruction. There is nothing we could do about it at this stage, but at the same time it is important to give a warning about this as we then need to interpolate for those unknown points. 

I have now tried to be more clear about the issue in the warning message hope that helps.

gadopt/gplates/gplatesfiles.py
```diff
@@ -1,33 +1,63 @@
...
+        "directory": "Muller_etal_2022_SE_1Ga_Opt_PlateMotionModel_v1.2",
+    },
+    "Muller 2022 SE v1.2.4": {
+        "plate_files": _default_muller2018_plate_files,
+        "directory": "Muller_etal_2022_SE_1Ga_Opt_PlateMotionModel_v1.2.4",
+    },
+    # Cao et al 2024 extends Muller 2018 to 1.8 Byrs.
+    # DOI: 10.1016/j.gsf.2024.101922
+    # Zenodo: 11536687
+    "Cao 2024": {
```
sghelichkhani, line 27: @drhodrid This is now the newest model available.

gadopt/gplates/gplates.py
```diff
@@ -26,6 +27,9 @@ def update_plate_reconstruction(self, ndtime):
             ndtime (float): The model time for which to update the plate
                 velocities. This should be non-dimensionalised time.
         """
+
+        # Print ndtime translated to geological age
+        log(f"pyGplates: Updating surface velocities for {self.gplates_connector.ndtime2age(ndtime):.2f} Ma.")
```
sghelichkhani, line 32: @drhodrid Is this clear enough?

sghelichkhani:
Thanks to @angus-g again! Once @drhodrid has a look, this is good to go.

gadopt/gplates/gplates.py
drhodrid: Perfect.

drhodrid:
Thanks @sghelichkhani. This looks good to me and I can confirm that the Cao case runs as expected. I'll leave final approval to @angus-g, just to ensure he's happy with the code structure of the most recent commits, but from my perspective this is now good to go.

gadopt/gplates/gplatesfiles.py
```diff
@@ -1,33 +1,63 @@
...
-        ]
-    }
-
-    # This is the path to download and extract files
-    base_path = Path(base_path)
-    base_path = base_path / "Muller_etal_2022_SE_1Ga_Opt_PlateMotionModel_v1.2"
-
-    # Check if all files are already present
-    all_files_present = all((base_path / filename).exists() for filenames in plate_reconstruction_files.values() for filename in filenames)
+_default_muller2018_plate_files = {
```
angus-g, line 3: Why muller2018 when the datasets are Muller 2022?

sghelichkhani: Silly me! Thanks for spotting it.

angus-g (approved):

Issue #111, opened 2024-09-22
TALA_DG test case is incorrectly set up


Patol75:
In the current TALA_DG test case, the reference viscosity profile is given in the [`StokesSolver`](https://github.com/g-adopt/g-adopt/blob/c0d298688b0c1fbf5f4ad92c550dc90d3fa93018/tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py#L208-L210) call, but it is omitted in the [`TruncatedAnelasticLiquidApproximation`](https://github.com/g-adopt/g-adopt/blob/c0d298688b0c1fbf5f4ad92c550dc90d3fa93018/tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py#L125) call. The consequence is an incorrect viscosity value in `viscous_dissipation`. The structure of the test case needs to be updated to account for this change. An example can be found [here](https://github.com/g-adopt/g-adopt/blob/update_approx/tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py).

drhodrid:
Thanks for this Thomas. This was a really important spot, and it turns out it makes quite the difference to some of our global simulations! I will push a change to address this in the coming days.

PR #112, opened 2024-09-23
Ensure approximation is aware of viscosity.

drhodrid:
Address Issue highlighted here: https://github.com/g-adopt/g-adopt/issues/111

Where TALA approximation is not aware of the viscosity. This change necessitates a change to the expected values for the TALA_DG test, which I append below. 

`20 9.179973966724344e-05 4.817959005812114e-06 570.4729595213425 4.0527513455978985 15.18647692671351 11.133725581115613 0.6363801581209881 0.0001688057305039466 1.0011841176113832`

### DIFF
--- a/tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py
+++ b/tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py
@@ -29,6 +29,7 @@ rmin, rmax, ncells, nlayers = 1.208, 2.208, 512, 128
 # In this example, we load the mesh from a checkpoint, although the following code was used to generate
 # the original mesh. It is included here for completeness.
 
+
 def original_mesh():
     def gaussian(center, c, a):
         return a*np.exp(-(np.linspace(rmin, rmax, nlayers)-center)**2/(2*c**2))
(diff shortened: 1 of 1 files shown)


angus-g:
Hmm, looks like your numbers are reasonably different to what's coming out of the CI machine:

| &nbsp; | `u_rms` | `nu_top` | `avg_t` | `FullT_min` | `FullT_max` |
| --- | --- | --- | --- | --- | --- |
| actual | 570.5405141652053 | 15.186137483046036 | 0.6363851935782374 | 0.0001693939995075 | 1.0011846065377914 |
| expected | 570.4729595213425 | 15.18647692671351 | 0.6363801581209881 | 0.0001688057305039 |1.0011841176113832 |
| relative error | **1.18404639e-04** | **2.23517060e-05** | 7.91259335e-06 | **3.47278537e-03** | 4.88347908e-07 |

Noting that the relative error tolerance is 1e-5 by default, and we only override that to 1e-4 in the `3d_cartesian` case.

Patol75:
I ran that case locally and got the same results as the CI:
`20 9.18402977404832e-05 4.8188572591709454e-06 570.5405141653384 4.054356175429129 15.186137483048855 11.131781307619725 0.6363851935782382 0.00016939399951321682 1.0011846065377907`

Patol75 (changes requested):
`expected.pkl` needs to be updated.

Patol75:
What do you mean by "on the CI"? Would not this serial versus parallel inconsistency happen on all machines?

angus-g:
I think it's a parallel-parallel inconsistency, which we addressed for the compressible demos in #95. Sometimes it replicates the serial results, and other times it doesn't!

PR #113, opened 2024-09-23
Backend changes: Equations and terms

Patol75:
This first set of changes extracted from gh-102 proposes to make `Equation` a dataclass that holds the entire state for a given equation and implements terms as functions. These changes are motivated by the fact that the current class implementation of terms is unnecessary and creates several namespaces that obscure the origin of objects used in solvers. With the proposed changes, there is only one namespace per equation (`Equation`), and terms accept an `Equation` instance as their first argument to draw from that single namespace. Note that terms also require a second argument: the trial function object. This choice results from the fact that time steppers must evaluate trial functions at different times, and so having the trial object inside the namespace is troublesome.

**CI status**
Apart from a small mistake that prevented TALA and ALA cases from running, CI results are similar to those of gh-102: `2d_cylindrical` and `3d_spherical` report failures as tolerances are slightly above 1e-5.


gadopt/scalar_equation.py
```diff
@@ -18,69 +18,50 @@
...
+    if hasattr(eq, "su_nubar"):  # SU(PG) à la Donea & Huerta (2003)
+        phi = eq.test + eq.su_nubar / (dot(u, u) + 1e-12) * dot(u, grad(eq.test))
+
+        # The advection term is not integrated by parts so there are no boundary terms
+        F = phi * dot(u, grad(trial)) * eq.dx
+    else:
+        F = -trial * div(eq.test * u) * eq.dx
+        F += trial * dot(eq.n, u) * eq.test * eq.ds  # Boundary term in the weak form
+
+        if not all([is_continuous(eq.trial_space), normal_is_continuous(eq.u)]):
```
angus-g, line 47: I prefer the original

```suggestion
        if not (is_continuous(eq.trial_space) and normal_is_continuous(eq.u)):
```

gadopt/scalar_equation.py
```diff
@@ -93,165 +74,92 @@ class ScalarDiffusionTerm(BaseTerm):
...
+            F -= eq.test * bc["flux"] * eq.ds(bc_id)
+
+    return -F
+
+
+def scalar_source_term(
+    eq: Equation, trial: Argument | ufl.indexed.Indexed | Function
+) -> Form:
     r"""Scalar source term `s_T`."""
+    source = eq.source if hasattr(eq, "source") else 0
```
angus-g, line 135: I think the previous early return behaviour is preferable to always computing a dot product with 0.

gadopt/scalar_equation.py
```diff
@@ -93,165 +74,92 @@ class ScalarDiffusionTerm(BaseTerm):
...
-        phi = test
-        source = fields['source']
 
-        # NOTE, here source term F is already on the RHS
-        F = dot(phi, source)*self.dx
+def scalar_absorption_term(
+    eq: Equation, trial: Argument | ufl.indexed.Indexed | Function
+) -> Form:
+    r"""Scalar absorption term `\alpha_T T`."""
+    sink = eq.sink if hasattr(eq, "sink") else 0
```
angus-g, line 145: Ditto above re: early return.

gadopt/energy_solver.py
```diff
@@ -145,16 +121,70 @@ def __init__(
 
         self.timestepper = timestepper
         self.T_old = Function(self.Q)
+
+        rho_cp = approximation.rhocp()
+        eq_terms = [
+            scal_eq.scalar_advection_term,
+            scal_eq.scalar_diffusion_term,
+            scal_eq.scalar_absorption_term,
+            scal_eq.scalar_source_term,
```
angus-g, line 130:
```suggestion
            scalar_eq.advection_term,
            scalar_eq.diffusion_term,
            scalar_eq.absorption_term,
            scalar_eq.source_term,
```

In my opinion it's clearer to use the qualification to clarify it's scalar, rather than the term name.

gadopt/energy_solver.py
```diff
@@ -145,16 +121,70 @@ def __init__(
...
+        ]
+        terms_kwargs = {
+            "advective_velocity_scaling": rho_cp,
+            "diffusivity": approximation.kappa(),
+            "reference_for_diffusion": approximation.Tbar,
+            "sink": approximation.linearized_energy_sink(u),
+            "source": approximation.energy_source(u),
+            "u": u,
```
angus-g, line 138: I like that the equation is fairly generic, but I don't like the discoverability of having to provide the various terms this way. How are they documented? How are they enforced, i.e. in case of a typo?

I know it's not very different to before, but just something that comes to mind when reading this new approach is that everything is stuffed in a fairly opaque "kwargs" bucket.

gadopt/stokes_integrators.py
```diff
@@ -11,7 +11,8 @@
 import firedrake as fd
 
 from .approximations import BaseApproximation, AnelasticLiquidApproximation
-from .momentum_equation import StokesEquations
+from .equations import Equation
+from .momentum_equation import residual_terms_stokes as terms_stokes
```
angus-g, line 15: If you have to rename "residual terms" to just "terms" either:
- it's redundant to include "residual", and it should just be `terms_stokes`;
- don't rename the import here.

gadopt/stokes_integrators.py
```diff
@@ -255,9 +243,25 @@ def __init__(
                     weak_bc[bc_type] = value
             self.weak_bcs[id] = weak_bc
 
+        u, p = fd.split(self.solution)
+        terms_kwargs = {"u": u, "p": p, "T": T, "mu": mu}
```
angus-g, line 247: As with the energy solver, I don't think it's made clear anywhere that this is the required set of functions for the particular set of equations.

gadopt/momentum_equation.py
```diff
@@ -18,217 +18,175 @@
...
+
+    return -F
+
+
+def divergence_term(
+    eq: Equation, trial: Argument | ufl.indexed.Indexed | Function
+) -> Form:
+    assert normal_is_continuous(eq.u)
+
+    rho = eq.approximation.rho
```
angus-g, line 169: This seems to have changed from `rho_continuity`

gadopt/momentum_equation.py
```diff
@@ -18,217 +18,175 @@
...
+    eq: Equation, trial: Argument | ufl.indexed.Indexed | Function
+) -> Form:
+    source = eq.approximation.buoyancy(eq.p, eq.T) * upward_normal(eq.mesh)
+    F = -dot(eq.test, source) * eq.dx
+
+    return -F
+
+
+residual_terms_momentum = [momentum_source_term, pressure_gradient_term, viscosity_term]
+residual_terms_stokes = [residual_terms_momentum, divergence_term]
```
angus-g, line 192: Should this be
```suggestion
residual_terms_stokes = residual_terms_momentum + [divergence_term]
```
to avoid nesting the lists of terms?

angus-g (changes requested):
Thanks Thomas. I think this is indeed a cleaner and more modular approach. I've got a few specific comments here and there about implementation particulars.

My one overall concern is about the `terms_kwargs` mechanism. I guess this is fairly equivalent to what was `fields`, but I'm not sure I like the name. Additionally, I can't see anywhere that documents the required keys for a given term or set of terms. It would be good for discoverability and error-checking to be able to have this.

gadopt/scalar_equation.py
Patol75: @stephankramer @w3168 Any idea if this was intended to be `not x and not y` or `not x and y`? I would tend to say the former, but then it disagrees with what we have in `master`.

gadopt/momentum_equation.py
```diff
@@ -18,217 +18,175 @@
...
+from .utility import is_continuous, normal_is_continuous, tensor_jump, upward_normal
 
-from .equations import BaseTerm, BaseEquation
-from .utility import is_continuous, normal_is_continuous, tensor_jump, cell_edge_integral_ratio
+__all__ = [
+    "divergence_term",
+    "momentum_source_term",
+    "pressure_gradient_term",
+    "viscosity_term",
+]
```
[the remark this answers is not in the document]
Patol75: I do not think we have a use case for it so far. I will remove it.

gadopt/momentum_equation.py
```diff
@@ -18,217 +18,175 @@
...
+            # (u_ext = trial) and stress = n . (mu . stress_tensor).
+            F -= dot(eq.test, bc["stress"]) * eq.ds(bc_id)
+
+        if "normal_stress" in bc:
+            F += dot(eq.test, bc["normal_stress"] * eq.n) * eq.ds(bc_id)
+
+        # if "drag" in bc:  # (bottom) drag of the form tau = -drag trial |trial|
+        #     drag = bc["drag"]
+        #     unorm = pow(dot(trial_lagged, trial_lagged) + 1e-6, 0.5)
+        #     F += dot(eq.test, drag * unorm * trial) * eq.ds(bc_id)
```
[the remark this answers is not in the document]
Patol75: @w3168 Any idea if this is something we need or if it can just be removed for now?

gadopt/momentum_equation.py
Patol75: True, thank you. I realised only recently with Will that this was not appropriate. I corrected it in `update_approx` but forgot to push here.

gadopt/momentum_equation.py
Patol75: Currently, nesting is desired because it separates the momentum and mass terms; I also allow terms to be either a list or a callable. It is true that we could also group all terms without changing the final UFL form. However, given that the mass equation should receive some additional terms in the near future, that may not be optimal.

gadopt/equations.py
```diff
@@ -1,194 +1,162 @@
...
+    """
+    degree = eq.trial_space.ufl_element().degree()
+    if not isinstance(degree, int):
+        degree = max(degree)
+
+    if degree == 0:  # probably only works for orthogonal quads and hexes
+        sigma = 1.0
+    else:
+        # safety factor: 1.0 is theoretical minimum
+        alpha = eq.interior_penalty if hasattr(eq, "interior_penalty") else 2.0
```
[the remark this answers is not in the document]
Patol75: Neat, thank you!

gadopt/scalar_equation.py
w3168: hmm I wonder if the original logic was actually right... 

I remember years ago finding a bug with @stephankramer where we had copied the code from the DG scalar advection demo in Firedrake and this assumed that the velocity was in a continuous function space. 

Here's the demo: https://www.firedrakeproject.org/demos/DG_advection.py.html

and the commit in thwaties where we changed it: https://github.com/thwaitesproject/thwaites/commit/c291d09c959fd506d639cf05cf709cb7ede2ecb0#diff-8b82ede7584525921942c3de6d02bea35e9a0fd6a14652133e7bce49d488a792L43

So I think the code in Gadopt (and thwaites) now is the right thing to do if both the tracer and the velocity are discontinuous. I think should satisfy the existing if statement.  (@Patol75 I don't quite see what you mean about it disagreeing with what is currently in master?)

For context when doing ocean modelling with thwaites we always used to use P1DG for velocity and tracers (and then later a tensor product element that was discontinuous in the normal direction...).

I wonder if the gadopt usecase with P2 for velocity and DG for the tracers is doing the right thing... It probably should use the example firedrake code (though it might end  end up being the same as the modified thwaites code when the (normal) velocity is continuous). Would it actually be turned on by the current (and thomas') if statement though?

gadopt/momentum_equation.py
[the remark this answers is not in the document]
Patol75: I see. I would remove it then. @drhodrid Any objections?

gadopt/scalar_equation.py
Patol75: Quick answer about the confusion: Python evaluates `not x and y` as `(not x) and (y)`, which is different from `not x and not y`. So if `x` and `y` both evaluate to `False`, then `not x and y` evaluates to False, whereas `not x and not y` evaluates to True. Does that change your answer?

Patol75: Neverming, I misread the parentheses in the code...

w3168: I'm probably being stupid but isnt the not outside a bracket... so it evaluates not(x and y) as opposed to not(x) and y which you seem to be saying...

I think the original ocean modelling case: 
x: tracer is continuous - > False
y: velocity is continuous -> False
x and y -> False
so not (x and y) -> True, i.e. add the extra ufl in.



but for mantle convection now... 
x: tracer is continous -> False
y: velocity is continuous -> True
x and y = False
so not(x and y) -> True add the extra ufl in.  
(edit: Ah so I was slightly wrong above. The mantle convection case will also add the ufl in ... with the caveat this might be wrong if u is continuous, but it probably is ok... )

w3168: I guess this logic might catch the case where you have continuous tracers but discontinuous velocity...?

gadopt/energy_solver.py
```diff
@@ -145,16 +121,70 @@ def __init__(
...
+            vel = terms_kwargs["u"]
+            Pe = absv(dot(vel, J)) / (2 * kappa)  # Calculate grid peclet number
+            nubar = su_nubar(vel, J, Pe)  # Calculate SU artifical diffusion
+
+            terms_kwargs["su_nubar"] = nubar
+        self.eq = Equation(
+            TestFunction(self.Q),
+            self.Q,
+            eq_terms,
+            mass_term=lambda eq, trial: scal_eq.mass_term(eq, rho_cp * trial),
```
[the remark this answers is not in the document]
Patol75: I am not too sure what to do here. I am quite against defining multiple `mass_term` functions if they do essentially the same thing up to a factor. I thought a `lambda` would be a reasonable solution, but if not, then I am unsure how to proceed.

gadopt/stokes_integrators.py
Patol75: Makes sense, thank you.

gadopt/energy_solver.py
Patol75: I agree with what you raised here. I tried to provide some logic in the Equation class based on attributes I added to each term function. Let me know what you think of it.

Patol75:
@angus-g Many thanks for the review. I think I have addressed all of your points and merged new changes from `master`. The CI should hopefully return green, and I will be happy to have your opinion on the latest commits.

gadopt/energy_solver.py
angus-g: Definitely an improvement, yeah. Runtime validation is nice, and being able to see all the required attributes in one place at least makes it a bit easier to figure out documentation and discoverability.

gadopt/momentum_equation.py
angus-g: That's fair enough, if it's handled then I agree it gives a bit of semantic meaning to the list of terms.

gadopt/momentum_equation.py
```diff
@@ -47,195 +42,147 @@ class ViscosityTerm(BaseTerm):
...
+        F -= inner(tensor_jump(eq.n, eq.test), avg(stress)) * eq.dS
+
+    # NOTE: Unspecified boundaries are equivalent to free stress (i.e. free in all
+    # directions).
+    # NOTE: "un" can be combined with "stress" provided the stress force is tangential
+    # (e.g. no normal flow with wind)
+    for bc_id, bc in eq.bcs.items():
+        if "u" in bc and any(bc_type in bc for bc_type in ["drag", "stress", "un"]):
+            raise ValueError(
+                '"drag", "stress", or "un" cannot be specified if "u" is already given.'
```
w3168, line 81: I think you've removed the "drag" option now

gadopt/stokes_integrators.py
```diff
@@ -281,9 +267,34 @@ def __init__(
                     weak_bc[bc_type] = value
             self.weak_bcs[id] = weak_bc
 
+        # eta is a list of 0, 1 or multiple free surface fields
+        u, p, *eta = fd.split(self.solution)
+        rho_mass = approximation.rho_continuity()
+        stress = approximation.stress(u)
+        eqs_attrs = [
+            {"u": u, "p": p, "T": T, "stress": stress},
```
w3168, line 275: Does u need to be provided to momentum now as it's the trial function?

gadopt/equations.py
```diff
@@ -1,204 +1,177 @@
...
-
-        self.dx = firedrake.dx(domain=self.mesh, degree=quad_degree)
+            self.ds = fd.ds(**measure_kwargs)
+            self.dS = fd.dS(**measure_kwargs)
 
-        # General prefactor multiplying all terms in the equation
-        # N.b. setting this to a firedrake constant (i.e. prefactor = Constant(1)) breaks
-        # Drucker-Prager rheology test even though it is being multiplied by 1...
-        self.prefactor = prefactor
+    def mass(
```
w3168, line 114: Any reason why you are calling this mass intead of mass_term?

gadopt/momentum_equation.py
```diff
@@ -18,26 +18,21 @@
...
-
 """
 
-from typing import Optional
-
-import firedrake as fd
-from firedrake import dot, inner, outer, transpose, nabla_grad, div
-from firedrake import avg, Identity, jump
-from firedrake import FacetArea, CellVolume
+from firedrake import *
```
w3168, line 23: I know this has come up in meetings before... How come you are now importing all of firedrake?

gadopt/scalar_equation.py
```diff
@@ -18,69 +18,48 @@
...
-the solver provided in the stokes_integrators module.
-
 """
 
-from typing import Optional
-
-import firedrake as fd
-from firedrake import avg, div, dot, grad, inner, jump, min_value, sign
-from firedrake import CellVolume, FacetArea, Identity
+from firedrake import *
```
w3168, line 23: similar here why not keep the old way?

gadopt/free_surface_equation.py
```diff
@@ -1,48 +1,55 @@
...
+however, we like to think about the terms as they are on the LHS. Therefore, in the
+function below, we assemble in F as it would be on the LHS:
 
   dq/dt + F(q) = 0
 
-and at the very end "return -F".
+and at the very end return "-F".
 """
 
+import firedrake as fd
```
w3168, line 16: again slightly different to what was before and in momentum/scalar...

gadopt/energy_solver.py
```diff
@@ -7,12 +7,21 @@
 from numbers import Number
 from typing import Any, Optional
 
-from firedrake import Constant, DirichletBC, Function, Jacobian, TensorFunctionSpace, dot
-
+from firedrake import (
```
w3168, line 10: firedrake imports is not consistent across the different modules.... Probably not a big issue but probably better be consistent

gadopt/time_stepper.py
```diff
@@ -272,8 +267,7 @@ def __init__(
                         u = self.solution_old + self.a[i][j]*self.dt_const*self.k[j]
                     else:
                         u += self.a[i][j]*self.dt_const*self.k[j]
-                self.F.append(self.equation.mass_term(self.test, self.k[i]) -
-                              self.equation.residual(self.test, u, self.solution_old, fields, bnd_conditions))
+                self.F.append(self.equation.mass(self.k[i]) - self.equation.residual(u))
```
w3168, line 270: as above mass -> mass_term?

gadopt/momentum_equation.py
```diff
@@ -47,195 +42,147 @@ class ViscosityTerm(BaseTerm):
...
+def momentum_source_term(
+    eq: Equation, trial: Argument | ufl.indexed.Indexed | Function
+) -> Form:
+    source = eq.approximation.buoyancy(eq.p, eq.T) * upward_normal(eq.mesh)
+    F = -dot(eq.test, source) * eq.dx
+
+    return -F
+
+
+viscosity_term.required_attrs = {"u", "stress"}
```
w3168, line 177: as mentioned in stokes integrators, is 'u' a required attribute for viscosity given its the trial function?

w3168 (commented):
Looks really nice @Patol75 I think it is cleaner than before, especially removing the disctinction between term_kwargs and fields which always seemed a bit blurred. 

I've highlighted a few name queries etc but otherwise looks good to me.

gadopt/scalar_equation.py
```diff
@@ -93,165 +72,96 @@ class ScalarDiffusionTerm(BaseTerm):
...
-        # NOTE, here source term F is already on the RHS
-        F = dot(phi, source)*self.dx
 
-        return F
+def absorption_term(
+    eq: Equation, trial: Argument | ufl.indexed.Indexed | Function
+) -> Form:
+    r"""Scalar absorption term `\alpha_T T`."""
+    # Implement absorption term implicitly at current time step.
+    F = dot(eq.test, eq.sink * trial) * eq.dx
```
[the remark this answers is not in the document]
Patol75: Sounds fair to me. What about `sink_coeff`? Or would you prefer `absorption_coeff`?

gadopt/momentum_equation.py
Patol75: Good catch, thank you.

gadopt/momentum_equation.py
Patol75: Good catch, thank you.

gadopt/stokes_integrators.py
Patol75: Good catch, thank you.

gadopt/equations.py
Patol75: I think I wanted to reserve `_term` to functions defined in modules such as `momentum_equation.py` or `scalar_equation.py`. You can see below that I also just have `residual`. Is that an issue?

w3168: ok, but isn't it still a mass_term. the reason you can put it in equations.py is because it is pretty general to any time dependent pde...

gadopt/scalar_equation.py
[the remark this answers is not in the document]
w3168: actually on second thoughts it's probably a bad idea to change the user interface. You've managed so far to keep everything the same from the demo perspective which is a good thing. I guess if it was raised at something like the gadopt workshop or multiple mention that it is weird we should change it in a separate PR

gadopt/momentum_equation.py
Patol75: Putting this link [here](https://stackoverflow.com/questions/710551/use-import-module-or-from-module-import), as it can be useful.
I decided to use `from firedrake import *` because we now have a lot of objects in this module that come from `firedrake` and other developers have argued against `import firedrake as fd` in the past. Listing all the objects imported seems tedious to me and results in a never-ending import statement. From a Python perspective, I still think the right thing to do is `import firedrake as fd`, but I also see why it is not popular among developers. @angus-g @dsroberts Any opinions here?

gadopt/scalar_equation.py
Patol75: See comment in `momentum_equation.py`.

gadopt/free_surface_equation.py
Patol75: Given the comment in `momentum_equation.py`, this module can serve as an example documenting what the source code looks like when `import firedrake as fd` is used.

gadopt/energy_solver.py
Patol75: Consistency would be nice, yes. See other related comments.

gadopt/time_stepper.py
Patol75: I like the symmetry between `mass` and `residual` here. Is `mass` incorrect?

gadopt/scalar_equation.py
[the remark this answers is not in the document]
Patol75: I am not sure I follow. In `master`, this term is currently called `absorption_coefficient` (see [here](https://github.com/g-adopt/g-adopt/blob/77d9d524f2547b0d167cc17dbbff804d2e06c58e/gadopt/energy_solver.py#L93) and [here](https://github.com/g-adopt/g-adopt/blob/77d9d524f2547b0d167cc17dbbff804d2e06c58e/gadopt/scalar_equation.py#L207)).

gadopt/equations.py
Patol75: I do not understand what `_term` brings here. Surely no one will confuse this with the physical mass. From the finite elements perspective, what we are implementing here is a mass matrix. Given that we define terms in other modules, `mass` seems appropriate to me here.

gadopt/scalar_equation.py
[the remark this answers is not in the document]
w3168: I just mean that the keyword used in approximation is 'sink' so it proably makes most sense to be consistent with that... otherwise sink_coeff is good

gadopt/momentum_equation.py
w3168: thanks for the link

angus-g: I think `from firedrake import *` is better than the list of explicit imports. Yes, it would be nice to be consistent across all the g-adopt modules, but that's not in the scope of this PR. Firedrake/UFL is _almost_ its own DSL, so I don't think the usual rules apply anyway.

gadopt/scalar_equation.py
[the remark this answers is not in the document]
Patol75: Well, we have `linearized_energy_sink` in `approximations.py`, but we have `absorption_term` in `scalar_equation`. Should that be homogenised too?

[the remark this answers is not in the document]
w3168: I guess sink_term is easier to write than absorption_term!

Patol75:
Thanks for the review, @w3168. It is mostly addressed now, except if we want some further name or import changes.

gadopt/momentum_equation.py
Patol75: For those who do not know what a DSL is: https://www.jetbrains.com/mps/concepts/domain-specific-languages/

angus-g:
Great! I think the only remaining concern here is the interaction with #122. I imagine there's not a huge overlap, so it's probably safe to get this in first?

Patol75:
It will probably affect some of the solvers in gh-122, but I suspect it will not be a deal-breaker. I have already ported gh-117 to `update_approx` and plan to look at porting gh-122 later today.

w3168:
Nice! @angus-g  I don't mind re #122. In case it changes anything I've tidied up the viscoelastic demo more now so I'm ready for a review on that. It should hopefully be fairly easy to copy over some of @Patol75 changes from his #102 branch to here to get it to work with the new equations setup...

angus-g (approved):

PR #114, opened 2024-09-23
Backend changes: GenericTransportSolver

Patol75:
This second set of changes extracted from gh-102 proposes to reformat the `energy_solver` module to expose both `AdvectionDiffusionSolver` and `EnergySolver` classes. These changes do not impact the use of `EnergySolver`. The motivation here is the fact that some of our demos and tests set up scenarios that only require an advection or advection-diffusion equation. Currently, we use `BoussinesqApproximation` and `EnergySolver` in these scripts, but this is not straightforward. In other words, we make it unnecessarily complicated for a user to solve such problems using G-ADOPT. The proposed changes solve this issue and allow additionally to solve a pure diffusion problem.  

Needs gh-113.


gadopt/utility.py
```diff
@@ -191,31 +191,6 @@ def cell_size(mesh):
         return CellDiameter(mesh)
 
 
-def cell_edge_integral_ratio(mesh, p):
```
drhodrid, line 194: Is there a reason for deleting this? I think Steph has a branch working on limiters that may use this, but there's a chance I'm wrong.

Patol75: I moved it to the `equation` module because it was only used for the interior penalty method found in some specific terms. Might not be the best idea, but I felt like it did not belong in `utility`. By the way, `utility` is quite large now, and perhaps it should be broken up into multiple modules.

demos/scalar_advection/scalar_advection.py
```diff
@@ -1,6 +1,6 @@
 # Demo for pure scalar advection - this is adapted from the Firedrake DG_advection demo,
```
Patol75, line 1: @w3168 pointed out that this case should not be in demos. What do we do with it?

drhodrid:
It should be in tests - I thought it was already, but I’m clearly tired!!

On 21 Oct 2024, at 1:42 pm, Thomas Duvernay ***@***.***> wrote:



@Patol75 commented on this pull request.

________________________________

In demos/scalar_advection/scalar_advection.py<https://github.com/g-adopt/g-adopt/pull/114#discussion_r1808033407>:

> @@ -1,6 +1,6 @@
 # Demo for pure scalar advection - this is adapted from the Firedrake DG_advection demo,


@w3168<https://github.com/w3168> pointed out that this case should not be in demos. What do we do with it?

angus-g:
> It should be in tests

We're not even running/testing it... :fearful:

gadopt/energy_solver.py
```diff
@@ -53,82 +50,317 @@
...
-        bcs = bcs or {}
-        for id, bc in bcs.items():
+
+        apply_strongly = is_continuous(self.solution)
+
+        for bc_id, bc in bcs.items():
             weak_bc = {}
             for type, value in bc.items():
-                if type == 'T':
+                if type == "T":
```
w3168, line 131: Just a small suggestion but probably strong dirichlet bcs in the generic transport solver should be something like q_strong? or Q? (but that is sometimes used for function spaces...) 

And then energy solver can accept T but it gets passed to generic transport solver...

gadopt/energy_solver.py
```diff
@@ -53,82 +50,317 @@
...
+    All combinations of advection, diffusion, sink, and source terms are valid.
+
+    The scalar field is updated in place.
+
+    Arguments:
+      terms:
+        List of equation terms defined in scalar_equation.py
+      scalar_field:
+        Firedrake function for scalar field of interest
+      u:
```
w3168, line 224: Can we make velocity an optional argument of Generic transport solver? I'm thinking of the internal variable timestepping which is just the source and the sink term. 

The energy solver on top would still have it as a required argument but then passes to the Generic transport solver...

w3168:
> > It should be in tests
> 
> We're not even running/testing it... 😨

There's actually two more scripts in the scalar_advection_diffusion folder that never get run... 

https://github.com/g-adopt/g-adopt/tree/master/tests/scalar_advection_diffusion

gadopt/energy_solver.py
Patol75: I think this is possible through a class variable, e.g. `self.strong_bcs_tag = "T"`. Then, you would have `if bc_type == self.strong_bcs_tag:`. Is that desirable though?

gadopt/energy_solver.py
Patol75: Yes, sure, that makes sense.

Patol75:
This PR is now ready for review. I have merged the latest `master` changes and revamped the `energy_solver` module given recent developments of the `stokes_integrator` module in `update_approx`. I am out of time for today, so I only checked that `base_case` and `scalar_advection_diffusion_DH27` pass locally. I will make the minor solver call changes for all other demos and test tomorrow.

gadopt/energy_solver.py
```diff
@@ -1,34 +1,28 @@
-r"""This module provides a fine-tuned solver class for the energy conservation equation.
-Users instantiate the `EnergySolver` class by providing relevant parameters and call
-the `solve` method to request a solver update.
+r"""This module provides a minimal solver for generic transport equations, which may
```
drhodrid, line 1: In light of the change here to energy solver inheriting from a generic setup, should the file name be changed from energy_solver to something different?

gadopt/energy_solver.py
```diff
@@ -38,12 +32,16 @@
...
+"""Default iterative solver parameters for solution of energy equation.
+
+Configured to use the GMRES Krylov scheme with Successive Over Relaxation (SOR)
+preconditioning. Note that default energy solver parameters can be augmented or adjusted
+by accessing the solver_parameter dictionary, for example:
+energy_solver.solver_parameters['ksp_converged_reason'] = None
+energy_solver.solver_parameters['ksp_rtol'] = 1e-4
+
+Note:
+  G-ADOPT defaults to iterative solvers in 3-D.
```
drhodrid, line 44: Does it also default to iterative solvers in 2-D cylinders? I know this comment it old, but perhaps worth updating to be correct.

gadopt/energy_solver.py
```diff
@@ -53,143 +51,335 @@
...
     "pc_factor_mat_solver_type": "mumps",
 }
-"""Default direct solver parameters for solution of energy equation. Configured to use LU factorisation,
-   using the MUMPS library. G-ADOPT defaults to direct solvers in 2-D.
+"""Default direct solver parameters for solution of energy equation.
+
+Configured to use LU factorisation performed via the MUMPS library.
+
+Note:
+  G-ADOPT defaults to direct solvers in 2-D.
```
drhodrid, line 59: See previous comment.

tests/scalar_advection/scalar_advection.py
```diff
@@ -1,6 +1,6 @@
 # Demo for pure scalar advection - this is adapted from the Firedrake DG_advection demo,
-# but here we use G-ADOPT's Energy solver and use a CG discretisation
-# with Streamline Upwind (SU) stabilisation.
+# but here we use G-ADOPT's GenericTransportSolver and use a CG discretisation with
```
drhodrid, line 2: This maybe semantics on my part, but we are still solving the energy equation here, right, albeit with SU stabilisation? Why then are we switching to Generic Transport solver? Or is that just to ensure that it is tested? If the latter, could we also ensure that the energy solver is tested (and configurable) with SU?

tests/scalar_advection_diffusion/scalar_advection_diffusion.py
```diff
@@ -58,27 +58,28 @@
 dt = T/600.0
 
 
-# Use G-ADOPT's Energy Solver to advect the tracer. By setting the Rayleigh number to 1
-# the choice of units is up to the user. We use the diagonaly implicit DIRK33 Runge-Kutta
-# method for timestepping. 'T' means that the boundary conditions will be applied strongly
-# by the energy solver.
-approximation = BoussinesqApproximation(Ra=1, kappa=kappa)
+# Use G-ADOPT's GenericTransportSolver to advect the tracer. We use the diagonally
```
drhodrid, line 61: See previous comment.

drhodrid (commented):
Some small things from me. Generally like this, but would appreciated @angus-g and/or @dsroberts having a careful look at the code specifics - and @w3168 always picks up something we all miss too!!

gadopt/energy_solver.py
Patol75: I have been considering changing file names, but I stopped myself from doing so. If developers are for it, then why not `transport_solvers.py`?

gadopt/energy_solver.py
Patol75: There is a difference between the transport and Stokes solvers: the former do not consider the model's geometry when choosing their solver strategy. This choice is likely a consequence of the easier energy solves compared to Stokes's. Should this behaviour change?

drhodrid: I think it is better if all solvers transition to iterative under the same conditions. Let me think about this and perhaps it can be a future pull request, but leave your comments as is for now then.

tests/scalar_advection_diffusion/scalar_advection_diffusion.py
drhodrid: Sorry you are right. I was getting confused with different test cases. Since we still have users (earthbyte) using SU with our energy solver, I just want to make sure that this aspect is also tested.

Patol75: This time, it is an advection-diffusion problem, but again, it does not have much to do with the usual energy conservation we usually refer to. Moreover, there is no reason to involve an approximation at all. A simple, generic solver can perfectly handle the situation.

dsroberts:
All I really have to add is some nitpicky python and stylistic stuff. That being said, I did see an interesting discussion about the use of native python types in UFL expressions causing poor performance on the Firedrake slack, I'll add a comment where I think this might affect your modifications. Feel free to tell me I'm wrong, I don't know this code well.

Patol75:
@dsroberts Do you mean something in gh-113 can cause performance degradation? Also, feel free to put some style and Python-specific comments; I will consider them. :)

demos/2d_compressible_ALA/2d_compressible_ALA.py
```diff
@@ -136,7 +136,9 @@
 # pass in different nullspace and transpose_nullspace objects to the stokes solver.
 
 # +
-energy_solver = EnergySolver(T, u, approximation, delta_t, ImplicitMidpoint, bcs=temp_bcs)
+energy_solver = EnergySolver(
```
w3168, line 139: Is there a reason you changed the order from T,u, approximation -> approximation, T, u...? 

The PR would be neater if this would be the same as before and it wouldnt break existing code.

demos/2d_compressible_ALA/2d_compressible_ALA.py
```diff
@@ -168,7 +170,7 @@
     energy_conservation_2 = abs(rate_work_against_gravity - rate_viscous_dissipation)
 
     # Calculate L2-norm of change in temperature:
-    maxchange = sqrt(assemble((T - energy_solver.T_old)**2 * dx))
+    maxchange = sqrt(assemble((T - energy_solver.solution_old) ** 2 * dx))
```
w3168, line 173: I think it would be neater if energy solver has an attribute T_old  = solution_old. so the demos/existing user code doesnt have to change.

gadopt/energy_solver.py
```diff
@@ -53,143 +51,335 @@
...
 """
 
 
-class EnergySolver:
-    """Timestepper and solver for the energy equation. The temperature, T, is updated in place.
+class MetaPostInit(abc.ABCMeta):
```
w3168, line 63: I am happy to learn more about this but I still don't see why it's really necessary... 

A few commits ago you didnt have them in and it seemed a fairly simple PR change...

I was reading here towards the end 
_'''Now the big question. Why would you use some obscure error-prone feature?

Well, usually you don't:

    Metaclasses are deeper magic that 99% of users should never worry about it. If you wonder whether you need them, you don't (the people who actually need them know with certainty that they need them, and don't need an explanation about why).

Python Guru Tim Peters'''_
https://stackoverflow.com/questions/100003/what-are-metaclasses-in-python

gadopt/energy_solver.py
```diff
@@ -53,143 +51,335 @@
...
+        eq_attrs = {
+            "advective_velocity_scaling": rho_cp,
+            "diffusivity": self.approximation.kappa(),
+            "reference_for_diffusion": self.approximation.Tbar,
+            "sink_coeff": self.approximation.linearized_energy_sink(self.u),
+            "source": self.approximation.energy_source(self.u),
+            "u": self.u,
+        }
+
+        if self.su_diffusivity is not None:
```
w3168, line 372: Do you need to repeat this code?

w3168 (commented):
Just a few more thoughts

demos/2d_compressible_ALA/2d_compressible_ALA.py
Patol75: I think that it is a discussion I wanted to raise. So happy to have devs' thoughts on it. I do not mind reverting changes. Also, I doubt any of my PRs will ever be neat.

demos/2d_compressible_ALA/2d_compressible_ALA.py
Patol75: Happy for devs to discuss that too.

demos/2d_compressible_ALA/2d_compressible_ALA.py
angus-g: I agree that the parameter order should probably be kept the same.

demos/2d_compressible_ALA/2d_compressible_ALA.py
angus-g: I think this would be a nice thing to keep for compatibility too. It doesn't cost anything to have the same underlying function with two different names.

gadopt/energy_solver.py
Patol75: I had a go, and I think I can get away without repeating. Thank you.

.github/workflows/test.yml
```diff
@@ -74,6 +74,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           . /home/firedrake/firedrake/bin/activate
+          python3 -m pip install siphash24
```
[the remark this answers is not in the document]
Patol75: Cool, I will remove it.

demos/2d_compressible_ALA/2d_compressible_ALA.py
Patol75: The council has spoken. Reverted.

demos/2d_compressible_ALA/2d_compressible_ALA.py
Patol75: Added in. Thank you.

gadopt/energy_solver.py
Patol75: I do not think I am wondering whether I need the metaclass here. :)

In `master`, we have a lengthy `__init__` method. This is not optimal when trying to implement a class inheritance logic. So, I started by creating some methods, such as `set_boundary_conditions`, to tidy the `__init__` method. I wanted these methods to lie in an abstract class from which to define subclasses. I require the class to be abstract because subclasses must implement their own `set_equation` method; thus, it makes sense to have the latter as an `abstractmethod`. With the inheritance logic, I realised that the newly implemented methods are quite general and should not require any method overwriting. However, that same inheritance logic implies that it is unlikely one bypasses the overwriting of `__init__` when implementing a new solver. Now the problem starts to shape. The abstract base class `__init__` eventually calls all the methods I defined. You could move these calls outside of `__init__`, but I do not think we want a user to have to call another method after calling the solver's constructor. Given the nature of the inheritance pattern, I want the subclass to call its parent's `__init__` method using `super` but also bring in additional functionalities. That is, I believe, where a problem can occur. What if new functionalities must be defined before the methods are called but after the remainder of the parent's `__init__` code? In other words, what if, whether you called `super` first or last, you still had a problem? The solution I propose using a metaclass addresses this issue. It separates the methods from the rest of the `__init__` code in the parent class and always makes them execute after the subclass's `__init__`. When subclassing the abstract base class, a developer can clearly separate specific initialisation code from generic method calls.

To be honest, this issue arose when I was working on the `stokes_integrators` module in `update_approx` following changes from the viscoelastic branches. However, I found the solution quite neat, so I thought it would be desirable to also have it in the `energy_solver` module.

I do not see why this approach should be considered error-prone. I see it as a feature to ease further solver development.

Patol75:
Thank you everyone for the reviews. I believe I have addressed all comments either through code changes or discussions. CI is currently running and should return green (except if I missed something I guess). There are still open items for discussion, so happy to have further input, if any.

gadopt/energy_solver.py
dsroberts: I agree with @w3168. Please think about the long-term maintainability of any changes you're going to commit. If this were your own project, it would be absolutely fine to write code that pushes the boundaries of your knowledge in an attempt to understand the language better. This is a community project however, and its going to be worked on by people with a range of scientific and programming abilities. Not everyone who wants to develop on G-ADOPT for their own research will have your ability or your motivation to explore Python. 'Get good' isn't the right answer to that, as that kind of attitude could very much hurt adoption of G-ADOPT in the long term. After all, G-ADOPT is a thin wrapper to Firedrake, if building on it is too hard, I'm not going to bother and just write my own Firedrake wrapper. This is why larger projects have standards documents that spell out what language features can and can't be used. That kind of thing is way overkill for G-ADOPT, but I hope you see my point. Apologies if I'm stepping out of my lane here, I've had this battle a few times.

On to the actual code, I've not come across metaclasses before, but what you're doing here seems to attempt to emulate the init behaviour of the `dataclass` decorator. Could you just use `@dataclass` and guarantee the order of `__init__` calls and the other methods that way? This looks like a more appropriate use of that decorator than in e.g. `equations.py` where half the stuff set by the default `__init__` method is overwritten in `__post_init__`.

Patol75:
```quote
Data classes are just regular classes that are geared towards storing state, rather than containing a lot of logic. Every time you create a class that mostly consists of attributes, you make a data class.
```
On the `dataclass` aspect, I try to follow this statement from [Martijn Pieters](https://stackoverflow.com/a/47955313/10640534). From that perspective, `Equation` seems like an obvious candidate to me but not any G-ADOPT solver. I would add that nothing is overwritten in `Equation`'s `__post_init__`: `InitVar` is part of the language specification, and I use it to replicate the behaviour that was previously in `master`. It makes it clear which arguments are stored as-is and which ones imply some post-processing.

Regarding maintainability, I fail to see the issue. Indeed, metaclasses are not your common Python object, but it is not as if the one I implemented is achieving anything mysterious or unreasonable. The docstring is clear and there is only one added call. As I wrote elsewhere, this change is geared towards making the implementation of solvers easier. Frankly, look at the `EnergySolver` code. Would you argue there is anything complicated or unreasonable in it? From my perspective, it is extremely straightforward, such that anyone who would attempt to add a different solver could easily be inspired by that one and achieve what they want.

dsroberts: OK, you've convinced me that `@dataclass` isn't the right construct for this. But lets look at how `@dataclass` implements the `__post_init__` call then. Surely the implementation in the language itself will be the preferred option? [First, it tests if `__post_init__` exists](https://github.com/python/cpython/blob/34653bba644aa5481613f398153757d7357e39ea/Lib/dataclasses.py#L1076), then follows on to `_init_fn` which is dynamically building an `__init__` method for the class in a `_FuncBuilder` instance, [notably adding the `__post_init__` call last](https://github.com/python/cpython/blob/34653bba644aa5481613f398153757d7357e39ea/Lib/dataclasses.py#L649). Once that's done it dynamically creates `__repr__`, `__eq__` and ordering methods, then calls `add_fns_to_class`. Finally, [the whole thing is run with `exec`](https://github.com/python/cpython/blob/34653bba644aa5481613f398153757d7357e39ea/Lib/dataclasses.py#L502).

Obviously I would not advocate for this entire workflow in this case. `@dataclass` needs far more flexibility than we do, and we all know about `exec`. What I note, however, is that there are no metaclasses defined here. It simply creates an `__init__` that runs `__post_init__` last. If Python itself does not use a metaclass to implement this behaviour, we shouldn't either.

Patol75: Very interesting, thank you. How would that work with the subclassing idea though? My whole point is to separate `__post_init__` from `__init__` such that `__post_init__` would always execute after all `__init__` have run (involving `super` in the process).

dsroberts: Custom decorator? https://gist.github.com/dsroberts/cbd9ed5b9ac607bf4ddbab171102fdf1

dsroberts: You could make the whole thing a little more user friendly by dropping `__post_init__` and naming the function after what its actually doing. So rename `__post_init__` to `solver_initialise` or something and call the decorator `@solver`. Documentation can explain why we have a decorator on those classes.

w3168: Thanks @Patol75 and @dsroberts I'm learning a lot about python and classes! 

_What if new functionalities must be defined before the methods are called but after the remainder of the parent's __init__ code?_

Do we have this scenario? Maybe it is more complicated in the update_approx branch with the stokes integrators, but cant you just define the extra attributes for the later clasess before you call the super().__init__ method? And then you could put the set_equations method etc just inside the init method of the abstract base class.

Patol75: Fair (provided it is compatible with an abstract base class; I have not checked). However, how is it simpler, more user-friendly, and more maintainable than the metaclass approach? If the user/developer forgets to add the decorator, then the solver will not work.

Patol75: @w3168 Yes, that would work here. However, it would not handle the case where `super` comes first. It makes sense to me that the parent's `__init__` should be called first, then the subclass's `__init__` adds to it, and finally `__post_init__` (or any other name for it) is called, having access to all `__init__` code.

demos/2d_compressible_ALA/2d_compressible_ALA.py
Patol75: Looking at `update_approx` today, I remembered why I changed the order of arguments. Given that temperature becomes optional in `StokesSolver`, it now has a default value (`fd.Constant(0.0)`). `approximation`, on the other hand, is not optional and does not have a default value. Therefore, it must precede `T` in the initialisation call. So that left two options: `z, approximation, T` or `approximation, z, T`. I thought the second one made more sense, hence why my initial proposed changes here. Happy to hear your thoughts.

angus-g: This isn't `update_approx`, so I think it doesn't need to be making premature API changes. It's much easier to reason about the changes if the API stays the same. If it eventually becomes optional in `StokesSolver`, then it could make sense to make a more sweeping change to the order of arguments.

gadopt/energy_solver.py
Patol75: @drhodrid What is your update?

drhodrid: Ohh I didn't realise I needed to give one!! Leave it as it is for now - and we can look at this in a subsequent pull request.

gadopt/energy_solver.py
Patol75: @w3168 Any updates on this?

gadopt/transport_solver.py
```diff
@@ -0,0 +1,368 @@
...
+        delta_t: Constant,
+        timestepper: RungeKuttaTimeIntegrator,
+        **kwargs,
+    ) -> None:
+        self.u = u
+        self.approximation = approximation
+
+        super().__init__(solution, delta_t, timestepper, **kwargs)
+
+        self.T_old = self.solution_old
```
Patol75, line 346: This is how `__init__` in `EnergySolver` looks like without the metaclass. I am not a fan of it, but do other devs prefer it better like that?

gadopt/energy_solver.py
w3168: I think self.strong_bcs_tag is ok. you could simplify it to `self.strong_bc`

Patol75:
This should be good to go provided that CI returns green and devs are happy with the changes.

gadopt/transport_solver.py
```diff
@@ -0,0 +1,372 @@
...
+
+    @abc.abstractmethod
+    def set_equation(self):
+        """Sets up the term contributions in the equation."""
+        raise NotImplementedError
+
+    def set_solver_options(self) -> None:
+        """Sets PETSc solver parameters."""
+        if isinstance(solver_preset := self.solver_parameters, dict):
+            return
```
angus-g, line 200:
```suggestion
        solver_preset = self.solver_parameters
        if isinstance(solver_preset, dict):
            return
```

Prefer this to be more explicit

tests/scalar_advection_diffusion/scalar_advection_diffusion.py
```diff
@@ -58,27 +58,28 @@
...
-# the choice of units is up to the user. We use the diagonaly implicit DIRK33 Runge-Kutta
-# method for timestepping. 'T' means that the boundary conditions will be applied strongly
-# by the energy solver.
-approximation = BoussinesqApproximation(Ra=1, kappa=kappa)
-q_top = 1.0
-q_bottom = 0.0
-bcs = {3: {'T': q_bottom}, 4: {'T': q_top}}
-energy_solver = EnergySolver(q, u, approximation, dt, DIRK33, bcs=bcs, su_advection=True)
+# Use G-ADOPT's GenericTransportSolver to advect the tracer. We use the diagonally
+# implicit DIRK33 Runge-Kutta method for timestepping. 'T' means that the boundary
```
angus-g, line 62: BCs are no longer 'T'

tests/scalar_advection_diffusion/scalar_advection_diffusion_DH219_skew.py
```diff
@@ -44,27 +44,29 @@
 T = 10.
 dt = 0.01
 
-# Use G-ADOPT's Energy Solver to advect the tracer. By setting the Rayleigh number to 1
-# the choice of units is up to the user. We use the diagonaly implicit DIRK33 Runge-Kutta
-# method for timestepping. 'T' means that the boundary conditions will be applied strongly
-# by the energy solver.
-approximation = BoussinesqApproximation(Ra=1, kappa=kappa)
+# Use G-ADOPT's GenericTransportSolver to advect the tracer. We use the diagonally
+# implicit DIRK33 Runge-Kutta method for timestepping. 'T' means that the boundary
```
angus-g, line 48: BCs are no longer 'T'

gadopt/transport_solver.py
```diff
@@ -0,0 +1,372 @@
...
+class GenericTransportSolver(GenericTransportBase):
+    """Advances in time a generic transport equation.
+
+    All combinations of advection, diffusion, sink, and source terms are handled.
+
+    Note: The solution field is updated in place.
+
+    Args:
+      terms:
+        List of equation terms (refer to terms_mapping)
```
angus-g, line 256: Given that this is intended for user-facing code, I think the expected terms should be explicitly detailed in the docstring. The other concern is that depending on the terms, there are different arguments expected in `eq_attrs`? This needs to be explicit as well.

angus-g (commented):
Thanks Thomas, I think this is a much cleaner approach. A little bit of extra documentation is required, but other than that it looks good!

gadopt/transport_solver.py
Patol75: That is fair. I realise now that I probably copied that logic from [stokes_integrators](https://github.com/g-adopt/g-adopt/blob/3dffa3b1427695f90b313b3f72dfd229808053d7/gadopt/stokes_integrators.py#L241), where it makes sense, but it is not required here. I simplified.

gadopt/transport_solver.py
Patol75: What do you think of my little table?

angus-g: Thanks! I've just tweaked its formatting so it renders correctly on the API pages, but exactly the thing I had in mind.

Patol75:
Thank you, @angus-g, for the nice documentation improvements! I just merged `master`; if CI returns a green tick, is this PR good to go?

drhodrid:
Is this one good to go now?

angus-g (approved):
I think it's probably fine if it passes tests. API-breaking stuff like this should be minimised, but I think we're still going to have to change a few things before we're more concrete about it.

PR #115, opened 2024-09-25
Add DG viscoplastic case

drhodrid:
As it says above - in an attempt to bolster our testing of the DG code path. 

Final results for a serial run at my end are:
`5117 0.6343800669770835 0.00012097921667851196 9.992641565430135e-10 79.06535579885782 75.09711547890507 88.87478827562207 6.694714569011368 6.711845622222197 0.01713105321082864 0.527520897902003`

Let's see if the CI (on 4 cores) matches this, allowing us to further isolate the parallel issue identified with the cylindrical_TALA_DG test.


angus-g:
Looks like the 4 core run does indeed match your serial expected results.

angus-g:
Looks like we're matching the serial results in parallel. Only other thought I had here: we're gathering quite an assortment of test cases now. Is it worth having a README file in the tests directory just to give an idea of what the various cases are doing/focusing on?

drhodrid:
Hi @angus-g. Thanks for this. On point 1, I think this implies that the issues seen in the TALA DG case are to do with iterative solvers in parallel, because this new test case uses MUMPS. I will look into this next. On point 2 - yes I think a README would be useful. OK for me to put that in as a seperate pull request, or should it go with this?

angus-g:
> OK for me to put that in as a seperate pull request, or should it go with this?

We can do that separately, mainly just raising it while it was fresh in the brain! I'll get this one in now.

drhodrid:
Perfect!

angus-g (approved):

PR #120, opened 2024-10-03
Fix pytest directory for longtest

angus-g:
We also skip the unit test directory, since pytest collection will fail when trying to import Firedrake in some of the tests.


w3168 (approved):
Since the test passed here I guess this is fine...!
https://github.com/g-adopt/g-adopt/actions/runs/11156390899

PR #121, opened 2024-10-14
Move stress definition and input viscosity to approximation

w3168:
This moves the definition of stress to the approximation. It also removes viscosity as an input argument to stokes solver instead being passed through the approximation. 

This should simplify the setup e.g. see Issue #111 and the PR #112 to try and fix it.

I would like to implement something like this so that it is easier to change the definition of stress for viscoelastic consitutive equations which depend on time.


gadopt/approximations.py
```diff
@@ -335,6 +347,11 @@ def __init__(self,
         self.Tbar = Tbar
         self.cp = cp
 
+    def stress(self, u):
+        stress = super().stress(u)
+        dim = len(u)  # Geometric dimension, i.e. 2D or 3D
```
angus-g, line 352: Is this always the same as `u.function_space().mesh().geometric_dimension()` (which the previous definition used)? Other than that, it's curious that there ends up being a big difference in the TALA DG benchmark...

w3168:
Output of tala dg tests now matches PR #112 
https://github.com/g-adopt/g-adopt/actions/runs/11337666787

angus-g:
Code-wise this looks all good and supersedes #112 (since it basically incorporates it). I'd like to see if I can learn anything about the TALA DG case in parallel, but in the meantime maybe we can make that test run in serial so that at least it's deterministic.

angus-g:
Thanks, that is expected. I think we can be confident that it's correct, but I'm against doubling the CI runtime by serialising the test. I'll try to get it to pass on either of the possible outputs for the moment, and try to figure out what the underlying cause is.

angus-g (approved):
I think we're all good now!

PR #122, opened 2024-10-15
Viscoelastic with stress defined in approximation

w3168:
This is another go at viscoelasticity based off #121. The motivation is to tidy up the code so that the stress can be calculated through the approximation. Hopefully this should be tidier than #117 and it makes adding in the internal variable approach easier where the stress definition is quite different. 

It still passes the same tests as #117 (with the same values) : 1) convergence tests using incremental displacement formulation for three end members (elastic, viscoelastic and viscous) for an analytical surface loading problem. 2) EDIT: ~A regression test for a 2d loading problem based off the 3d test case from weerdesteijn et al. 2023.~ Now consistent with other demos testing through params.log (on a reworked coarser/faster demo example)

EDIT: ~The demo is still fairly messy. I have added equations at the beginning and some text. I need to add some plots. Also I would like to make the run time quicker by coarsening the mesh and making the timestep larger (I think both are quite conservative estimates at the moment).~ I've made a demo. The aim is for this to be a standalone to the mantle convection demos, i.e. a base case for GIA if people don't want to read the mantle convection problems first. 

After speaking with @Patol75 I have kept the stokes and viscoleastic stokes solver separate for now. There is a remaining issue that stokes expects temperature as an argument but this doesnt make sense for GIA. Instead fields like displacement and (old) deviatoric stress are needed in the solver. Possibly we could streamline the solvers to change the arguments to be passed as a user defined fields dictionary. and then depending on the approximation provided we would expect to find different fields...


demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+# Now let's run the simulation! We are going to control the ice thickness using the `ramp` parameter. At each step we call `solve` to calculate the incremental displacement and pressure fields. This will update the displacement at the surface and stress values accounting for the time dependent Maxwell consitutive equation.
+
+for timestep in range(max_timesteps):
+    ramp.assign(conditional(time < T1_load, time / T1_load,
+                            conditional(time < T2_load, 1 - (time - T1_load) / (T2_load - T1_load),
+                                        0)
+                            )
+                )
+
+    ice_load.interpolate(ramp * rho_ice * g * Hice * disc)
```
Patol75, line 407: Whilst running this demo on `update_approx`, I realised that diagnostics change if `ice_load` is defined as some [UFL algebra](https://github.com/g-adopt/g-adopt/blob/63a38d0d9c1c3183a9dd6c3bdc309c46200de638/demos/viscoelastic/viscoelastic_loading.py#L404) instead of a `Function`. Changes are small enough to preserve the physical interpretation of the simulation but large enough to be [caught by the CI](https://github.com/g-adopt/g-adopt/actions/runs/11339720680).

w3168: maybe it is the disc? I guess if we interpolate then we are forcing it to be a P1 field but otherwise it might do a different (better?) approximation of the tanh function if it is still an expression... @angus-g any ideas?

demos/Makefile
```diff
@@ -1,5 +1,5 @@
-cases := base_case 2d_compressible_ALA 2d_compressible_TALA viscoplastic_case 2d_cylindrical 3d_spherical 3d_cartesian adjoint multi_material gplates_global free_surface
-demo_cases := base_case 2d_cylindrical 3d_spherical 2d_compressible_ALA visualise_ALA_p_nullspace 2d_compressible_TALA viscoplastic_case 3d_cartesian gplates_global multi_material/compositional_buoyancy multi_material/thermochemical_buoyancy adjoint free_surface
+cases := base_case 2d_compressible_ALA 2d_compressible_TALA viscoplastic_case 2d_cylindrical 3d_spherical 3d_cartesian adjoint multi_material gplates_global free_surface viscoelastic
+demo_cases := base_case 2d_cylindrical 3d_spherical 2d_compressible_ALA visualise_ALA_p_nullspace 2d_compressible_TALA viscoplastic_case 3d_cartesian gplates_global multi_material/compositional_buoyancy multi_material/thermochemical_buoyancy adjoint free_surface viscoelastic
```
drhodrid, line 2: For ease of navigation, as the number of demos and tests grow, I'd say start a viscoelastic folder, and mark this as base. I think it's probably now an idea to also add a mantle_dynamics folder, throwing all other demos in there, excluding multi_material, which may justify it's own folder.

gadopt/approximations.py
```diff
@@ -411,3 +411,70 @@ def buoyancy(self, p, T):
         pressure_part = self.dbuoyancydp(p, T) * p
         temperature_part = super().buoyancy(p, T)
         return pressure_part + temperature_part
+
+
+class SmallDisplacementViscoelasticApproximation():
+    """Expressions for the small displacement viscoelastic approximation.
+
+    Small displacement linearises the problem. Perturbation about a reference state.
```
drhodrid, line 419: By assuming a small displacement, we can linearise the problem, assuming a perturbation away from a reference state.

gadopt/stokes_integrators.py
```diff
@@ -228,18 +228,26 @@ def __init__(
...
 
+        self.k = upward_normal(self.mesh)
+
         # Setup boundary conditions
         self.weak_bcs = {}
         self.strong_bcs = []
-        free_surface_dict = {}  # Separate dictionary for copying free surface information
+
+        # Free surface parameters
+        self.free_surface_dict = {}  # Separate dictionary for copying free surface information
```
drhodrid, line 248: What do you mean by "copying free surface information"?

tests/Makefile
```diff
@@ -1,4 +1,4 @@
-cases := analytical_comparisons multi_material optimisation_checkpointing scalar_advection_diffusion Drucker-Prager_rheology adjoint 2d_cylindrical_TALA_DG viscoplastic_case_dg free_surface
+cases := analytical_comparisons multi_material optimisation_checkpointing scalar_advection_diffusion Drucker-Prager_rheology adjoint 2d_cylindrical_TALA_DG viscoplastic_case_dg free_surface viscoelastic
```
drhodrid, line 1: See previous comment on names/directories.

tests/viscoelastic/zhong_viscoelastic_free_surface.py
```diff
@@ -0,0 +1,188 @@
...
+    W = FunctionSpace(mesh, "CG", 1)  # Pressure function space (scalar)
+    Q = FunctionSpace(mesh, "CG", 2)  # Temperature function space (scalar)
+    Z = MixedFunctionSpace([V, W])  # Mixed function space.
+    TP1 = TensorFunctionSpace(mesh, "DG", 1)
+    R = FunctionSpace(mesh, "R", 0)
+
+    # Function to store the solutions:
+    z = Function(Z)  # a field over the mixed function space Z.
+    u, p = split(z)  # Returns symbolic UFL expression for u and p
+    u_, p_ = z.subfunctions  # Returns individual Function for output
```
drhodrid, line 56: Can we drop this and use *z.subfunctions or similar when outputting lower down? This would be consistent with our other demos.

tests/viscoelastic/zhong_viscoelastic_free_surface.py
```diff
@@ -0,0 +1,188 @@
...
+    Cs = (1.-b) * sinh(a*z_scaled) / sinh(a) + b*(tanh(a*(z_scaled + 0.5))/(2*tanh(0.5*a)) - 0.5)
+    Vc = mesh.coordinates.function_space()
+    f = Function(Vc).interpolate(as_vector([x, depth_c*z_scaled + (D - depth_c)*Cs]))
+    mesh.coordinates.assign(f)
+    left_id, right_id, bottom_id, top_id = 1, 2, 3, 4  # Boundary IDs
+
+    # Set up function spaces - currently using P2P1 element pair:
+    V = VectorFunctionSpace(mesh, "CG", 2)  # Displacement function space (vector)
+    W = FunctionSpace(mesh, "CG", 1)  # Pressure function space (scalar)
+    Q = FunctionSpace(mesh, "CG", 2)  # Temperature function space (scalar)
```
drhodrid, line 48: Feels weird calling it a T function space in a viscoelastic model.

tests/viscoelastic/zhong_viscoelastic_free_surface.py
```diff
@@ -0,0 +1,188 @@
...
+    p_.rename("Pressure")
+
+    displacement = Function(V, name="displacement")
+    stress_old = Function(TP1, name='stress old')
+
+    # Output function space information:
+    log("Number of Velocity DOF:", V.dim())
+    log("Number of Pressure DOF:", W.dim())
+    log("Number of Velocity and Pressure DOF:", V.dim()+W.dim())
+    log("Number of Temperature DOF:", Q.dim())
```
drhodrid, line 68: Drop T DOF? Maybe add Free surface DOF instead, or similar?

drhodrid (commented):
Minor things from me. Going to look at demo carefully next.

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+# pointing in the radially inward direction.
+#
+# Boundaries are automatically tagged by the built-in meshes supported by Firedrake. For the `RectangleMesh` being used here, tag 1 corresponds to the plane $x=0$; 2 to the $x=L$ plane; 3 to the $y=0$ plane; and 4 to the $y=D$ plane. For convenience, we can rename these to `left_id`, `right_id`, `bottom_id` and `top_id`.
+
+# +
+# Set up geometry:
+L = 1500e3  # length of the domain in m
+D = 2891e3  # Depth of domain in m
+nx = 40  # Number of horizontal cells
+nz = 40  # number of vertical cells
```
drhodrid, line 194: Capitalisation inconsistent in 4 lines above (Number, Length/depth).

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+
+# Let's print out the grid resolution in km
+log(f"Horizontal resolution {L/nx/1000} km")
+log(f"Vertical resolution {D/nz/1000} km")
+
+mesh = RectangleMesh(nx, nz, L, D, name="mesh", quadrilateral=True)
+mesh.cartesian = True
+
+# Shift top of mesh to z = 0.
```
drhodrid, line 203: This needs a little more of a detailed justification. Why is this done?

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+# We now need to choose finite element function spaces. `V` , `W`, `TP1` and `R` are symbolic
+# variables representing function spaces. They also contain the
+# function space’s computational implementation, recording the
+# association of degrees of freedom with the mesh and pointing to the
+# finite element basis. We will choose Q2-Q1 for the mixed incremental displacement-pressure similar to our mantle convection demos. This is a Taylor-Hood element pair which has good properties for Stokes modelling. We also initilaise a discontinuous tensor function space that wil store our previous values of the deviatoric stress, as the gradient of the continous incremental displacement field will be discontinuous.
+
+# Set up function spaces - currently using the bilinear Q2Q1 element pair:
+V = VectorFunctionSpace(mesh, "Q", 2)  # (Incremental) Displacement function space (vector)
+W = FunctionSpace(mesh, "Q", 1)  # Pressure function space (scalar)
+TP1 = TensorFunctionSpace(mesh, "DQ", 2)  # (Discontinuous) Stress tensor function space (tensor)
```
drhodrid, line 217: Maybe me being a pedant, but it would be nice if functions spaces were all consistently named - perhaps 1 letter for each? S or Y maybe instead of TP1?

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+# Now we can set up the background profiles for the material properties. In this case the density, shear modulus and viscosity only vary in the vertical direction. We will approximate the series of layers using a smooth tanh function with a width of 20 km.
+
+
+# +
+# layer properties from spada et al 2011
+radius_values = [6371e3, 6301e3, 5951e3, 5701e3]
+density_values = [3037, 3438, 3871, 4978]
+shear_modulus_values = [0.50605e11, 0.70363e11, 1.05490e11, 2.28340e11]
+viscosity_values = [1e25, 1e21, 1e21, 2e21]
+# N.b. that we have modified the viscosity of the Lithosphere viscosity from
```
drhodrid, line 265: superfluous "viscosity" here

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+initialise_background_field(density, density_values)
+
+shear_modulus = Function(W, name="shear modulus")
+initialise_background_field(shear_modulus, shear_modulus_values)
+
+viscosity = Function(W, name="viscosity")
+initialise_background_field(viscosity, viscosity_values)
+# -
+
+# Next let's create a function to store our ice load. Following the long test from Weeredesteijn et al 2023, during the first 90 thousand years of the simulation the ice sheet will grow to a thickness of 1 km. The ice thickness will rapidly shrink to ice free conditons in the next 10 thousand years. Finally, the simulation will run for a further 10 thousand years to allow the system to relax towards isostatic equilibrium. This is approximately the length of an interglacial-glacial cycle. The width of the ice sheet is 100 km and we have used a tanh function again to smooth out the transition.
```
drhodrid, line 291: Conditions.

drhodrid, line 291: to smooth out the transition "from ice, to ice-free regions".

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+disc_dx = 5e3
+k_disc = 2*pi/(8*disc_dx)  # wavenumber for disk 2pi / lambda
+r = X[0]
+disc = 0.5*(1-tanh(k_disc * (r - disc_radius)))
+ramp = Constant(0)
+
+
+# -
+
+# Next let's define the length of our time step. If we want to accurately resolve the elastic response we should choose a timestep lower than the Maxwell time, $\alpha = \eta / \mu$. The Maxwell time is the time taken for the viscous deformation to 'catch up' the initial, instantaneous elastic deformation.
```
drhodrid, line 315: 'catch up' WITH the initial

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+#
+# Let's print out the maxwell time for each layer
+
+for layer_visc, layer_mu in zip(viscosity_values, shear_modulus_values):
+    log(f"Maxwell time: {float(layer_visc/layer_mu/year_in_seconds):.0f} years")
+
+# As we can see the shortest Maxwell time is given by the lower mantle and is about 280 years, i.e. it will take about 280 years for the viscous deformation in that layer to catch up any instantaneous elastic deformation. Conversely the top layer, our lithosphere, has a maxwell time of 6 million years. Given that our simulations only run for 110000 years the viscous deformation over the course of the simulation will always be negligible compared with the elastic deformation. For now let's choose a timestep of 250 years and an output time step of 2000 years.
```
drhodrid, line 322: Capitalisation of Maxwell needs to be consistent throughout.

drhodrid, line 322: An output "frequency" of 2000 years.

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+# +
+# Timestepping parameters
+Tstart = 0
+time = Function(R).assign(Tstart * year_in_seconds)
+
+dt_years = 250
+dt = Constant(dt_years * year_in_seconds)
+Tend_years = 110e3
+Tend = Constant(Tend_years * year_in_seconds)
+dt_out_years = 2e3
```
drhodrid, line 333: For consiustency with other demos, perhaps output_frequency_years instead of dt_out_years?

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+
+max_timesteps = round((Tend - Tstart * year_in_seconds) / dt)
+log("max timesteps: ", max_timesteps)
+
+dump_period = round(dt_out / dt)
+log("dump_period:", dump_period)
+log(f"dt: {float(dt / year_in_seconds)} years")
+log(f"Simulation start time: {Tstart} years")
+
+do_write = True
```
drhodrid, line 344: Tell a user what this flag does.

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+# -
+
+# We can now define the boundary conditions to be used in this simulation.  Let's set the bottom and side boundaries to be free slip with no normal flow $\textbf{u} \cdot \textbf{n} =0$. By passing the string `ux` and `uy`, G-ADOPT knows to specify these as Strong Dirichlet boundary conditions.
+#
+# For the top surface we need to specify a normal stress, i.e. the weight of the ice load, as well as indicating this is a free surface.
+#
+# The `delta_rho_fs` option accounts for the density contrast across the free surface whether there is ice or air (or in later examples ocean!) above a particular region of the mantle.
```
drhodrid, line 351: ice, water, or air (drop comment re later examples for now).

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+    top_id: {'normal_stress': ice_load, 'free_surface': {'delta_rho_fs': density - exterior_density}},
+    1: {'ux': 0},
+    2: {'ux': 0},
+}
+
+gd = GeodynamicalDiagnostics(z, density, bottom_id, top_id)
+# -
+
+
+# We also need to specify a G-ADOPT approximation which sets up the various parameters and fields needed for the viscoelastic loading problem.
```
drhodrid, line 367: Stokes system (not systems)

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+#     plotter.clear()
+#
+# # Closes and finalizes movie
+# plotter.close()
+# -
+# Looking at the animation, we can see that as the weight of the ice load builds up the mantle deforms, pushing up material away from the ice load. If we kept the ice load fixed this forebulge will eventually grow enough that it balances the weight of the ice, i.e the mantle is in isostatic isostatic equilbrium and the deformation due to the ice load stops. At 100 thousand years when the ice is removed the topographic highs associated with forebulges are now out of equilibrium so the flow of material in the mantle reverses back towards the previously glaciated region.
+
+# ![SegmentLocal](displacement_warp.gif "segment")
+
+# In this demo we have seen how to setup a simple viscoelastic loading problem. In the next demos we will start to look at more realistic cases in Earth-like geometry and with additional physics including gravity, transient rheology and sea level.
```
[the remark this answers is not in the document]
drhodrid: I think we should drop this promise, unless there are plans to get some of this in prior to the workshop.

demos/viscoelastic/viscoelastic_loading.py
```diff
@@ -0,0 +1,517 @@
...
+#
+# Subbing into the stress balance gives
+#
+# \begin{equation}
+#     \nabla \cdot \boldsymbol{\sigma}_1 = \nabla \cdot (\boldsymbol{\sigma}_{L1}  - \textbf{u} \cdot \nabla \boldsymbol{\sigma}_0 ) = \nabla \cdot \boldsymbol{\sigma}_{L1} - \nabla (\rho_0 g u_r)
+# \end{equation}
+# where $u_r$ is the radial displacement vector.
+#
+
+# Maxwell Rheology
```
drhodrid, line 98: Use of \cdot on epsilon terms below seems inconsistent.

demos/viscoelastic/viscoelastic_loading.py
w3168: (ah I actually assumed z was already like this for setting up the depth profiles. I've just modified it in the intialise background profile now

gadopt/stokes_integrators.py
w3168: sorry that was an unhelpful comment. it is just a dictionary to store user specified parameters about the free surface from a runscript. I've taken the comment out now

tests/viscoelastic/zhong_viscoelastic_free_surface.py
w3168: yep thanks for spotting this!

demos/Makefile
w3168: I'll have a go at doing this!

w3168 (commented):
Thanks for the suggestions. I still need to change the directory structure but hopefully nearly there on the changes suggested so far.

demos/viscoelastic/viscoelastic_loading.py
w3168: So i've gone with your suggestion of doing the ice load symbolically as it is very simple. we are not currently plotting the ice load as well so I think this is fine. 

As discussed on zulip with a coarse mesh you can get fairly different 10% effects integrating the ice load over the surface whether using an expression (with default quadrature of 6) vs a P1 function. I think the expression is actually a bit more accurate (closer to the resolved value) for coarse meshes which makes sense.

demos/glacial_isostatic_adjustment/gia_base_case/Makefile
```diff
@@ -0,0 +1,8 @@
+all: params.log
+
+params.log: viscoelastic.py
```
Patol75, line 3: @w3168 You forgot that one :p

w3168: thanks!

gadopt/stokes_integrators.py
```diff
@@ -536,3 +555,88 @@ def ala_right_nullspace(
...
+        **kwargs,
+    ):
+        self.stress_old = stress_old  # Function to store deviatoric stress from previous time step
+        self.displacement = displacement
+        self.dt = dt
+
+        approximation.mu = approximation.effective_viscosity(self.dt)
+
+        # This isn't used in the viscoelastic code but StokesSolver currently assumes temperature is required as an argument
+        T_placeholder = approximation.density
```
Patol75, line 601: Why not just set `T` to 0?

w3168: good idea - I'll just check the recent changes didnt break the ci then will do this

demos/glacial_isostatic_adjustment/gia_2d_cylindrical/unstructured_annulus_refined_surface.msh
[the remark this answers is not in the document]
Patol75: Oh sorry, I did not realise you have a Python file that generates the MSH file. So you could either run it from the `Makefile` or `import` it or use `subprocess` in your main script. @angus-g What do you think is best?

dsroberts:
Since the Makefile conflicts are my doing, I'll have a go at resolving them.

w3168:
> Since the Makefile conflicts are my doing, I'll have a go at resolving them.

Thanks @dsroberts !

demos/glacial_isostatic_adjustment/gia_2d_cylindrical/unstructured_annulus_refined_surface.msh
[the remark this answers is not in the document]
w3168: I've made a generate_mesh function so this can be called in the jupyter notebook (and avoiding linting errors)

Patol75:
> Since the Makefile conflicts are my doing, I'll have a go at resolving them.

Sorry, I just realised that this was related to the new directory tree. I have a working version [here](https://github.com/g-adopt/g-adopt/blob/update_approx/demos/Makefile). Not that you need it, but in case it is useful.

demos/glacial_isostatic_adjustment/gia_2d_cylindrical/gia_2d_cylindrical.py
```diff
@@ -0,0 +1,550 @@
...
+disc_halfwidth2 = (2*pi/360) * 20  # Disk half width in radians
+surface_dx = 200*1e3
+ncells = 2*pi*radius_values[0] / surface_dx
+surface_resolution_radians = 2*pi / ncells
+colatitude = atan2(X[0], X[1])
+disc1_centre = (2*pi/360) * 25  # centre of disc1
+disc2_centre = pi  # centre of disc2
+disc1 = 0.5*(1-tanh((abs(colatitude-disc1_centre) - disc_halfwidth1) / (2*surface_resolution_radians)))
+disc2 = 0.5*(1-tanh((abs(abs(colatitude)-disc2_centre) - disc_halfwidth2) / (2*surface_resolution_radians)))
+ice_load = Function(W).interpolate(Constant(1)*rho_ice * g * (Hice1 * disc1 + Hice2 * disc2))
```
Patol75, line 257: Should we drop the `Function` here?

demos/glacial_isostatic_adjustment/gia_2d_cylindrical/gia_2d_cylindrical.py
```diff
@@ -0,0 +1,550 @@
...
+log("Number of Incremental Displacement DOF:", V.dim())
+log("Number of Pressure DOF:", W.dim())
+log("Number of Velocity and Pressure DOF:", V.dim()+W.dim())
+
+# We can now visualise the resulting mesh. As you can see there is finer resolutation near the surface compared with the lower mantle.
+
+VTKFile("mesh.pvd").write(Function(V))
+mesh_data = pv.read("mesh/mesh_0.vtu")
+edges = mesh_data.extract_all_edges()
+plotter = pv.Plotter(notebook=True)
```
dsroberts, line 69: There is some graphical stuff here that needs to be gated with 
```
# + tags=["active-ipynb"]
```

demos/glacial_isostatic_adjustment/gia_2d_cylindrical/gia_2d_cylindrical.py
```diff
@@ -0,0 +1,550 @@
...
+# Let's have a quick look at the density field using pyvista. We can see that the mesh is still quite coarse, but it is able to capture the layered structure.
+
+# +
+# Read the PVD file
+density_file = VTKFile('density.pvd').write(density)
+reader = pv.get_reader("density.pvd")
+data = reader.read()[0]  # MultiBlock mesh with only 1 block
+
+# Create a plotter object
+plotter = pv.Plotter(shape=(1, 1), border=False, notebook=True, off_screen=False)
```
dsroberts, line 121: and here

demos/glacial_isostatic_adjustment/gia_2d_cylindrical/gia_2d_cylindrical.py
```diff
@@ -0,0 +1,550 @@
...
+# Now let's plot the normalised viscosity viscosity field on a log plot (we have divided the viscosity by 1x10$^{23}$ Pa s). Again although we are using a coarse mesh we are able to capture the key features of the viscosity field.
+
+# +
+# Read the PVD file
+visc_file = VTKFile('viscosity.pvd').write(normalised_viscosity)
+reader = pv.get_reader("viscosity.pvd")
+data = reader.read()[0]  # MultiBlock mesh with only 1 block
+
+# Create a plotter object
+plotter = pv.Plotter(shape=(1, 1), border=False, notebook=True, off_screen=False)
```
dsroberts, line 208: here too

demos/glacial_isostatic_adjustment/gia_2d_cylindrical/gia_2d_cylindrical.py
```diff
@@ -0,0 +1,550 @@
...
+)
+
+transformed_arc_data = arc_data.transform(transform_matrix)
+ice_cmap = plt.get_cmap("Blues", 25)
+
+reader = pv.get_reader("viscosity.pvd")
+data = reader.read()[0]  # MultiBlock mesh with only 1 block
+
+# Create a plotter object
+plotter = pv.Plotter(shape=(1, 1), border=False, notebook=True, off_screen=False)
```
dsroberts, line 298: last one I think

demos/glacial_isostatic_adjustment/gia_2d_cylindrical/gia_2d_cylindrical.py
w3168: Thanks!

demos/glacial_isostatic_adjustment/gia_2d_cylindrical/gia_2d_cylindrical.py
w3168: thanks thomas just changed it

dsroberts:
I've just pushed the makefile change. There has been a slight structural change to go with this. Since we now have `mantle_convection`, `multi_material` and `glacial_isostatic_adjustment` demo directories, the demo names according to the makefiles are now `<directory>/<name>` rather than just `<name>`. This means that common cases like `base_case` won't conflict in the makefiles. As such, I've renamed `gia_base_case` to `base_case` and `gia_2d_cylindrical` to `2d_cylindrical` after chatting with @w3168 about it yesterday. I think `gia_base_case.py` and `gia_2d_cylindrical.py` might have to be renamed too for correct notebook generation, but I'm not sure.

w3168:
Thanks @dsroberts that sounds great!

angus-g:
I think that should be all the demo rendering stuff, other than some tweaks to the tutorials page/diagram to make navigating the categories a bit more streamlined.

drhodrid:
I had a quick look now at the cylindrical case. It looks nice in general, but there are minor (unnecessary) differences between some of its setup, and the equivalent base case, which I think we should minimise. For example - TP1 being used for function space names etc... I'd recommend using meld or similar to ensure they are both as in sync as possible. 

I'd also recommend using an extruded Firedrake mesh, rather than gmsh - just to keep things simpler and minimize the differences from the base case - the less changes one needs to explain in a tutorial, the better.

dsroberts:
Merge conflict sorted. Github didn't like that I used cherry-pick to grab the makefile changes from master. Had to do `git merge master`. There is literally no difference between 5aaea2532ebf1bd7f9530620046a51adec9a6ace and 010cebcfc73d0a70a6127e71f8560bd81a06d5cc

demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -0,0 +1,551 @@
...
+#         "position_y": 0.2,
+#         "vertical": True,
+#         "title_font_size": 20,
+#         "label_font_size": 16,
+#         "fmt": "%.0f",
+#         "font_family": "arial",
+#     }
+# )
+# plotter.camera_position = 'xy'
+# plotter.show()
```
angus-g, line 148: Do you use any of the interactivity in these plots? I discovered a way to ignore the big scary trame warning (screenshot below) when showing the plotter without `# plotter.show(jupyter_backend="static", interactive=False)` (as above in this file). But if there's no benefit to the interactivity, maybe we should just force it to be static?

image

w3168: At the moment I'm not using the interactiviity (maybe it would be nice for showing timestepping?) so happy to force it to b static

w3168:
> I had a quick look now at the cylindrical case. It looks nice in general, but there are minor (unnecessary) differences between some of its setup, and the equivalent base case, which I think we should minimise. For example - TP1 being used for function space names etc... I'd recommend using meld or similar to ensure they are both as in sync as possible.
> 
> I'd also recommend using an extruded Firedrake mesh, rather than gmsh - just to keep things simpler and minimize the differences from the base case - the less changes one needs to explain in a tutorial, the better.

@drhodrid thanks for checking earlier :) I think I've made these changes now

angus-g:
Alright, I did some labelling of the runs so the summary can pull out which module they're coming from. But I was too dumb to do the merge over to run the two `PDE_constrained_optimisation` cases so the tests are failing. Probably missed something exceedingly obvious!

angus-g (approved):

Issue #123, opened 2024-10-16
Parallel reproducibility issue of Cylindrical TALA DG case


angus-g:
When we run this case in parallel, we get different answers to serial. Previously, some parallel reproducibility issues were due to MUMPS and direct solves, but given this case is cylindrical, it defaults to an iterative solver.

| &nbsp; | `u_rms` | `nu_top` | `avg_t` | `FullT_min` | `FullT_max` |
| --- | --- | --- | --- | --- | --- |
| actual | 570.5405141652053 | 15.186137483046036 | 0.6363851935782374 | 0.0001693939995075 | 1.0011846065377914 |
| expected | 570.4729595213425 | 15.18647692671351 | 0.6363801581209881 | 0.0001688057305039 |1.0011841176113832 |
| relative error | **1.18404639e-04** | **2.23517060e-05** | 7.91259335e-06 | **3.47278537e-03** | 4.88347908e-07 |

Noting that the relative error tolerance is 1e-5 by default, and we only override that to 1e-4 in the `3d_cartesian` case.

_Originally posted by @angus-g in https://github.com/g-adopt/g-adopt/issues/112#issuecomment-2367174471_

Patol75:
Given the discussion in gh-124, this reproducibility issue likely arises from solver tolerances that were loosened too much.

PR #124, opened 2024-10-24
Improve TALA DG serial/parallel consistency

drhodrid:
…mprove match between serial and parallel runs. At least on my local system this fixes our parallel reproducibility issue - https://github.com/g-adopt/g-adopt/issues/123.


drhodrid:
Noting that the final log in my param file is:
`20 9.160323938349703e-05 4.8095544889431735e-06 571.4481428623797 4.061804619771714 15.19622807271593 11.134423452944215 0.6363433141867296 0.00016879887766496287 1.0011825311877605`

Subtly different from the previous results, largely because we now load both T and u from checkpoint.

tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py
```diff
@@ -123,12 +124,7 @@ def gaussian(center, c, a):
 mu_rad = Function(Q, name="Viscosity_Radial")  # Depth dependent component of viscosity
 interpolate_1d_profile(function=mu_rad, one_d_filename="initial_condition_mat_prop/mu2_radial.txt")
 
-# We next set up and initialise our Temperature field from a checkpoint:
-X = SpatialCoordinate(mesh)
-T = Function(Q, name="Temperature")
-r = sqrt(X[0]**2 + X[1]**2)
-with CheckpointFile("initial_condition_mat_prop/Final_State.h5", mode="r") as f:
-    T = f.load_function(mesh, "Temperature")
+# We next set up and initialise auxilliary fields associated with temperature:
```
Patol75, line 127: `auxiliary`

Patol75:
What is the game changer here? Loading from checkpoint or tightening solver tolerances?

drhodrid:
Both are important. Loading from checkpoint ensures that the velocities after the first few timesteps are stable and sensible. The solver tolerances are the most important for consistency between serial and parallel howewver.

Patol75:
That makes sense. Do they converge towards a third value as you tighten the tolerances, or does one converge towards the other? In any case, is that a reasonable case to consider tightening the tolerances globally?

drhodrid:
Our default tolerances are actually higher than I specify in this test. I slacked them to ensure it was tractable for the CI. Even with my recent changes they are slacker than the defaults - but should produce reasonably comparable results from serial to 4 cores.

Patol75:
> Even with my recent changes they are slacker than the defaults

I have had a second look: here, you specify `ksp_rtol` of `1e-7` and `1e-5`, but their [default values](https://github.com/g-adopt/g-adopt/blob/master/gadopt/stokes_integrators.py#L20-L50) are `1e-5` and `1e-4`. That seems inconsistent to me with your statement above.

The other change you have is adding a value of `1e-2` for `snes_rtol`; I am unsure what the default value is.

tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py
```diff
@@ -205,16 +201,16 @@ def gaussian(center, c, a):
...
                              nullspace=Z_nullspace, transpose_nullspace=Z_nullspace,
                              near_nullspace=Z_near_nullspace)
 stokes_solver.solver_parameters['snes_rtol'] = 1e-2
 stokes_solver.solver_parameters['fieldsplit_0']['ksp_converged_reason'] = None
-stokes_solver.solver_parameters['fieldsplit_0']['ksp_rtol'] = 1e-3
+stokes_solver.solver_parameters['fieldsplit_0']['ksp_rtol'] = 1e-7
 stokes_solver.solver_parameters['fieldsplit_1']['ksp_converged_reason'] = None
-stokes_solver.solver_parameters['fieldsplit_1']['ksp_rtol'] = 1e-2
+stokes_solver.solver_parameters['fieldsplit_1']['ksp_rtol'] = 1e-5
+
```
Patol75, line 213: I ran the case locally on 4 processes with your changes but I deleted these lines here (to use the respective G-ADOPT default values). Execution completed in under an hour and a half, and the `u_rms` diagnostic I obtained yields a relative difference on the order of `1e-6` compared with the CI run here.

drhodrid:
Thanks for following up here Thomas. My memory of solver tolerances was incorrect so thanks for checking. Saying that, if I leave solver tolerances as the default, which are fine for the majority of our cases, I get the following results in serial:

`20 9.160328945489007e-05 4.809562912621724e-06 571.4474654675657 4.061805094717602 15.196228783151183 11.13442368843358 0.6363433131125036 0.00016879903943008593 1.0011825317130207`

And the following in parallel (4 cores):

`20 9.160330333903966e-05 4.809563193360037e-06 571.4473757467131 4.0618052854758275 15.196228867684939 11.134423582209111 0.6363433127930306 0.0001687990309105361 1.0011825316879848`

Focussing on the RMS velocity, this differs in the 4th decimal place. 

My preference moving forward would be to drop all changes to the solver tolerances in the test case and proceed with the defaults, but this would necessitate a small change to the test tolerances. 

@angus-g if you are happy with this, could you update the expected values to those included above in serial, and I will then remove the tolerances and get this through the CI?

Patol75:
So you end up with an absolute difference of `9e-5` and a relative difference of `1e-7`. That should be good enough for the CI, right ?

drhodrid:
Hopefully it's good enough for the CI, but if not, I think we should just tweak the CI tolerance values in this case. And yes - a long test is probably wise.

angus-g:
Just copying a quick note from an offline discussion here: the required solver tolerances would _significantly_ increase the runtime of whichever test suite it gets added to. I think Rhodri is going to give some thought to a different strategy.

Patol75:
What kind of figures, in terms of the number of processes, run time, and solver tolerances, are we talking about here? And what would be acceptable?

angus-g:
This case is 4 cores, but takes about 3 hours. Given the entire rest of the suite runs in just over 2 hours on GitHub, or 40 minutes for the long tests (that I think could be renamed to _large_ tests), this is quite an increase...

Patol75:
Did that run have the `snes_rtol` value in? I had a much quicker run time earlier this week on 4 cores, which is why I find it odd.

angus-g:
Looks like on 16 cores, we bring the total runtime back to a more managable ~90-100 mins, although that's still about a 20-30 minute increase on previously.

drhodrid:
I think this is acceptable for now. Let's get it in and tested, and perhaps revisit early next year to see if we can improve solution time.

angus-g (approved):

Patol75:
Any reasons why these [lines](https://github.com/g-adopt/g-adopt/blob/5d31966983dca273688484834337ae9bfadaa11b/tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py#L209-L212) were committed? I thought we agreed not to include them. 

> My preference moving forward would be to drop all changes to the solver tolerances in the test case and proceed with the defaults, but this would necessitate a small change to the test tolerances.

Removing them would speed things up, and we could even go back to using 4 cores.

PR #128, opened 2024-10-29
Backend changes: Update level set

Patol75:
This third set of changes extracted from gh-102 proposes to revisit some of the level-set implementation logic. The main two changes are the removal of `density_RaB` and `Material`. The former is simply not needed, whilst the latter is not Pythonic. The new logic proposed here is to define material properties either within the run script (as is done in demos) or inside modules (as is done in tests). The class approach is simply awkward: writing a class and using it without the need to instantiate it likely means that writing a module was the appropriate thing to do. The proposed set of changes also includes several fixes and improvements. 

Needs gh-114.

### Removed
- `Material` Storing material properties in a class is not Pythonic. Users should either define material properties in their simulation script (as in updated demos) or in separate modules that can then be imported into the simulation script (as in updated tests).
- `density_RaB` Extraneous function replaced by explicit logic in `run_benchmark.py`.
### Modified
- `LevelSetSolver` The advection and reinitialisation components of the solver are now clearly separated. Users provide arguments for each component using dictionaries with documented keys. If both components are initialised, users can disable one for a given solve by providing an explicit boolean argument.
### Renamed
- `field_interface` -> `material_field`
### Added
- `signed_distance` G-ADOPT should provide a user-friendly way to generate values of the signed-distance function required to initialise the level-set function. Here, the user provides a geometrical description of the material interface — either a curve, polygon, or circle — to obtain a list of signed-distance values at the level-set nodes.
- `curve_interface` and `rectangle_interface` As some interface geometries arise frequently, G-ADOPT should provide convenience functions that automatically generate the keyword arguments expected by `signed_distance`. Here, the cases of a straight line, cosine curve, and rectangular objects are handled.
- `interface_thickness` G-ADOPT should provide a default strategy to generate the value of the interface thickness required to initialise the conservative level set. Here, a Firedrake function is created based on the cell sizes across the numerical domain.
- `conservative_level_set` It should not be up to the user to type the mathematical formula of the conservative level-set transformation. Here, we provide a helper function.
- `min_max_height` G-ADOPT should include a way to determine the extremum height or depth of the material interface, which is useful for diagnostic purposes. Here, the extremum is calculated using masking to isolate the sought-after node across the parallel subdivisions of the mesh.
### New demo
- Free subduction with free surface as described in Schmeling et al. (2008). This demo documents the setup of a dimensional case involving multi-material and free-surface functionalities.
### New test
- Salt and overburden dynamics as described in Woidt (1978). This test represents the dynamics of a dimensional system with a high aspect ratio and contrasting viscosities.


gadopt/level_set_tools.py
```diff
@@ -1,135 +1,247 @@
...
+        "interface_geometry": "polygon",
+    }
+
+    return signed_distance_kwargs
+
+
+def signed_distance(
+    level_set: fd.Function,
+    /,
+    interface_coordinates: list[list[float]] | np.ndarray | list[tuple[float], float],
```
w3168, line 110: As discussed in the demos, perhaps the first argument given could be *interface_geometry*. If it is *"curve"* then it looks at the *interface_args* which would include the *interface_x_coordinates*, callable/special curve like *cosine* etc and then uses *curve_interface* under the hood to get the *interface_coordinates*.

 Otherwise if polygon then you need to provide the *interface_coordinates* directly to *signed_distance*. Maybe this approach is simpler because it removes the extra *curve_interface* function in the demo which was only supposed to be a helper function to create the *interface_coordiantes*... What do you think?

gadopt/level_set_tools.py
```diff
@@ -274,258 +458,199 @@ def gradient_L2_proj(self) -> fd.variational_solver.LinearVariationalSolver:
...
+        if not self._solvers_ready:
             self.set_up_solvers()
-            self.solvers_ready = True
-
-        for subcycle in range(self.subcycles):
-            self.ls_solver.solve()
 
-            if step >= self.reini_params["frequency"]:
-                self.reini_ts.solution_old.assign(self.level_set)
+        if self.advection and not disable_advection:
```
w3168, line 528: Do you need two ways of turning on/off advection/reinitialisation?

w3168 (commented):
Hi @Patol75 this looks really nice! 

I hadn't really looked through the level sets before (sorry!) but this seems very neatly organised and very well documented and I guess follows what I was expecting from talking with you about it...!

Some general comments:
- I think removing the Material class and density_RaB seems reasonable
- Clearly separating the advection and reinitialisation seems like a good idea - especially as the reinitialisation still seems a bit adhoc re grid size / timesteps based on your comments in the docs.
- I think the new helper functions for generating the conservative level set, interface thickness and signed distance functions are very useful. I wrote some comments in the review about maybe making *curve_interface* hidden from the user in the demos... I found it a bit confusing that the function returns arguments to pass to another function. I ended up going to the source code to try and work out what was going on... I think it would be simpler if a user passes a few more options to *signed_distance* like the special curve type (cos or line) and then *curve_interface* is called by *signed_distance* and just returns *interface_coordinates*... That way the user only has to worry about the *signed_distance* function
- min_max_height - sorry I havent looked at this yet but it sounds useful
- New free surface demo - cool! I still need to run it!

Other comments
- I think some pictures of the initial level set contour / tanh/signed distance function (and the gmsh generated mesh) might be good for people in the demos to gain an intuition for how to set up the level set field
- I ran the (first two) demos with jupyter notebook on my laptop and they are pretty slow (more than 30minutes). I think at the least you could put some print statements every so often to give an indication for the user how the simulation is progressing. Can you just run them by default for a shorter number of timesteps and still see some meaningful evolution of the flow?
- I know you have done a lot of work testing all this and there are so many options for users to try (hence the need to choose defaults...!) Have you written up a document with this in or just a few pictures highlighting some of the pitfalls of the different methods? I guess I am still curious about reinitialisation and also what happens if you make the thickness of the interface larger than the grid cells. Similarly, I remember that you looked a lot at limiters and different functions spaces for psi too.  think none of the demos use limiters -  is that correct? Is it worth adding a short comment about these in the demos. (Apologies this probably was covered a lot in the previous PR!)

I'll look at the free surface demo in more detail asap.

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -197,92 +186,65 @@ def cosine_curve(x, amplitude, wavelength, vertical_shift):
...
 while True:
-    # Write output
-    if time_now >= output_counter * output_frequency:
-        output_file.write(*z.subfunctions, T, psi)
-        output_counter += 1
-
     # Update timestep
-    if time_end is not None:
-        t_adapt.maximum_timestep = min(output_frequency, time_end - time_now)
+    if time_end - time_now < output_frequency:
```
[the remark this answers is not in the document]
Patol75: The idea is to get the simulation to finish exactly at the set end time. The way I implemented it is to check at each loop iteration if the remaining time is shorter than the set output frequency, in which case the latter needs to be adjusted not to exceed the end time.

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -168,27 +136,48 @@ def cosine_curve(x, amplitude, wavelength, vertical_shift):
 # output frequency (in time units).
 
 time_now = 0  # Initial time
```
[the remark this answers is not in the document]
Patol75: I am not a fan of calling it `time` because `time` is a module in the Python standard library. I do not think it is good practice to reuse these names.

gadopt/level_set_tools.py
```diff
@@ -149,122 +261,194 @@ def reinitialisation_term(
...
+                "ksp_type": "preonly",
+                "pc_type": "bjacobi",
+                "sub_pc_type": "ilu",
+            }
+        if "steps" not in self.reini_kwargs:
+            self.reini_kwargs["steps"] = 1
+        if "frequency" not in self.reini_kwargs and self.mesh.cartesian:
+            self.reini_kwargs["frequency"] = self.reinitialisation_frequency()
+
+    def reinitialisation_frequency(self) -> int:
```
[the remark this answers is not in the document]
Patol75: Nothing to back it up, no, except perhaps my experience with the cases I implemented and tested. The rationale, I believe, is that smaller time steps generate smaller advection errors, so calling reinitialisation at every time step does not yield many benefits anymore, and it becomes very computationally intensive.

gadopt/level_set_tools.py
Patol75: `self.advection` internally records if the advection part of the solver has been initialised, whereas `disable_advection` is provided at solve time to indicate if advection should be omitted for the particular solve to follow. Essentially, I would say there is only one way of turning advection or reinitialisation off, but given that I use optional booleans and give them a default value of `True`, I need a way to make sure that a solver only executes if it has been initialised.

Patol75:
Hey Will,

Thank you for the thorough review and kind words; they are much appreciated. I put a few replies below:

- I had a go at getting rid of the `curve_interface` and `rectangle_interface` functions. The resulting `signed_distance_function` has become more involved, but I agree that the user interface is now more straightforward. I think @dsroberts also suggested in a previous developer meeting that such a change would be welcomed, so perhaps he could comment on the current implementation.
- I have added more visualisations to the demos, which should help newcomers better understand what happens. For the time required to run the demos, I have actually drastically sped the execution. However, these demos run best on 4 processing cores, so perhaps you ran them in serial.
- Sadly, there is no document. The main pitfall with reinitialisation at the moment is the accuracy of the gradient calculation, which, in my opinion, limits the potential benefits of that process. Making the thickness of the interface larger than the cell sizes is counter-productive: materials are considered immiscible, so each mesh vertex should represent a distinct material or sit extremely close to the material interface. Limiters were useful when working with a different form of the reinitialisation equation that required solving a Hamilton-Jacobi equation. There is not much to say about the level-set function space. Discontinuous has better advection properties, and increased polynomial degree brings more accurate results. I am happy to hear other developers' thoughts regarding how much of this information should be included in demos.

gadopt/level_set_tools.py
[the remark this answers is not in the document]
w3168: ok thanks!

gadopt/level_set_tools.py
w3168: Fair enough!

gadopt/level_set_tools.py
```diff
@@ -1,135 +1,246 @@
...
+        epsilon = epsilon.dat.data
+
+    return (1 + np.tanh(np.asarray(signed_distance) / 2 / epsilon)) / 2
 
 
 def reinitialisation_term(
     eq: Equation, trial: fd.Argument | fd.ufl.indexed.Indexed | fd.Function
 ) -> fd.Form:
-    """Term for the conservative level set reinitialisation equation.
+    """Term for the conservative-level-set reinitialisation equation.
```
sghelichkhani, line 243: Why the first hyphen?

Patol75: Without it, it feels like conservative is associated with equation, but maybe I am dreaming it.

Patol75:
For some reason, the thermochemical demo does not yield a reproducible rms velocity diagnostic... The mismatch is on the order of 1e-4, and results change at every run from what I can see.

gadopt/transport_solver.py
```diff
@@ -238,12 +238,14 @@ def solver_callback(self) -> None:
...
-    def solve(self, t: Number = 0, update_forcings: Callable | None = None) -> None:
+    def solve(
+        self, update_forcings: Callable | None = None, t: float | None = None
+    ) -> None:
         """Advances solver in time."""
         if not self._solver_ready:
             self.setup_solver()
 
-        self.ts.advance(t, update_forcings)
+        self.ts.advance(update_forcings, t)
```
sghelichkhani, line 248: Is this a change in interface?

Patol75: Time is only required if `update_forcings` is provided, and `update_forcings` does not always require time. It makes more sense to swap the order of arguments, which prevents having to define time when it is not needed.

gadopt/utility.py
```diff
@@ -450,14 +450,18 @@ def step_func(r, centre, mag, increasing=True, sharpness=50):
     )
 
 
-def node_coordinates(function):
-    """Extract mesh coordinates and interpolate them onto the relevant function space"""
-    func_space = function.function_space()
-    mesh_coords = SpatialCoordinate(func_space.mesh())
+def node_coordinates(function: Function) -> np.ndarray:
```
sghelichkhani, line 453: Not really relevant to this PR, is it?

Patol75: `node_coordinates` is only used in level-set initialisation, if I am not mistaken. I figured out a way to write it more efficiently compared to what I committed months ago. I think these changes should be included.

gadopt/transport_solver.py
sghelichkhani: What you are saying makes sense. But does not make this necessary. Changing interface might require a lot of time from everybody else to updating all their branches.

gadopt/utility.py
sghelichkhani: hmm ... is it more efficient? I think it takes exactly same amount of time, considering only interpolation does any computation.

gadopt/level_set_tools.py
sghelichkhani: I see!

gadopt/utility.py
Patol75: It is more efficiently written, meaning that it better leverages Firedrake syntax to achieve what it needs to. It also makes it simpler to read and understand. Coordinates are a vector field and, as such, should be handled using a VectorFunctionSpace.

gadopt/transport_solver.py
Patol75: I have seen quite a few occurrences of people providing time to `advance` when they did not need to or asking what this time is for. I think it is a useful change. I do not believe that having to update branches is a deal-breaker, especially with GitHub integrations in most IDEs.

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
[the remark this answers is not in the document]
sghelichkhani: Does that mean `output_frequency` is actualy of unit time? Then the naming can be adjusted.

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -197,96 +211,70 @@ def cosine_curve(x, amplitude, wavelength, vertical_shift):
...
-    # Write output
-    if time_now >= output_counter * output_frequency:
-        output_file.write(*z.subfunctions, T, psi)
-        output_counter += 1
-
     # Update timestep
-    if time_end is not None:
-        t_adapt.maximum_timestep = min(output_frequency, time_end - time_now)
+    if time_end - time_now < output_frequency:
+        t_adapt.maximum_timestep = time_end - time_now
```
sghelichkhani, line 235: Why adjusting `t_adapt` instead of directly `delta_t`?
```suggestion
delta_t.assign(time_end - time_now)
```

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -62,104 +64,95 @@
...
-
-# Initialise level set as a smooth step function
-psi.interpolate((1 + tanh(signed_dist_to_interface / 2 / epsilon)) / 2)
+from numpy import linspace  # noqa: E402
+
+# Initialise the level-set field. First, determine the signed-distance function at each
+# level-set node using a mathematical description of the material-interface location.
+# Then, define the thickness of the hyperbolic tangent profile used in the conservative
+# level-set approach. Finally, overwrite level-set data array.
+interface_coords_x = linspace(0, lx, 1000)
```
sghelichkhani, line 103: I think this is nice! Good job on cleaning up the signed distance stuff!

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -62,104 +64,95 @@
...
 lx, ly = 0.9142, 1  # Domain dimensions in x and y directions
 # Rectangle mesh generated via Firedrake
 mesh = RectangleMesh(nx, ny, lx, ly, quadrilateral=True)
-mesh.cartesian = True
+mesh.cartesian = True  # Tag the mesh as Cartesian to inform other G-ADOPT objects.
 left_id, right_id, bottom_id, top_id = 1, 2, 3, 4  # Boundary IDs
 
-V = VectorFunctionSpace(mesh, "CG", 2)  # Velocity function space (vector)
-W = FunctionSpace(mesh, "CG", 1)  # Pressure function space (scalar)
+V = VectorFunctionSpace(mesh, "Q", 2)  # Velocity function space (vector)
```
sghelichkhani, line 74: Can you say why you are using `Q` elements now? Isn't it the same as `CG` when having `quadrilateral=True` in mesh?

gadopt/transport_solver.py
sghelichkhani: This is a change in interface that is affecting `time_stepper.py`. If you still insist on changing interface, this should be another PR.

gadopt/time_stepper.py
```diff
@@ -22,69 +22,78 @@ class TimeIntegratorBase(ABC):
...
+            Callable updating any time-dependent equation forcings
+          t:
+            Current simulation time
 
         """
-        pass
 
     @abstractmethod
-    def initialize(self, init_solution):
+    def initialise(self, init_solution):
```
sghelichkhani, line 37: All the changes here should be a seperate PR! These are not level-set changes.

gadopt/time_stepper.py
```diff
@@ -22,69 +22,78 @@ class TimeIntegratorBase(ABC):
...
         equation: Equation,
-        solution: firedrake.Function,
-        dt: float,
-        solution_old: Optional[firedrake.Function] = None,
-        solver_parameters: Optional[dict[str, Any]] = None,
-        strong_bcs: Optional[list[firedrake.DirichletBC]] = None,
-    ):
+        solution: fd.Function,
+        dt: fd.Constant | float,
+        /,
```
sghelichkhani, line 70: Why is it necessary/better to make them strictly positional?

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
Patol75: That could make the solver crash if `output_frequency` is much larger than what is allowed by the CFL criterion.

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
Patol75: It was for consistency with DQ. If you use DG, you get the following warning: `UserWarning: Discontinuous Lagrange element requested on quadrilateral, creating DQ element.`.

sghelichkhani (changes requested):
Thanks @Patol75 for this. The level set stuff definitely makes sense. My main concern at this point is that some of the changes are not strictly level-set. Specially the `time_stepper.py`. Can I suggest we discuss those in dev meets and potentially if all agree move that to a seperate PR? I am also not sure about making the arguments positional only!

gadopt/time_stepper.py
Patol75: It is neither necessary nor better. I find it helpful and more readable to have required arguments as positional and optional arguments as keyword.

gadopt/time_stepper.py
Patol75: Probably true.

Patol75:
I believe I have identified the root cause of the reproducibility issues for the thermochemical buoyancy demo. I changed the material interface for the compositional Rayleigh number from arithmetic to sharp, and that seems to create trouble. I will reverse that change, update expected diagnostics, incorporate new master changes, and remove proposed modifications not fully necessary.

Patol75:
The last commit should fix the CI. However, I will note that the thermochemical case from van Keken et al. (1997) is a pain. If I use `arithmetic`, I get a reproducible yet inaccurate result. So I put `arithmetic` in the demo and `sharp` in the test. This should be better once we can use an adaptive mesh. Currently, it is too computationally expensive to increase the resolution.

angus-g:
Thanks @Patol75, this is nice work. I am against merging this in its current form, however. You have mixed complete refactorings of the demos and tests that use this API in with the API change, which makes it really hard to evaluate how the API itself has changed. I'd advocate for again splitting this: just the API change alone, and the *minimal* changes to bring the demos and tests using the API back to compatibility. Then reintroduce the demo and test changes, ideally in separate PRs.

The smaller and more contained a set of changes is, the easier it is to evaluate and review.

Patol75:
Closing as most of the changes are now either in `master` or in other open PRs.

drhodrid:
Goodbye, my lover; goodbye, sweet PR,
Your buttons now grey, but you’ve traveled far.
Though closed in form, your soul was spared —
For all your gifts have been elsewhere shared.

You’ve merged by proxy, split but whole,
Reborn in master — bless your scroll.
No conflicts now, your work complete,
A quiet end, a merge discreet.

PR #129, opened 2024-10-31
PDE-constrained optimisation

drhodrid:
This ports across and improves two of the demos we ran at our 2023 meeting. They provide straightforward examples of setting up and running simple adjoint scenarios.


drhodrid:
This is not yet ready for a full review. However, I chatted to @dsroberts about these demos last week and figured he would be a good test case. I am generally happy with PDE_constrained_field demo. @dsroberts would you mind running this locally and providing feedback / suggestions for improvement? This will likely be the first adjoint case that people see in the 2024 workshop.

Whilst you're playing I have two requests for you (since I ran out of time):
1. Could you improve my plots? I screwed up the aspect ratios a little and smashed the colorbar into the second subplot rather than giving it it's own space.
2. At the end I want to plot the evolution of the functional with iterations. I've screwed something up here, because it doesn't plot values that are consistent with what ROL is spitting out to my screen.

demos/PDE_constrained_optimisation/PDE_constrained_field.py
```diff
@@ -54,7 +52,7 @@
 x0, y0 = 0.75, 0.5
 w = 0.1
 r2 = (x - x0) ** 2 + (y - y0) ** 2
-T0.interpolate(exp(-r2 / w**2))
+T0 = interpolate(exp(-r2 / w**2), Q)
```
dsroberts, line 55: I don't understand this change. `T0` is being initialised on line 37, then discarded and replaced by the result of the `interpolate` function on Line 55, whereas before `interpolate` was being applied to the existing `T0` object. Its fairly inconsequential here, you only lose the "Initial Temperature" name, but it may be more consequential in other situations.

demos/PDE_constrained_optimisation/PDE_constrained_field.py
```diff
@@ -34,9 +34,7 @@
 V = VectorFunctionSpace(mesh, "CG", 2)  # Function space for velocity.
 Q = FunctionSpace(mesh, "CG", 1)  # Function space for the scalar (Temperature).
 T = Function(Q, name="Temperature")
-T0 = Function(
-    Q, name="Initial_Temperature"
-)  # Initial condition for temperature which we will invert for.
+T0 = Function(Q, name="Initial_Temperature")  # T Initial condition which we will invert for.
```
dsroberts, line 37: Apologies for this needless change, my linter decided that line was too long.

demos/PDE_constrained_optimisation/PDE_constrained_field.py
```diff
@@ -117,6 +115,7 @@
 approximation = BoussinesqApproximation(Ra=1, kappa=2e-2)
 delta_t = 0.1
 energy_solver = EnergySolver(T, u, approximation, delta_t, ImplicitMidpoint)
+energy_solver.solver_parameters.pop('ksp_converged_reason')
```
dsroberts, line 118: Replace with `del energy_solver.solver_parameters["ksp_converged_reason"]` if the value to be removed isn't needed for anything. Also, why does this need to be deleted? Is it a once off for this test or is it an attribute that shouldn't be created until later?

dsroberts (commented):
Some super pedantic comments. Feel free to disregard.

demos/PDE_constrained_optimisation/PDE_constrained_field.py
drhodrid: This was me being silly. Good spot. All fixed up.

demos/PDE_constrained_optimisation/PDE_constrained_field.py
drhodrid: Done.

drhodrid:
Now happy with the first demo - need to think more about the second - it doesn't quite do what I'd expect, but I can't currently see my error. Maybe I'm just thinking about it wrongly!

dsroberts:
Looks good to me. Fixed a typo at the bottom of the notebook.

drhodrid:
OK - I think if @angus-g is happy then this can go in, with it's own mermaid!

drhodrid:
Ohhh wait - we actually also need values for testing these. I'll supply them shortly.

drhodrid:
For field case, ensure that final functional value is **< 7e-3.**
For boundary case, ensure that final functional value is **< 2e-3.**
There is also the usual warning we get from "interpolate" - could somebody remind me how to fix this, or better still, fix it for me?!

angus-g:
So this still needs to be run for testing, and for the notebooks to appear on the tutorials page?

drhodrid:
> So this still needs to be run for testing, and for the notebooks to appear on the tutorials page?

Yes - I think we have to test it, to ensure ongoing functionality.
I would also give it is own mermaid on the tutorials page (because it's general, and not really mantle convection, GIA etc...).

angus-g:
I think if CI goes properly green here, we should be good? I've added the tutorial to the sidebar on the tutorials page, but not the diagram. We could add some text on the website side of things to explain its purpose.

Note that because of the TSFC issue, this will contain a temporary pin of the Firedrake image to the one prior to current to ensure we don't get spurious testing failures.

drhodrid:
Looks good to me.

angus-g (approved):
Should be good now!

PR #131, opened 2024-11-01
Makefile cleanup

dsroberts:
Revamp makefiles. make check now runs appropriate tests wherever its invoked. Top level behaviour remains the same, make check in demos only tests demos, make check in an individual test only tests that, etc. Most make check invocations depend on test output. Some tests were modified to make this work properly. make clean now removes all artefacts from testing.

### DIFF
--- a/tests/analytical_comparisons/analytical.py
+++ b/tests/analytical_comparisons/analytical.py
@@ -73,7 +73,6 @@ cases = {
 
 
 def get_case(cases, config):
-    print(config)
     config = config.split("_", maxsplit=2)
     while config:
         cases = cases[config.pop(0)]
--- a/tests/analytical_comparisons/analytical.py
+++ b/tests/analytical_comparisons/analytical.py
@@ -2,23 +2,24 @@ import argparse
 import itertools
 import subprocess
 import sys
+import importlib
 
 cases = {
     "smooth": {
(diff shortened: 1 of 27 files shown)


tests/analytical_comparisons/analytical.py
```diff
@@ -72,7 +73,8 @@
 
 
 def get_case(cases, config):
-    config = config.split("/")
+    print(config)
```
angus-g, line 76:
```suggestion
```

Drop?

angus-g (commented):
Awesome, thanks! Should we use `make check` in https://github.com/g-adopt/g-adopt/blob/1d10b20dbbfbc1cd93b7fd51e3b27589b2c7e7bb/.github/workflows/test.yml#L113? Although it would need to learn how to do the JUnit XML output.

tests/analytical_comparisons/analytical.py
dsroberts: Yes please. Looks like I left some debugging in...

dsroberts:
I think the actions workflows are fine the way they are. I envisioned that this setup would be used for interactive testing during development or deployment to a new system. The machines don't need anything simplified for them...

angus-g (approved):
All good then!

PR #132, opened 2024-11-04
Use default solver tolerances in TALA DG

Patol75:
Fixing what I believe was forgotten in gh-124. Let us see what the CI returns.

### DIFF
--- a/tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py
+++ b/tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py
@@ -205,7 +205,6 @@ energy_solver = EnergySolver(T, u, approximation, delta_t, ImplicitMidpoint, bcs
 stokes_solver = StokesSolver(z, T, approximation, bcs=stokes_bcs,
                              nullspace=Z_nullspace, transpose_nullspace=Z_nullspace,
                              near_nullspace=Z_near_nullspace)
-stokes_solver.solver_parameters['snes_rtol'] = 1e-2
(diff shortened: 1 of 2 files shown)


angus-g:
This will likely fail. The part of the comment you missed was that it requires an adjustment to the testing tolerances.

drhodrid:
There has been some confusion here, probably due to poor communication on my side. The results that are currently being tested against used default tolerances for everything. It took over an hour on 16 cores on Erebus. What we need is to leave this on 16 cores (@angus-g and I have already agreed to this), but remove the additional rtol line. I have just done this - so assuming it goes green this can be merged.

Patol75:
The idea is that you could keep `stokes_solver.solver_parameters['snes_rtol'] = 1e-2` (anyone knows what the default value is?) and run on 16 cores if you wanted to retain a reasonable run time. That would not affect CI results, as demonstrated above (this specific case took 1 hour and 40 minutes on 4 cores).

drhodrid:
That line is not needed in this example, hence why I’ve removed it.

On 4 Nov 2024, at 1:52 pm, Thomas Duvernay ***@***.***> wrote:



The idea is that you could keep stokes_solver.solver_parameters['snes_rtol'] = 1e-2 (anyone knows what the default value is?) and run on 16 cores if you wanted to retain a reasonable run time. That would not affect CI results, as demonstrated above (this specific case took 1 hour and 40 minutes on 4 cores).

Patol75:
CI is green, the test passed in ~40 minutes.

drhodrid (approved):
Thanks @Patol75! 40 mins on 16 cores is good! Go go go!

Issue #127, opened 2024-10-29
#121 not applied to demos/adjoint_2d_cylindrical


dsroberts:
API change implemented in #121 was not applied to demos/adjoint_2d_cylindrical. Tests pass as this demo is explicitly skipped. Leaving this here as I'll address it separately to the Makefile stuff I'm working on.

w3168:
Thanks @dsroberts . 

How slow is that test? Is there a reason we are not running it?

Issue #126, opened 2024-10-29
Untested run cases


Patol75:
A number of Python source files are currently untested, e.g. in `demos/adjoint_2d_cylindrical`, `demos/scalar_advection`, `tests/adjoint`, and  `tests/scalar_advection_diffusion`. What should we do with them?

Patol75:
gh-133 took care of the adjoint cases; the advection cases remain.

PR #133, opened 2024-11-04
Cleanup of unused tests and rationalisation of adjoint testing. 

drhodrid:
On the adjoint side, we currently have:

demos/adjoint - a nice succinct demo, for recovery of an initial condition in a square domain. There is a forward script and an associated adjoint (including a Taylor test). Currently only Taylor test is tested. My view is that we should test all aspects (including minimisation up to perhaps 5 iterations) and I will update that today. 

demos/cylindrical - a longer more challenging test, including a non-linear rheology, in a cylinder - this is essentially the final synthetic test from Sia's 2024 paper. My plan is to actually have the CI run the Taylor test (from a checkpoint) and remove inverse.py - testing it within the CI is simply not tractable (many many many hours on many cores!) and since it is not tested, it ends up not compatible with master either way. 

tests/adjoint - this is quite similar to demos/adjoint, albeit with a more complex functional. Taylor tests are calculated for all 4 terms in the functional, which is currently part of the CI. We will keep this, but again remove inverse.py, because it is untested and has already become incompatible with master. 

Note that both inverse scripts that I removed are archived on zenodo - associated with Ghelichkhan et al. 2024 - so that we can recover the versions used in that paper easily. 

This will address https://github.com/g-adopt/g-adopt/issues/127 and most of https://github.com/g-adopt/g-adopt/issues/126.


demos/adjoint_2d_cylindrical/forward.py
```diff
@@ -136,14 +119,18 @@ def step_func(centre, mag, increasing=True, sharpness=50):
...
-        solver_parameters=newton_stokes_solver_parameters,
+        near_nullspace=Z_near_nullspace
     )
 
+    # Overwrite detault solver parameters to use a direct solver for Stokes system.
+    stokes_solver.solver_parameters["mat_type"] = "aij"
+    stokes_solver.solver_parameters["ksp_type"] = "preonly"
+    stokes_solver.solver_parameters["pc_type"] = "lu"
+    stokes_solver.solver_parameters["pc_factor_mat_solver_type"] = "mumps"
+
```
Patol75, line 133: In this case, you should use `solver_parameters="direct"` in the call to `StokesSolver`.

demos/adjoint_2d_cylindrical/taylor_test.py
```diff
@@ -161,14 +144,18 @@ def step_func(centre, mag, increasing=True, sharpness=50):
...
-        near_nullspace=Z_near_nullspace,
-        solver_parameters=newton_stokes_solver_parameters,
+        near_nullspace=Z_near_nullspace
     )
 
+    # Overwrite detault solver parameters to use a direct solver for Stokes system.
+    stokes_solver.solver_parameters["mat_type"] = "aij"
+    stokes_solver.solver_parameters["ksp_type"] = "preonly"
+    stokes_solver.solver_parameters["pc_type"] = "lu"
+    stokes_solver.solver_parameters["pc_factor_mat_solver_type"] = "mumps"
```
Patol75, line 157: Same here.

drhodrid:
OK - I have just pushed a change to adjoint.py based on advice from @sghelichkhani, that I think nicely fixes the issues identified in https://github.com/g-adopt/g-adopt/issues/134. 

@angus-g - we are now in a position to test adjoint.py too in the folder demos/adjoint. It runs for 10 timesteps, for 5 optimisation iterations. What I would check is the final value for the functional after these 5 iterations. It should be < 1e-4. Could you add this to the testing/makefiles and I will try to box off the outstanding 2d_cylindrical_adjoint issues next?

demos/adjoint/adjoint.py
```diff
@@ -163,19 +166,20 @@
...
+# guess is set to the 1-D average of the forward model:
+Tic = Function(Q1, name="Initial_Condition_Temperature").assign(Taverage)
 
-# Project the temperature field from the reference simulation's final time-step onto the control space as our
-# initial guess:
-with CheckpointFile(checkpoint_filename, mode="r") as fi:
-    Tic.project(fi.load_function(mesh, "Temperature", idx=timesteps - 1))
-# -
+# Given that Tic will be updated during the optimisation, we also create a function to store our initial guess,
+# which we will later use for smoothing. Note that sinces smoothing is executed in the control space, we must
```
Patol75, line 174: `since`

drhodrid:
@angus-g @dsroberts - does this adjoint demo now also need a "make check" for consistency?

drhodrid:
Hew file for web download (outcome of forward model) available here: please thumbs up when this is uploaded -- https://www.dropbox.com/scl/fi/8chlsgy2zpkhuuhlwvk8y/adjoint-demo-checkpoint-state.h5?rlkey=5vdzyzw1oevysq7l5pritq8w4

drhodrid:
Ok - I think I've reached as far as I can reach here. To update:

demos/adjoint - a nice succinct demo, updated, with bugs fixed. Now tested and working well. 

tests/adjoint - this is quite similar to demos/adjoint, albeit with a more complex functional. Taylor tests are now calculated for all 4 terms in the functional, and they work well. I have removed inverse.py, because it is untested and has already become incompatible with master. There maybe an outstanding issue with how this test is set up, which I have pinged @dsroberts and @angus-g  on in chat. 

demos/adjoint_2d_cylindrical - I have tidied up these files somewhat, but when we added the Taylor test to the CI, it failed. I've tried to figure out what is going on here, but currently can't find a solution. My suggestion is that I raise this as an issue, which we can fix post workshop. For now, we should incorporate the changes I have made to the tests, but remove it from our CI. @angus-g - are you happy to make this final change, removing it from CI?

drhodrid:
Issue raised here: https://github.com/g-adopt/g-adopt/issues/136

angus-g:
Is this ready to go now?

drhodrid:
I think it is!

drhodrid:
I can't approve, but it looks good to me!

angus-g (approved):

PR #137, opened 2024-11-08
Viscoelastic adjoint ice demo

w3168:
Adding a demo ice inversion based on the 2d cylindrical setup. I've made the timesteps longer (from 250 years to 1000 years) to try and speed up the demo. 

I need to add the save checkpoint file to the server


demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
```diff
@@ -0,0 +1,717 @@
...
+# To bring in G-ADOPT's adjoint functionality we need to start taping the forward problem, which we do below. It's also good practice to clear the tape, so that we are starting fresh each time.
+
+from gadopt.inverse import *
+tape = get_working_tape()
+tape.clear_tape()
+
+# In this tutorial we are going load the mesh created by the forward cylindrical demo in the previous tutorial. This makes it easier to load the synthetic data from the previous tutorial for our 'twin' experiment.
+
+# Set up geometry:
+checkpoint_file = "./forward-2d-cylindrical-disp-incdisp-dt1ka.h5"
```
angus-g, line 32: i.e. the `glacial_isostatic_adjustment/2d_cylindrical`? It would be clearer if this filename matched what that one outputs. Then we can make sure all the dependency ordering is in place (see the other adjoint demo). And then we can also pop the checkpoint on gadopt.org.

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
```diff
@@ -0,0 +1,717 @@
...
+u, p = split(z)  # Returns symbolic UFL expression for u and p
+z.subfunctions[0].rename("Incremental Displacement")
+z.subfunctions[1].rename("Pressure")
+
+displacement = Function(V, name="displacement").assign(0)
+stress_old = Function(S, name="stress_old").assign(0)
+# -
+
+# Let's set up the background profiles for the material properties with the same values as before.
+
```
angus-g, line 60:
```suggestion
```

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
```diff
@@ -0,0 +1,717 @@
...
+adj_ice_file = VTKFile("adj_ice.pvd")
+converter = RieszL2BoundaryRepresentation(W, top_id)  # convert to surface L2 representation
+tape.add_block(DiagnosticBlock(adj_ice_file, normalised_ice_thickness, riesz_options={'riesz_representation': converter}))
+# -
+
+
+# Let's visualise the ice thickness using pyvista, by plotting a ring outside our synthetic Earth.
+
+# +
```
angus-g, line 175:
```suggestion
# + tags=["active-ipynb"]
```

(And comment the rest of the block)

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
```diff
@@ -0,0 +1,717 @@
...
+
+    slab_x, slab_y = 3e6, 4.5e6
+    slab = bivariate_gaussian(X[0], X[1], slab_x, slab_y, 0.7e6, 0.35e6, 0.7)
+    heterogenous_viscosity_field.interpolate(-1*slab + heterogenous_viscosity_field * (1-slab))
+
+    high_viscosity_craton_x, high_viscosity_craton_y = 0, 6.2e6
+    high_viscosity_craton = bivariate_gaussian(X[0], X[1], high_viscosity_craton_x, high_viscosity_craton_y, 1.5e6, 0.5e6, 0.2)
+    heterogenous_viscosity_field.interpolate(-1*high_viscosity_craton + heterogenous_viscosity_field * (1-high_viscosity_craton))
+
+    return heterogenous_viscosity_field
```
angus-g, line 126: I wonder if it'd be cleaner to just load these directly from the other demo (if they are indeed all the same), but then it's a decision of how to do the import from a sibling directory... `sys.path` modification? Make everything in `demos` part of the `gadopt` package so it can get a fully-qualified import?

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
```diff
@@ -0,0 +1,717 @@
...
+
+normalised_ice_thickness = Function(W, name="normalised ice thickness")
+
+
+control = Control(normalised_ice_thickness)
+ice_load = rho_ice * g * Hice1 * normalised_ice_thickness
+
+adj_ice_file = VTKFile("adj_ice.pvd")
+converter = RieszL2BoundaryRepresentation(W, top_id)  # convert to surface L2 representation
+tape.add_block(DiagnosticBlock(adj_ice_file, normalised_ice_thickness, riesz_options={'riesz_representation': converter}))
```
angus-g, line 169: Does this need explaining? I'm certainly curious!

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
```diff
@@ -0,0 +1,717 @@
...
+log(f"Simulation start time: {Tstart} years")
+# -
+
+# Similar to before, we setup the boundary conditions, this time using the normalised ice thickness to account for ice covered regions when calculating the density contrast across the free surface.
+
+# Setup boundary conditions
+exterior_density = rho_ice * normalised_ice_thickness
+stokes_bcs = {top_id: {'normal_stress': ice_load, 'free_surface': {'delta_rho_fs': density - exterior_density}},
+              bottom_id: {'un': 0}
+              }
```
angus-g, line 300:
```suggestion
stokes_bcs = {
    top_id: {
        'normal_stress': ice_load,
        'free_surface': {'delta_rho_fs': density - exterior_density}
    },
    bottom_id: {'un': 0}
}
```

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
```diff
@@ -0,0 +1,717 @@
...
+    plog.log_str(
+        f"{timestep} {float(time)} {float(dt)} "
+        f"{gd.u_rms()} {gd.u_rms_top()} {gd.ux_max(top_id)} "
+        f"{vertical_displacement.dat.data.min()} {vertical_displacement.dat.data.max()}"
+    )
+# -
+
+# As we can see from the plot below there is no displacement at the final time given there is no ice load!
+
+# +
```
angus-g, line 399:
```suggestion
# + tags=["active-ipynb"]
```

(Plus comment the rest)

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
```diff
@@ -0,0 +1,717 @@
...
+            [0, 0, 0, 1],
+        ])
+
+    transformed_arc_data = arc_data.transform(transform_matrix)
+    p.add_mesh(transformed_arc_data, line_width=8, scalar_bar_args=scalar_bar_args, clim=[-5e-6, 5e-6], cmap=adj_cmap)
+
+
+# Next we read in the file that was written out as part of the diagnostic callback added to the tape earlier. We can see there is a clear hemispherical pattern in the gradients. Red indicates that increasing the ice thickness here would increase out objective function and blue areas indicates that increasing the ice thickness here would decrease our objective function. In the 'southern' hemisphere where we have the biggest ice load the gradient is negative, which makes sense as we expect increasing the ice thickness here to reduce our surface misfit.
+
+# +
```
angus-g, line 564:
```suggestion
# + tags=["active-ipynb"]
```

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
```diff
@@ -0,0 +1,717 @@
...
+optimiser.run()
+
+# If we're performing mulitple successive optimisations, we want
+# to ensure the annotations are switched back on for the next code
+# to use them
+continue_annotation()
+
+# -
```
angus-g, line 650:
```suggestion
# Now let's run the inversion!

optimiser.run()

# If we're performing mulitple successive optimisations, we want
# to ensure the annotations are switched back on for the next code
# to use them

continue_annotation()
```

Make the continue annotation step a bit more explicit with its own markdown cell

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
```diff
@@ -0,0 +1,717 @@
...
+
+# Let's plot the results of the inversion at the final iteration.
+
+# +
```
angus-g, line 654:
```suggestion
# + tags=["active-ipynb"]
```

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
```diff
@@ -0,0 +1,717 @@
...
+plotter.show(jupyter_backend="static", interactive=False)
+# Closes and finalizes movie
+plotter.close()
+# -
+
+# We can see that we have been able to recover two ice sheets in the correct locations and the final displacement pattern is very similar to the target run. We can confirm that the surface misfit has reduced by plotting the objective function at each iteration.
+
+plt.plot(J_list, '-x')
+plt.xlabel("Iteration number")
+plt.ylabel("Objective function")
```
angus-g, line 717:
```suggestion
# + tags=["active-ipynb"]
# plt.plot(J_list, '-x')
# plt.xlabel("Iteration number")
# plt.ylabel("Objective function")
# -
```

gadopt/inverse.py
```diff
@@ -359,3 +360,54 @@ def __getstate__(self):
         "Iteration Limit": 100,
     },
 }
+
+
+class DiagnosticBlock(Block):
```
angus-g, line 365: Docstring: What/why is this block?

gadopt/inverse.py
```diff
@@ -359,3 +360,54 @@ def __getstate__(self):
...
+        return block_variable.checkpoint
+
+    def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx, prepared=None):
+        out = inputs[0]._ad_convert_type(adj_inputs[0], options=self.riesz_options)
+        out.rename('adjoint_'+self.f_name)
+        self.f.write(out)
+        return 0
+
+
+class DiagnosticConstantBlock(Block):
```
angus-g, line 384: Docstring: What/why is this block?

gadopt/inverse.py
```diff
@@ -359,3 +360,54 @@ def __getstate__(self):
...
+        return block_variable.checkpoint
+
+    def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx, prepared=None):
+        out = inputs[0]._ad_convert_type(adj_inputs[0])
+        print("Adjoint ", self.name, out.values()[0])
+        return 0
+
+
+class RieszL2BoundaryRepresentation:
+    """Callable that Converts l2-representatives to L2-boundary representatives"""
```
angus-g, line 401: Would be nice to have a bit more information about this...

angus-g (commented):
Nice, few notebook-y tweaks and documentation. We'll obviously need to set up all the makefiles and dependencies to have this run after the forward, and then a pre-run checkpoint file can be provided online.

Haven't gone through spelling, but would recommend line-wrapping all the really long comments for markdown blocks to help when editing as code.

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
w3168: So it's not quite the same output! I tried to speed up the adjoint demo by taking fewer timesteps (dt = 1000 years not 250 years) so the forward run and henche the checkpoint is a bit different

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
w3168: It's to do with visualing a surface gradient. We used to have it in thwaites. The point is to do the L2 correction using a surface integral instead of the usual volume integral for the mass matrix. this prevents spurious oscillations in cells not connected to the boundary where the gradient should be zero.

gadopt/inverse.py
w3168: I've added some docstrings. It was from thwaites basicailly a helper block to allow you to output gradients wrt to function provided. It was useful for time dependent simulations (and also for visualising gradient at different iteration of optimisation alogrithm)

gadopt/inverse.py
w3168: added a bit more based on similar comments above

gadopt/inverse.py
w3168: deleted this as dont need it for this PR. I think it would be neater to combine function/constant together to save redundant code anyway so i guess we can fix this when we have a need for it!

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
w3168: This is changed now so it is the same as the forward run. (i made the forward run take longer timesteps)

demos/glacial_isostatic_adjustment/gia_adjoint_ice/gia_adjoint_ice.py
w3168: I was thinking about this again... at the moment the cases are named 2d_cylindrical.py etc following the mantle convection cases so I get an invalid decimal literal error for the name import...  I guess we could rename everything cylindrical_2d.py

drhodrid:
Hi @w3168,

Fantastic work on this!

I’ve carefully reviewed the tutorial this afternoon, considering its use in sessions where users are primarily interested in GIA. Given that they are likely to engage with base_case, 2D_cylindrical, and this tutorial -- but not other demos/tutorials from the mantle convection space -- this will likely be their first exposure to inverse/adjoint problems in G-ADOPT. With that in mind, I have a few suggestions to enhance clarity and accessibility:

1. Expand the introduction to adjoint functionality. Since this may be the user’s first encounter with adjoint methods, a brief explanation of key concepts -- such as tape, reduced functional, control, etc. -- would be beneficial. In the first mantle convection demo, we took the time to spell these out clearly; I’d suggest incorporating some of that material here to support users who are working through the tutorial independently.
2. Provide a more detailed problem introduction. The tutorial would benefit from a clearer upfront explanation of the problem being solved. For example, explicitly state that viscosity is assumed to be known, but ice thickness is being inverted for, and provide some context on why this is important. A stronger introduction will help guide users more effectively.
3. Streamline the presentation of helper functions. The tutorial includes numerous helper functions for plotting and other utilities, which take up significant space and can make the notebook appear more complex than it actually is. Would it be possible to hide these from the main flow of the tutorial? Perhaps a dedicated plotting_utility module or similar would help keep things concise while still making these functions accessible?

I’ve also added a few minor comments directly to the text and committed and will comment on smaller details that stood out shortly. Looking forward to seeing how this develops!

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
```diff
@@ -0,0 +1,838 @@
...
+# a surface integral, so that our output gradient accounts for the surface
+# area of each cell.
+converter = RieszL2BoundaryRepresentation(W, top_id)  # Convert to surface L2 representation
+
+# We add a diagnostic block to the tape which will output the gradient
+# field every time the tape is replayed.
+tape.add_block(DiagnosticBlock(adj_ice_file, normalised_ice_thickness, riesz_options={'riesz_representation': converter}))
+# -
+
+
```
drhodrid, line 194: When running via jupyter, "13" is printed here for some reason. Not sure why.

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
```diff
@@ -0,0 +1,838 @@
...
+target_normalised_ice_thickness = Function(W, name="target normalised ice thickness")
+target_normalised_ice_thickness.interpolate(disc1 + (Hice2/Hice1)*disc2)
+
+normalised_ice_thickness = Function(W, name="normalised ice thickness")
+
+control = Control(normalised_ice_thickness)
+ice_load = rho_ice * g * Hice1 * normalised_ice_thickness
+
+adj_ice_file = VTKFile("adj_ice.pvd")
```
drhodrid, line 179: Whis is this loaded here? Explanation needed?

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
```diff
@@ -0,0 +1,838 @@
...
+# Since we are calculating the sensitivity to a field that is only defined
```
drhodrid, line 180: This Riesz boundary stuff needs more explanation and should be in its own notebook block for clarify rather than having the comment embedded.

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
```diff
@@ -0,0 +1,838 @@
...
+#
+# Now let's run the inversion!
+
+optimiser.run()
+
+# If we're performing mulitple successive optimisations, we want
+# to ensure the annotations are switched back on for the next code
+# to use them
+
+continue_annotation()
```
drhodrid, line 775: There is a "True" printed after this for some reason, which I found a little confusing.

demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -374,14 +371,14 @@ def setup_heterogenous_viscosity(viscosity):
 # plotter.close()
 # -
 
-# Let's setup the timestepping parameters with a timestep of 250 years and an output frequency of 1000 years.
+# Let's setup the timestepping parameters with a timestep and output frequency of 1000 years and an output frequency.
```
sghelichkhani, line 374:
```suggestion
# Let's set up timestepping parameters with:
# - Timestep: 1000 years
# - Output frequency: Every 1000 years
# - Total simulation duration: 10,000 years
```

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
```diff
@@ -0,0 +1,838 @@
+# Synthetic ice inversion using adjoints
+# =======================================================
+#
+# In this tutorial, we will use G-ADOPT's adjoint capability to invert for a synthetic ice
+# load in an annulus domain. We will be running a 'twin' experiment where we will try to
+# recover the ice load that we used as part of the earlier 2d cylindrical tutorial,
```
sghelichkhani, line 6: Not sure if this is easy, but can we provide a link to it?

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
```diff
@@ -0,0 +1,838 @@
...
+# Layer properties from Spada et al. (2011)
+radius_values = [6371e3, 6301e3, 5951e3, 5701e3, 3480e3]
+density_values = [3037, 3438, 3871, 4978]
+shear_modulus_values = [0.50605e11, 0.70363e11, 1.05490e11, 2.28340e11]
+viscosity_values = [2, -2, -2, -1.698970004]  # viscosity = 1e23 * 10**viscosity_values
+# N.b. that we have modified the viscosity of the lithosphere from
+# Spada et al. (2011) because we are using coarse grid resolution
+
+
+def initialise_background_field(field, background_values, vertical_tanh_width=40e3):
```
sghelichkhani, line 84: Would it be possible to import all these functions from the forward demo? It might not be as easy, but that would get rid of lots of duplications.

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
sghelichkhani: I am not sure even I understsand the purpose of these bits. I don't understand what *the projection* is. Also, if I understand right, this is only for sake of your diagnostics, if I am right. Can we explain that? Also the idea of the `DianosticBlock` is nice. But for the purpose of this tutorial, wouldn't it be easier for a lame user to just use the callback function to the functional? 

I feel these bits need a lot more explanation for a general user.

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
```diff
@@ -0,0 +1,838 @@
...
+
+adj_ice_file = VTKFile("adj_ice.pvd")
+# Since we are calculating the sensitivity to a field that is only defined
+# on the top boundary, if we do the usual L2 projection (using the
+# mass matrix) to account for the size of the mesh element then we
+# will get spurious oscillating values in the output gradient in
+# cells not connected to the boundary. Instead we do the projection using
+# a surface integral, so that our output gradient accounts for the surface
+# area of each cell.
+converter = RieszL2BoundaryRepresentation(W, top_id)  # Convert to surface L2 representation
```
sghelichkhani, line 187: So based on what you are saying here, you should be able to use a project with an InteriorBC of 0.0 to set the function zero inside the domain, right?

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
```diff
@@ -0,0 +1,838 @@
...
+
+with CheckpointFile(checkpoint_file, 'r') as afile:
+    final_target_incremental_displacement = afile.load_function(mesh, name="Incremental Displacement", idx=9)
+    final_target_displacement = afile.load_function(mesh, name="Displacement", idx=9)
+
+final_target_velocity = Function(V, name="target velocity").interpolate(final_target_incremental_displacement / dt_years)
+functional_values = []
+
+
+def eval_cb(J, m):
```
sghelichkhani, line 565: It might be better to change `eval_cb` into a class, and rewrite this as its __call__ method.

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
```diff
@@ -0,0 +1,838 @@
...
+
+minimisation_parameters["Status Test"]["Iteration Limit"] = 15
+minimisation_parameters["Step"]["Trust Region"]["Initial Radius"] = 1e4
+
+optimiser = LinMoreOptimiser(
+    minimisation_problem,
+    minimisation_parameters,
+    checkpoint_dir="optimisation_checkpoint",
+)
+# Restart file for optimisation...
```
sghelichkhani, line 758: I am a bit lost why you have functional_values here ...

sghelichkhani: I think I don't understand these VTKFiles either!

gadopt/inverse.py
sghelichkhani: This is very nice! But I am thinking it's beyond this PR. I would like to see a separate PR with this and a small demo+test for it. Can you not do this PR without `DiagnosticBlock`? Why not using StatusTest for ROL, or even a call back function to functional? Also the fact that this is coming from Thwaites, suggests this might a pyadjoint change?

gadopt/inverse.py
```diff
@@ -361,3 +362,64 @@ def __getstate__(self):
...
+        return block_variable.checkpoint
+
+    def evaluate_adj_component(self, inputs, adj_inputs, block_variable, idx, prepared=None):
+        out = inputs[0]._ad_convert_type(adj_inputs[0], options=self.riesz_options)
+        out.rename('adjoint_'+self.f_name)
+        self.f.write(out)
+        return 0
+
+
+class RieszL2BoundaryRepresentation:
```
sghelichkhani, line 403: Not sure this is really necessary. This is bascially a `project` with an InteriorBC of 0.

sghelichkhani (changes requested):
Good job @w3168! General comments from me:
    - the DiagnosticBlock is a great idea. But it needs to be separated from this, with its proper example demo! Also we might be able to make it more general and included in pyadjoint?
    - I believe your riesz boundary stuff can be done via a project. 
    - the adjoint ice demo is very bulky! Can it be made shorter to just cover the essential? Could also the plottings be shortened to the essentials? I am just thinking that a user would not come here to have the perfect visualisation, but to learn how things work.

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
w3168: Thanks as we discussed offline this does seem to help. i.e. gradient in L2 norm is confined to boundary and seems to have right weight to convert back to l2 when integrating over the cell.  Hopefully more consistent  with ROL

demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
w3168: Sorry this was just a hack so only outputs associated with the optimisation were stored in the pvd

gadopt/inverse.py
w3168: as discussed offline I'll remove this

PR #140, opened 2024-11-11
Make triggering a documentation rebuild optional

angus-g:
Sometimes we want to run a documentation render without updating the website, to check that it would complete. To support this, we add a "trigger" input to the workflow dispatch for this action.


Patol75 (approved):
Useful and straightforward, thank you.

PR #142, opened 2024-11-12
Upload all run logs for tests

angus-g:
I had originally made the runlogs upload when the tests step fails, but for some reason this was always temperamental. Let's instead just upload an artifact containing all the runlogs, regardless of test status. This should make it much easier to dig into failing tests.

An example of this is when I unpinned the Firedrake image leading to TSFC errors: https://github.com/g-adopt/g-adopt/actions/runs/11789097245

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -130,12 +130,22 @@ jobs:
 
           digest="$(tsp)"
 
+          mkdir test_outputs
+
           # extract info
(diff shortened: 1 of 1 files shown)


Patol75:
I do not necessarily understand the YML changes, but the example generated artifact is indeed helpful. Thank you!

dsroberts (approved):
Does what it says on the tin.

PR #147, opened 2024-11-13
Serialise startup of render jobs

### LINKED ISSUE #143: Occasional failures while rendering notebooks
I think there's some kind of resource contention while rendering notebooks. The actual rendering works just fine, but ZMQ instances run into each other.

See: https://github.com/g-adopt/g-adopt/actions/runs/11806483870/attempts/1

angus-g:
jupytext calls nbconvert to use its ExecutePreprocessor which starts a NotebookClient which starts a KernelManager through a LocalProvisioner. Of course, KernelManager includes a ConnectionFileMixin and tries to bind 5 random ports when it provisions the manager, then writes this to a file for when the Client actually starts. This is a big race condition, and means that the automatic retry code in the Client for random ports never gets hit, because the port is already specified in the connection file.

The solution herein isn't foolproof, but barring a big hang in the initialisation of jupyter it should work. We serialise just the startup of the render jobs through a lockfile which gives each process 1 second to start before releasing the lock.

Closes #143.

### DIFF
--- a/demos/Makefile
+++ b/demos/Makefile
@@ -36,5 +36,9 @@ check:
 	python3 -m pytest -m demo
 
 # pattern rule for executing demo scripts as a notebook
+# because jupyter is an insane jumble of components, there's absolutely no way to configure it
+# to try to bind to its control port when it needs it. instead, the provisioner checks for a free
+# port and writes that to a connection file. this is of course completely prone to race conditions.
+# instead of trying to fix it at the jupyter level, we can serialise the startup with a semaphore
 %.ipynb: %.py
-	python3 -m jupytext --to ipynb --execute $< --run-path $(dir $(abspath $<))
+	(flock 9; sleep 1 && flock -u 9 & python3 -m jupytext --to ipynb --execute $< --run-path $(dir $(abspath $<))) 9>render.lock


dsroberts:
Let me think about this a bit. Ideally we block until we're certain the port(s) each ipython kernel has requested are in use. We can use network utilities like `ss` to query open ports and which process owns them, we can search for sibling processes by inspecting ppids. The machinery exists to do it this way (arguably 'The Right Way<sup>TM</sup>'), but whether it's worth the effort I don't know.

angus-g:
I considered something like that. It's a shame cgroups can't lock processes into port ranges to just isolate them. I think SELinux can but that's a step (or two) too far!

dsroberts:
The potential problem with doing it that way would be that depending on how dumb the port selection algorithm is it may just guess 5 ports outside of its assigned range and give up. If it were me, and this was my project, I'd do it the way I suggested. But assuming there is less than 1 second between selecting a port and binding to it, this is a fine solution. If this is running on Erebus, there might be a situation in which 2 jobs are running at the same time on the same machine, so you may need to use an absolute path to the lock file in order to mitigate that particular situation.

angus-g:
The runners are isolated virtual machines, so I think it's safe to assume that runners won't collide in that way.

dsroberts (approved):
LGTM

angus-g:
Thanks. I agree that it would be better to do it "properly" but everything Jupyter-related feels like a dead end and this should at least get us through the fairly regular rendering failures. Definitely as more demos come up this becomes an increasingly worse solution!

PR #148, opened 2024-11-15
Run render jobs with namespaced networking

angus-g:
This is possibly a less hacky alternative to #147 for ensuring all the jupyter kernels don't run into each other when they try to bind ports on startup.

There are plusses and minuses to this approach. The environment needs to be more carefully set up so that the namespace works, including setting subuid and subgid. I think it's probably also the case that the render jobs won't have external network access without setting up a veth device in the namespace. Maybe that's not the worst thing -- we get a fairly robust guarantee that we don't have any network dependencies.

Actions run: https://github.com/g-adopt/g-adopt/actions/runs/11848173592

### DIFF
--- a/.github/workflows/render.yml
+++ b/.github/workflows/render.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: self-hosted
     container:
       image: ghcr.io/g-adopt/firedrake:2024-11-05
-      options: --shm-size 2g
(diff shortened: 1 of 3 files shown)


dsroberts (approved):
This is great. The perfect work-around to a problem that shouldn't exist.

PR #149, opened 2024-11-15
Detect boundary ids based on info from the mesh

dsroberts:
Use the underlying PETSc plex object (if it exists) to detect Firedrake's boundary subdomain ids and return them in a namespace. The actual `get_boundary_ids` function is in `utility.py`, the rest of the changes are updating demos/tests to use it. Not ready for merge yet, just want to see if it passes testing, I've only run the demos so far.


demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -348,13 +348,13 @@ def initialise_background_field(field, background_values, vertical_tanh_width=20
 # Setup boundary conditions
 exterior_density = conditional(time < t2_load, rho_ice*disc, 0)
 stokes_bcs = {
-    bottom_id: {'uy': 0},
-    top_id: {'normal_stress': ice_load, 'free_surface': {'delta_rho_fs': density - exterior_density}},
+    boundary.bottom: {'uy': 0},
+    boundary.top: {'normal_stress': ice_load, 'free_surface': {'delta_rho_fs': density - exterior_density}},
     1: {'ux': 0},
     2: {'ux': 0},
```
angus-g, line 354: Left and right here?

gadopt/utility.py
```diff
@@ -515,3 +516,50 @@ def interpolate_1d_profile(function: Function, one_d_filename: str):
...
+            kwargs = {
+                "left": 1,
+                "right": 2,
+                "front": 3,
+                "back": 4,
+                "bottom": "bottom",
+                "top": "top",
+            }
+        else:
+            kwargs = {j: i + 1 if i + 1 in final_ids else j for i, j in enumerate(fields[:2*dim])}  # noqa: E225,E226
```
angus-g, line 558: Nit: I think `j` should be a string-implying variable name just so this reads a bit more easily.

demos/PDE_constrained_optimisation/PDE_constrained_field.py
```diff
@@ -30,7 +30,7 @@
 # +
 mesh = UnitSquareMesh(40, 40)
 mesh.cartesian = True
-left, right, bottom, top = 1, 2, 3, 4  # Boundary IDs.
+boundary = get_boundary_ids(mesh)
```
Patol75, line 33: Let's remove that one; it is not used.

gadopt/utility.py
```diff
@@ -515,3 +516,50 @@ def interpolate_1d_profile(function: Function, one_d_filename: str):
     averager = LayerAveraging(mesh, rshl if mesh.layers is None else None)
     interpolated_visc = np.interp(averager.get_layer_average(rad), rshl, one_d_data)
     averager.extrapolate_layer_average(function, interpolated_visc)
+
+
+def get_boundary_ids(mesh) -> SimpleNamespace:
```
angus-g, line 521: Worthwhile to also unit test this against all the (feasible) Firedrake meshes? What if we give this e.g. a gmsh mesh?

gadopt/utility.py
```diff
@@ -515,3 +516,50 @@ def interpolate_1d_profile(function: Function, one_d_filename: str):
...
+                "back": 4,
+                "bottom": "bottom",
+                "top": "top",
+            }
+        else:
+            kwargs = {j: i + 1 if i + 1 in final_ids else j for i, j in enumerate(fields[:2*dim])}  # noqa: E225,E226
+    # A Firedrake extruded mesh does not set boundary IDs - when this happens
+    # the boundary subdomain becomes property of the CombinedSurfaceMeasure
+    # constructed in this module, and therefore only contains "top" and
+    # "bottom" subdomains
```
angus-g, line 562: I'm not sure I understand this comment, because you are handling extruded meshes already? E.g. `mantle_convection/3d_cartesian` is an extruded `RectangleMesh`

angus-g (commented):
Awesome, this is a great idea.

gadopt/utility.py
```diff
@@ -515,3 +516,50 @@ def interpolate_1d_profile(function: Function, one_d_filename: str):
...
+        # Recover the boundary ids Firedrake has inserted
+        mesh.topology_dm.markBoundaryFaces("TEMP_LABEL")
+        if mesh.topology_dm.getStratumSize("TEMP_LABEL", 1) > 0:
+            ids = {
+                mesh.topology_dm.getLabelValue("Face Sets", i)
+                for i in mesh.topology_dm.getStratumIS("TEMP_LABEL", 1).getIndices()
+            }
+        else:
+            ids = set()
+        # Not every MPI rank has every boundary of the mesh, so we to get the set of all boundaries
```
Patol75, line 539: Extra `to` or missing verb

Patol75 (commented):
Thanks, Dale. This is a very neat and useful addition. I have had a first peek, and everything looks reasonable to me. I have not yet delved into Firedrake's internals, but let's see first what CI returns.

demos/glacial_isostatic_adjustment/base_case/base_case.py
dsroberts: Missed that one, thanks!

gadopt/utility.py
dsroberts: Yes, I managed to trip myself up on this during testing. Happy to change to `for j,i in enumerate...`

gadopt/utility.py
dsroberts: I thought about that but don't really know the best way to go about doing that. Firedrake can make a lot of different meshes.

gadopt/utility.py
angus-g: Was thinking j -> bnd or something like that

gadopt/utility.py
dsroberts: Yes, I think I wrote this comment before seeing the `3d_cartesian` case. Will come back to it next week.

gadopt/utility.py
dsroberts: I words good sometimes

drhodrid:
This is a nice feature @dsroberts - great to see. One question - how does this work for a gmsh mesh? Does it simply provide a list of "physical ID's" specified on that mesh?

dsroberts:
For gmsh meshes, its actually PETSc assigning labels to faces. The [loop starting here](https://github.com/firedrakeproject/petsc/blob/ebc0e51e771e8e066d96ec03bdd660db573f111d/src/dm/impls/plex/plexgmsh.c#L1793) does that. I'm not exactly sure how it decides which surface gets which label value though. My guess is that given Firedrake is also using the 'Face Sets' label when tagging its utility meshes, that they reverse-engineered this behaviour from PETSc. By default I don't think PETSc uses the physical labels provided in a gmsh file unless explicitly told to, but none of our examples seem to use those, so we'll probably have to cross that bridge when we get to it.

drhodrid:
I'm going to add a gmsh test @dsroberts - so that we can at least get this validated there.

dsroberts:
> I'm going to add a gmsh test @dsroberts - so that we can at least get this validated there.

FWIW the [unit tests](https://github.com/g-adopt/g-adopt/blob/d6e92fc0423288fae646b2c08dc0c716f963921e/tests/unit/test_boundary_detection.py#L149-L163) do cover both gmsh cases in `tests/multi_material`. That being said, more testing can't hurt so once #150 goes into master I'll rebase and add it to the tests.

Patol75 (commented):
This is in great shape! Two minor syntax suggestions from my side, but apart from these, the PR looks good to me. Is there anything more to add before merging?

gadopt/utility.py
```diff
@@ -501,3 +502,85 @@ def interpolate_1d_profile(function: Function, one_d_filename: str):
...
+                            abs(face_coords[idim + plex_dim * i] - bounding_box[idim][0]) < boundary_tol[idim]
+                            for i in range(nvert)
+                        ]
+                    ):
+                        tmp_bdys[2 * idim] = axis_extremes_order[idim][0]
+                    if all(
+                        [
+                            abs(face_coords[idim + plex_dim * i] - bounding_box[idim][1]) < boundary_tol[idim]
+                            for i in range(nvert)
+                        ]
```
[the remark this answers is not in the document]
dsroberts: Yes, of course. Good spot. This is actually a little optimisation as `all` will stop evaluating the internal expressions as soon as one of them is false, whereas creating a list will necessarily evaluate all of the internal expressions first.

gadopt/utility.py
```diff
@@ -501,3 +502,85 @@ def interpolate_1d_profile(function: Function, one_d_filename: str):
...
+        for face_id in [face_id for proc_bdy in gathered_boundaries for face_id in proc_bdy]:
+            # If there is any disagreement about which label belongs to which boundary,
+            # the value found on the lowest rank of the mesh communicator is selected here
+            kwargs[
+                [
+                    proc_bdy[face_id]
+                    for proc_bdy in gathered_boundaries
+                    if face_id in proc_bdy and proc_bdy[face_id] is not None
+                ][0]
+            ] = face_id
```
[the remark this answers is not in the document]
dsroberts: Agreed, that bit of code is particularly ugly. If I can convince myself what you have there does the same thing, it can go in.

Patol75 (approved):

PR #150, opened 2024-12-02
Add gmsh version of base case.

drhodrid:
Does what it says on the tin. Good to have some gmsh in our test suite. Needs a makefile (gmsh -2 square.geo) and an expected values file. I get the following values when ran locally at the end of my params.log file:
`1950 0.4222995091455919 0.00021684500894418403 9.961585719954078e-10 42.86495224800331 45.408950235909465 61.62659524301669 4.911817173812056 4.912931764076252 0.0011145902641960248 0.5000000284187623`


Patol75:
I do not think we need this test case. We already have quite a few tested multi-material cases that use Gmsh.

dsroberts:
@Patol75 this is really a test for #149. @drhodrid has set the boundary labels at random, `get_boundary_ids` needs to correctly pick up on that before its good to be merged.

drhodrid:
> I do not think we need this test case. We already have quite a few tested multi-material cases that use Gmsh.

In addition to Dale's response, which type of boundary conditions are tested in these multi-material cases? Are strong dirichlet BCs tested anywhere?

Patol75:
@dsroberts But do we need to run a simulation to test that? Can we not just create a mesh and then make sure that the boundary ids get identified properly?
@drhodrid Would you mind clarifying? What do you mean by testing boundary conditions?

drhodrid:
In the multi-mateiral gmsh cases, what type of boundary conditions are prescribed to the boundary IDs that you use? Here I am testing application of Dirchlet BC's both in a mixed space and the temperature function space. It's one thing to read them in correctly - it's another to know they are being used properly.

Patol75:
This should be already covered in the thermochemical van Keken benchmark (both in demos and tests), except if I am missing something. Why would the boundary conditions not work properly if they are correctly identified?

dsroberts:
We may have a bigger problem here. @drhodrid and I must be using slightly older versions of firedrake. Our results agree with each other, but not with the results when run against `firedrake_latest`. For comparison, here is the final line of `params.log`:
```
1964 0.4222041022547477 0.0002154006187965906 9.964753633921174e-10 42.86495189858589 45.40894940932058 61.626605204024266 4.912422273659474 4.912935877822018 0.0005136041625437571 0.5000000155305157
```
Its done 14 more iterations and `nu_top` is very much outside of the tolerance from the value posted by @drhodrid, `4.912422273659475` vs `4.911817173812056`

drhodrid:
Could we rule out 2 things @dsroberts:
1. Different gmsh versions generating different meshes - please use my .msh file to check -- https://www.dropbox.com/scl/fi/caqwy5ssk7yvwraetn1ln/square.msh?rlkey=xvfkfm7g2vr3kfze8hqvmdowf
2. Parallel versus serial - my job was run in serial. Was yours?

dsroberts:
Good spot @drhodrid, there is a difference in behaviour between gmsh versions. From our version, installed by pip (4.13.1)
```
Info    : Done meshing 2D (Wall 0.108392s, CPU 0.09883s)
Info    : 1927 nodes 2010 elements
Info    : Writing 'square.msh'...
Info    : Done writing 'square.msh'
```
From `firedrake_latest` Ubuntu container (4.12.1)
```
...
Info    : Done meshing 2D (Wall 0.0673855s, CPU 0.064622s)
Info    : 1926 nodes 2009 elements
Info    : Writing 'square.msh'...
Info    : Done writing 'square.msh'
```
I don't know how `.geo` files work, is this an issue with gmsh itself or is it doing a best-effort attempt at handling ambiguity or some other issue in a `.geo` file? The different versions of gmsh create identical meshes out of our other two `.geo` files.

Patol75:
I cannot reproduce the 1927-node case. I tried with Gmsh 4.13.1 inside the Firedrake virtual environment and with Gmsh 4.12.2 outside that environment; both generated 1926 nodes.

dsroberts:
@angus-g would you mind taking a look at the test setup?

tests/base_gmsh/test_base_case.py
```diff
@@ -0,0 +1,34 @@
...
+import pytest
+import warnings
+
+# Different variations of gmsh come up with different meshes from the geo file.
+# If we see a mesh different to what we're expecting, warn the user and set
+# the tests to xfail
+b = Path(__file__).parent.resolve()
+expected_md5 = "41b8157a5e18f467dbd1e7538d1236d6"
+with open(b / "square.msh", "r") as f:
+    mesh_md5 = h = hashlib.md5(f.read().encode()).hexdigest()
```
angus-g, line 13: Somethingsomething hash collisions :grin: MD5 is probably fine for this anyway!

tests/base_gmsh/test_base_case.py
```diff
@@ -0,0 +1,34 @@
...
+    Known good md5sum: {expected_md5}, Created mesh md5sum: {mesh_md5}
+    """
+    warnings.warn(UserWarning(warn_str))
+else:
+    warn_str = ""
+
+
+@pytest.mark.xfail(mesh_md5 != expected_md5, reason=warn_str)
+def test_base_case():
+    b = Path(__file__).parent.resolve()
```
angus-g, line 26:
```suggestion
    b = Path(__file__).parent.resolve()
```

Can just use the global one in this file?

tests/base_gmsh/Makefile
```diff
@@ -0,0 +1,17 @@
+all: params.log
+
+.PHONY: all base_case.py
```
angus-g, line 3:
```suggestion
.PHONY: all
```

Also I guess we should make `clean` a phony target throughout everything...?

angus-g (commented):
Test setup looks reasonable to me. Just a couple of tweaks and then this should be good to go!

tests/base_gmsh/test_base_case.py
dsroberts: Yeah, I don't think we need to be cryptographically secure here.

angus-g: Concern is more that MD5 doesn’t get deprecated out, but I don’t imagine that being an issue either.

tests/base_gmsh/Makefile
dsroberts: Yeah, I think so, strictly speaking. I guess we can start doing that here and deal with the rest later.

angus-g (approved):

PR #151, opened 2024-12-03
Generate optimisation checkpoints with collective ID

angus-g:
Mistaken change in 30a3388 (Use comm-free UUID for optimisation checkpointing) broke parallel compatibility of optimisation checkpointing. Of course, the checkpoints for the ROL vectors need to be collective, so they all need to write to the same file with the same ID. We can go back to Firedrake's _get_uid, and just use the communicator associated with the function on the vector.

### DIFF
--- a/gadopt/inverse.py
+++ b/gadopt/inverse.py
@@ -5,8 +5,8 @@ relevant parameters and call the `run` method to perform the optimisation.
 """
 
 from pathlib import Path
-import uuid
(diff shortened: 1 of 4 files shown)


w3168 (approved):
Thanks @angus-g I've tested this fixes things for me in parallel for the adjoint ice inversions. And good to have a parallel unit test too

PR #153, opened 2024-12-10
Install pygplates from PyPI

angus-g:
Now that we have the exciting new package, let's test it out. I hope I haven't missed anything. I think we just use it in the render and test workflows.


sghelichkhani (approved):
Timely! And looks good to go.  But would it make sense to provide an install option to isntall with pygplates?

PR #155, opened 2025-01-08
Automatic formatting through Ruff and pre-commit

Patol75:
I propose to replace Pylint and Flake8 with Ruff, which is a more efficient, all-in-one solution. Additionally, we should add pre-commit to ensure automatic formatting and linting of the repository and future commits.


dsroberts:
Very in favour the pre-commit hook. This will be a real time saver, I'm looking forward to not having any more 'Lint' commits in the repo. Also, I think ignoring F401 just for `__init__.py` should be retained. 

As for switching to ruff, I understand that it is more efficient than flake8, but this isn't a big code base. This update will require adding ruff to the list of package dependencies, we can't be forcing people to change their preferred IDE to contribute to g-adopt. Flake8 is already bought into the firedrake virtual env by something else, so anyone working from an existing firedrake venv can run `make lint` and it'll work. That's valuable for me at least, I don't like waiting around wondering whether the CI will pass, so I prefer to run all tests manually first, even if I've already auto-formatted in VSCode. So if we do switch to ruff, what would be the best way to bring it in to the venv? End-users don't need it, only developers, so maybe we add another `optional-dependencies` set, e.g. `develop = [ "ruff" ]`? Just something to think about.

Patol75:
One solution for F401 and `__init__.py` is to rewrite the module as I did in my last commit. It might not appeal to everyone, but it has the benefit of silencing Ruff (but not Flake8?) and is part of the accepted Python style.

I had not realised that Flake8 was automatically included through Firedrake. Still, there is a trend in the Python community to transition to Ruff, so it would be nice to support it. Building on your idea, what do you think about the proposed changes to `pyproject.toml` and `Makefile`? They should allow Flake8 and Ruff to coexist.

dsroberts:
For what its worth, I think this is a good change. I was just thinking about potential gotchas of assuming other developers are using the same tooling you are. I've seen a wide variety of IDE configurations already among g-adopt developers, some of which don't support language servers or auto formatting or anything like that. Now that Firedrake is moving to `pip install`, they're separating out `flake8` and `pylint` from the runtime requirements by having the following in `pyproject.toml` under the `[project.optional-dependencies]` setting:
```
dev = [
  "flake8",
  "pylint",
]
```
so I see no reason not to do the same in g-adopt, except with `dev = [ "ruff", "pre-commit" ]`.

I'm a tad concerned about the ability of pre-commit to modify files. I had a go at running your config across the entire g-adopt code base and it modified 116 files. Also, the results of the hook depend on whatever Python version is being used to run `pre-commit`. For example, I'm not allowed to make any changes to `gadopt/transport_solver.py` as I've installed `pre-commit` with Python 3.9, and `match` statements only appeared in Python 3.10. I don't use Python 3.9 for g-adopt, but if I open a new terminal and run `git commit`, I get my OS's default Python installation. To start with, I think we should limit `pre-commit` to just `ruff`. So just this:
```
ci:
-   repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.8.6
    hooks:
    -   id: ruff
        types_or: [ python, pyi, jupyter ]
```
This alone will do away with all the failed `make lint` tests and the follow-up 'Lint' commits that come into the repo. We can see about the other hooks in time, maybe we add in `ruff-format` and some of the `pre-commit` hooks later and at the same time do one big 'auto format everything' commit to go along with it. We'd need to plan this, as every outstanding PR and active branch would probably need rebasing after that's done.

Patol75:
I doubt there will ever be a good time to lint and format everything. It is more a question of willingness to do it, given that, as you mention, there will inherently be a number of merge conflicts. I have adjusted the hooks, as suggested. For the optional dependencies, what about `style` instead of `dev`?

tests/parallel_scaling/stokes_cubed_sphere.py
dsroberts, line 1: This change doesn't belong as part of this PR.

Makefile
dsroberts, line 1: I think this is a bit awkward. If we're including ruff in the list of dependencies we don't need to make this bit compatible with flake8 as well.

setup.cfg
dsroberts, line 1: I think its safe to delete `setup.cfg` as you were going to do before

dsroberts (approved):
Like I said before, this is good. It'll help enforce a high standard of code going into the repo. In time we can look at the other hooks, I'm just uneasy about the code that winds up in the repo being different from the code I wrote. If there was a way for those hooks to just block the commit and tell you to fix it, I'd be all for it.

tests/parallel_scaling/stokes_cubed_sphere.py
Patol75: Ruff will complain if this is not fixed, or am I missing something?

setup.cfg
Patol75: Did not we want to have Flake8 and Ruff side-by-side? `setup.cfg` provides the Flake8 configuration.

Makefile
Patol75: I am confused here. Did not we say that Flake8 comes with Firedrake at the moment?

Patol75:
About the uneasy aspect you mention, I am not sure I understand. These hooks are purely there to enforce linting and formatting standards. The changes they make automatically do not affect code execution. Am I missing something?

tests/parallel_scaling/stokes_cubed_sphere.py
dsroberts: I guess I'm just curious as to why its here but not in earlier commits to this branch. Why would ruff be complaining about it now but not before?

Makefile
dsroberts: Yes, but does that mean we have to support it? I guess this is a lead-in to a broader conversation about how far we should stray from Firedrake development practices. That conversation is way beyond the scope of this PR though, so all we have for now is opinions. In my opinion (and it is just that, an opinion) is that we can set our own linting formatting standards without stepping on any toes. So I'd say we just support ruff. Especially given, as you've already noted, ruff and flake8 apply rules inconsistently.

dsroberts:
I don't know. It doesn't sit right with me. Maybe because this kind of tooling is new to me, and it is just code in the end and all code is fallible. Just because it says it doesn't affect execution, is that true? I certainly can't verify that. I don't want to waste time working around a pre-commit hook that's breaking my code. Maybe I'm a dinosaur and I need to get with the times? Who's to say...

tests/parallel_scaling/stokes_cubed_sphere.py
Patol75:
```python
(firedrake) thomas@tristan:~/Software/g-adopt$ pre-commit run --all-files
ruff.....................................................................Failed
- hook id: ruff
- exit code: 1

tests/parallel_scaling/stokes_cubed_sphere.py:100:26: E701 Multiple statements on one line (colon)
    |
 99 |         # Solve Stokes sytem:
100 |         with stokes_stage: stokes_solver.solve()
    |                          ^ E701
101 | 
102 |         # Temperature system:
    |

tests/parallel_scaling/stokes_cubed_sphere.py:103:26: E701 Multiple statements on one line (colon)
    |
102 |         # Temperature system:
103 |         with energy_stage: energy_solver.solve()
    |                          ^ E701
    |

Found 2 errors.
```

Patol75:
I respect your point, but I have a hard time believing that the proposed hooks have the potential to break anything. What do you think of including the hooks that currently pass, i.e. those who check existing files in the repo and do not make any modifications? I did that in the last commit.

dsroberts:
That sounds like an acceptable compromise.

drhodrid:
To be discussed, merged or closed at dev meeting.

angus-g:
Closing as not planned. Definitely a good idea to deprecate `setup.cfg` and look into ruff though.

PR #157, opened 2025-02-03
Use discontinuous boundary flux to calculate level-set gradient

Patol75:
The level-set gradient is a useful quantity to determine precisely. It coincides with the normal vector to the material interface and appears thus in the reinitialisation equation. Currently, in G-ADOPT, the level-set gradient is calculated through a separate solver that employs continuous test and trial functions before being passed to the reinitialisation solver. Ideally, there should not be a need for a separate solver. Here, I propose to employ a boundary-flux approach to determine the across-cell contributions of the discontinuous flux. This way, reinitialisation can be solved on its own, and the separate gradient solver is still relevant for visualisation purposes, for example. The strategy used to express the flux is ad hoc (inspired from `momentum_equation.py`). Diagnostics across tests and demos calculated using this new approach are in excellent agreement with the previous approach.

Needs #128


Patol75:
Reminder to myself I need to update this PR to isolate the change and add the newest formulation.

drhodrid:
I spoke to @stephankramer about this during my recent visit. He will take a look when he finds time. @Patol75 if there are any updates required before Steph looks, please push them through ASAP.

Patol75:
As also seen in #185, the Robey benchmark is sensitive to the reinitialisation tuning. Whilst it is possible to monkey-patch the failure in #185, it is not easily feasible here. Should we mark the test as `xfail` before merging if changes are approved?

angus-g:
> As also seen in #185, the Robey benchmark is sensitive to the reinitialisation tuning. Whilst it is possible to monkey-patch the failure in #185, it is not easily feasible here. Should we mark the test as `xfail` before merging if changes are approved?

I suppose it depends on how much you value the benchmark: if you'll be reintroducing it later in the series then xfail is probably fine. Of course, the ideal is to demonstrate that you can still satisfy the benchmark with every successive code change.

Patol75:
I see, thank you. The problem is that the way I designed some of these benchmark tests was too specific, meaning that legitimate and useful API or methodology improvements could break the test (what we are seeing), even though the benchmark continues to run fine. The idea behind breaking up the big changeset of #128 was to isolate API changes from test changes, which I did not have in mind when first implementing the whole changeset. Given that I already have the changes for the final test-update PR, I think that abusing `xfail` temporarily on very specific cases is OK. All `xfail` tests will be reintroduced with that final PR.

Patol75:
Essentially, the physics of the benchmark works and evolution (as visualised on ParaView) makes sense and remains compatible with the `master` branch and published results. At the test level, because the test is too specific and relies a little too much on computational approximations rather than on actual physics, changes to the methodology and/or API choices break the test. What I am basically saying is that some of the tests I designed were not very useful...

angus-g:
Is it possible to slacken the tolerances rather than completely disable the test?

stephankramer:
I'm afraid I can't make much chocolate out of this - as we say in Dutch. You say you are inspired by `momentum_equation.py`, I presume the SIPG formulation for viscosity, but that's a second order derivative, linear term, here you're dealing with a first order, nonlinear term. The point of the penalty term in SIPG is to ensure the discretised 2-form is coercive (positve definite). I'm not sure what the role of a penalty term would be in this case. Also not sure about the form of the `flux_term`; I honestly don't know what the correct answer is, I don't have much experience with these kind of nonlinear terms in DG: it kind of looks like it might be what I expect, it has the right dimensions - but what is that sigma doing there, that's a very specific non-dimensional coefficient to do with the flatness of a triangle that comes from quite a convoluted derivation of the minimal value you need there to ensure coerciveness. It honestly looks a bit cargo-culty, which is not unlikely with chatgpt.

I see that you're getting good results with it, but to what extent is that contingent on getting the exact solutiont to  the reinitialization equation you think you are solving?  In the end of the day, as far as I understand it, you just want a solution that has the 0-contour in the right place and a gradient that's normal to it. So I'm not against doing something a little experimental here, but that flux and penalty term just look a little random to me. Do you still get the same results without the `dS` terms? A worry is that you might get unexpected results in cases with more anisotropic meshes (say we start using adaptivity) or that are not non-dimensionalised. Have you done a literature research to see how other people do this?

Patol75:
I have now had the opportunity to revisit this PR. What I wanted to quantify better was how much reinitialisation could be impacting results, and how different reinitialisation approaches perform. To this end, I turned to the isothermal Rayleigh-Taylor instability of van Keken et al. (1997), using a 10x viscosity contrast between two materials. For all simulations I ran, I quantified material conservation and entrainment. I ran two reference simulations without reinitialisation, one on a 128x128 mesh and the other on a 256x256 mesh. Here is the result at non-dimensional time 2000.0:
image
The background is the conservative level-set field for the 256x256 case, the black contour is the material interface in that same simulation, and the blue contour is the same interface but in the 128x128 field. Material entrainment and conservation are 0.9130478 and 0.988903 for the 256x256 case and 0.897064 and 0.950546 for the 128x128 case. There is a clear improvement in material conservation, which is expected. However, at this time, secondary instabilities have developed and blur the picture. To have a better comparison point, I use results from non-dimensional time 1500.0:
image
Material entrainment and conservation are 0.861626 and 0.993271 for the 256x256 case and 0.888138 and 0.973206 for the 128x128 case.

Now, the idea is to focus on the 128x128 case and apply reinitialisation. My idea is that it should improve conservation and, hopefully, have entrainment trend towards the value from the 256x256 case. I use 4 different strategies:
- `master`, i.e. solving the gradient on the continuous function space
- PR, i.e. the changes proposed here
- no_dS, i.e. everything on a discontinuous function space but no internal facet integral
- new, a simplified SIPG-like treatment of internal facets

For the latter, I use the following code:
```python
def reinitialisation_term(
) -> fd.Form:
    """Term for the conservative-level-set reinitialisation equation.

    Implements terms on the right-hand side of Equation 17 from
    Parameswaran, S., & Mandal, J. C. (2023).
    A stable interface-preserving reinitialization equation for conservative level set
    method.
    European Journal of Mechanics-B/Fluids, 98, 40-63.
    """
    grad_trial = fd.grad(trial)
    grad_mag = fd.sqrt(fd.inner(grad_trial, grad_trial))

    sharpening = -trial * (1 - trial) * (1 - 2 * trial)
    correction = eq.epsilon * (1 - 2 * trial) * grad_mag
    volume_term = (sharpening + correction) * eq.test * eq.dx

    alpha = 1e-5 * trial.ufl_element().degree() ** 2 / fd.avg(fd.CellDiameter(eq.mesh))

    jump_penalty = alpha * fd.jump(trial) * fd.jump(eq.test) * eq.dS

    return volume_term - jump_penalty
```

Here are the results on ParaView at time 1500.0, with the same black and blue contours as before and added green (`master`), yellow (new), red (no_dS), and purple (PR) contours, for the full domain and a zoom towards the top:
image
image
It is pretty clear that applying reinitialisation improves results. Looking at conservation and entrainment, I get:
- `master`, 0.907745 and 0.989990
- PR, 0.892715 and 0.987091
- no_dS, 0.893250 and 0.989304
- new, 0.891800 and 0.987778

TBC (!)

PR #158, opened 2025-02-04
Backend changes: Stokes integrators and Approximations

Patol75:
Continuing the series of PRs tackling G-ADOPT's backend, the current set of changes demonstrates an alternate way to structure `approximations.py` and `stokes_integrators.py` alongside minimal changes to base demos. While early motivation included removing the need to provide a temperature field and a Rayleigh number, more recent endeavours focused on achieving a structure that can provide a basis for future solver developments and original contributions to G-ADOPT.

One can see that current GIA, MC, and MM paradigms fit well under the proposed structure. However, implementing the internal variable approach for compressible viscoelastic models demonstrates that a unique approximation model may not be desirable. I envisage that a family of approximations can be grouped within a class, as implemented here for viscous mantle approximations, and each family can inherit from a base approximation class. This differs from the current `master` implementation where the notion of family is approached through class inheritance. The structure I envisage will become more explicit as the internal variable development progresses.

For the solvers' structure, the inheritance from a common abstract base class drastically simplifies the implementation of new solvers with minimal, if any, code duplication.

Discussion regarding these proposed changes is most welcomed.

### DIFF
--- a/demos/glacial_isostatic_adjustment/base_case/base_case.py
+++ b/demos/glacial_isostatic_adjustment/base_case/base_case.py
@@ -411,13 +411,19 @@ ice_load = ramp * rho_ice * g * Hice * disc
 # Setup boundary conditions
 exterior_density = conditional(time < t2_load, rho_ice*disc, 0)
 stokes_bcs = {
-    bottom_id: {'uy': 0},
-    top_id: {'normal_stress': ice_load, 'free_surface': {'delta_rho_fs': density - exterior_density}},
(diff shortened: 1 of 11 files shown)


drhodrid:
Can this be closed @Patol75 - I realise it's a draft now, but is largely superseded by more recent pull requests, right?

Patol75:
The only reason it is opened is to provide an all-in-one view of the coupled changes to other developers.

drhodrid:
Closing for now.

PR #163, opened 2025-02-09
Swap argument order in timestepper advance method

Patol75:
The current argument order in `TimeIntegratorBase.advance` and `RungeKuttaTimeIntegrator.advance` is not ideal. `t` is a required argument, whilst `update_forcings` is an optional argument. However, `t` is only used as an argument to `update_forcings`, which breaks the previous logic. In other words, passing `t` currently without providing `update_forcings` has no impact, as can be seen from `tests/free_surface/explicit_free_surface.py`. Here, the argument order is swapped, and both arguments become optional.

### DIFF
--- a/gadopt/time_stepper.py
+++ b/gadopt/time_stepper.py
@@ -9,7 +9,7 @@ providing relevant parameters defined in the parent class (i.e. `ERKGeneric` or
 import operator
 from abc import ABC, abstractmethod
 from numbers import Number
-from typing import Any, Optional
+from typing import Any, Callable, Optional
 
 import firedrake
 import numpy as np
@@ -22,12 +22,14 @@ class TimeIntegratorBase(ABC):
     """Defines the API for all time integrators."""
 
     @abstractmethod
-    def advance(self, t: float, update_forcings: Optional[firedrake.Function] = None):
+    def advance(self, update_forcings: Callable | None = None, t: float | None = None):
         """Advances equations for one time step.
 
-        Arguments:
-          t: Current simulation time
-          update_forcings: Firedrake function used to update any time-dependent boundary conditions
+        Args:
+          update_forcings:
+            Callable updating any time-dependent equation forcings
+          t:
+            Current simulation time
 
         """
         pass
@@ -104,12 +106,15 @@ class RungeKuttaTimeIntegrator(TimeIntegrator):
         """
         pass
 
-    def advance(self, t, update_forcings=None):
+    def advance(
+        self, update_forcings: Callable | None = None, t: float | None = None
+    ) -> None:
         """Advances equations for one time step."""
         if not self._initialized:
             self.initialize(self.solution)
         for i in range(self.n_stages):
-            self.solve_stage(i, t, update_forcings)
+            self.solve_stage(i, update_forcings, t)
+
         self.get_final_solution()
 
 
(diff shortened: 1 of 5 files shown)


Patol75:
Just noting that test failure is unrelated and is addressed in gh-128.

gadopt/time_stepper.py
```diff
@@ -313,15 +321,19 @@ def solve_tendency(self, i_stage, t, update_forcings=None):
             self.solution_old.assign(self.solution)
         if not self._initialized:
             raise ValueError('Time integrator {:} is not initialized'.format(self.name))
-        if update_forcings is not None:
-            update_forcings(t + self.c[i_stage]*self.dt)
+
+        if update_forcings is not None and t is not None:
+            update_forcings(t + self.c[i_stage] * self.dt)
+        elif update_forcings is not None:
```
w3168, line 327: What's the use case here?

w3168 (commented):
Hi @Patol75 thanks for this! I think in principal it makes a lot of sense to rearrange the order like you suggest. 

At the dev meeting last Wednesday @sghelichkhani raised a point that maybe there is a different solution to update forcings for the rk stages... I guess if some forms are analytical expressions of time then time could be passed as an argument when setting up the residual terms so they just depend symbolically on time..... But maybe ```update_forcings``` is more convenient and doing an interpolation of the time dependent expression is ok at each stage of the timestepper when it is needed. 

Also when you have time do you mind explaining why you have the ```functools.partial``` method to setup the callable? And just to check it is only the Trim benchmark 

From looking at this I realised that probably that a time dependent ice loading bc could/should also be implemented using update forcings (if using multistage timestepping methods) ...

gadopt/time_stepper.py
Patol75: This is used at the moment for level-set reinitialisation. The forcing here is the gradient calculation.

w3168: Ok that makes sense!

Patol75:
Hey Will,

Thanks for looking at the PR!

Regarding what Sia suggested, `time_stepper.py` has no knowledge of time. All it knows is the time step and the integration scheme, which seems reasonable to me. Changing this would require some rewriting, but I am not sure if there would be merit to it. I am not sure either if having forms symbolically depend on time is desirable. Happy to be corrected if wrong. 

I am not exactly sure which `functools.partial` call you are referencing here, but I imagine the idea is to ensure the provided callable is only a function of time, as this is the expectation in `time_stepper.py`.

Yes, if you provide an `Equation` with a time-dependent term/condition to one of G-ADOPT's time integrators, then you should definitely use `update_forcings` to accurately calculate the temporal integration.

w3168:
Thanks @Patol75 I agree that changing timestepper to understand about time requires a major rewrite/think and maybe not that helpful since we have discrete timesteps anyway... 

Here's the ```functools.partial``` in master I was curious about https://github.com/g-adopt/g-adopt/blob/afd1a06c9160c19fc9bfeb442d9b19d84ecb884c/tests/multi_material/run_benchmark.py#L285

and in your new branch https://github.com/g-adopt/g-adopt/blob/0a9cac232a8d93862c0b73f909a4aa7e52691097/tests/multi_material/run_benchmark.py#L279

Patol75:
Well, yes, that is the one I had in mind. :)

w3168 (approved):

PR #164, opened 2025-02-09
Remove unused arguments in `time_stepper.py`

Patol75:
Time integrators in `time_stepper.py` accept two arguments that are not used: `bnd_conditions` and `terms_to_add`. This PR removes these extraneous arguments.

### DIFF
--- a/gadopt/time_stepper.py
+++ b/gadopt/time_stepper.py
@@ -124,7 +124,6 @@ class ERKGeneric(RungeKuttaTimeIntegrator):
       dt: Integration time step
       solution_old: Firedrake function representing the equation's solution
                       at the previous timestep
-      bnd_conditions: Dictionary of boundary conditions passed to the equation
       solver_parameters: Dictionary of solver parameters provided to PETSc
       strong_bcs: List of Firedrake Dirichlet boundary conditions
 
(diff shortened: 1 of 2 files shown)


w3168:
Hi @Patol75 is this a good idea? 

I think you are removing functionality that allows us to implement weak boundary conditions for time dependent equations and (presumably) mark whether terms are implicit/excplicit for an imex timestepping scheme. 

I have definitely used the first part to implement weak bcs as part of the thwaites code.

I guess there is a separate issue that none of this is tested. Maybe we could add tests as opposed to removing functionality?  (For context this code has been adapted from thetis https://thetisproject.org/ by @stephankramer )

Patol75:
I do not believe I am removing any functionality here. I am merely deleting arguments that should not be there because they are not associated with any code in G-ADOPT. I understand that they may facilitate some functionalities in other libraries, such as Thwaites or Thetis, however, this is currently not the case in G-ADOPT.

w3168:
Thanks for the reply and chatting just now. I should have checked the code first haha! I forgot/got confused with the changes here https://github.com/g-adopt/g-adopt/commit/1d10b20dbbfbc1cd93b7fd51e3b27589b2c7e7bb where bcs is now passed to the Equation object as opposed to the timestepper directly. I agree we should remove the weak bcs argument from the solver, as the new approach is definitely cleaner. 

I did a bit of digging trying to find where ```terms_to_add''' originated from. Maybe there is some functionality in Thetis to implement a leap frog scheme here https://github.com/thetisproject/thetis/blob/master/thetis/timeintegrator.py

Anyway probably best to remove the argument and then down the line if we need to add an imex scheme or something like it we can do it then....

w3168 (approved):

PR #170, opened 2025-02-19
Change su_diffusivity back to an on/off flag

w3168:
PR #114 changed the api to allow users to specify an ```su_diffusivity``` This is maybe confusing because if a 'real' background diffusivity is given then we should use this diffusivity used to calculate the grid peclet number for the su advection scheme.

In the case where there is no diffusion term we just set kappa=0 since we already have a hack to specify a (hopefully) small 1e-12 value for the diffusivity anyway.

### DIFF
--- a/gadopt/transport_solver.py
+++ b/gadopt/transport_solver.py
@@ -85,9 +85,9 @@ class GenericTransportBase(abc.ABC):
       solver_parameters:
         Dictionary of solver parameters or a string specifying a default configuration
         provided to PETSc
-      su_diffusivity:
-        Float activating the streamline-upwind stabilisation scheme and specifying the
-        corresponding diffusivity
+      su_advection:
+        Boolean activating the streamline-upwind stabilisation scheme when using
+        continuous finite elements
 
     """
 
@@ -109,7 +109,7 @@ class GenericTransportBase(abc.ABC):
         eq_attrs: dict[str, float] = {},
         bcs: dict[int, dict[str, Number]] = {},
         solver_parameters: dict[str, str | Number] | str | None = None,
-        su_diffusivity: float | None = None,
+        su_advection: bool = False,
     ) -> None:
         self.solution = solution
         self.delta_t = delta_t
@@ -118,7 +118,7 @@ class GenericTransportBase(abc.ABC):
         self.eq_attrs = eq_attrs
         self.bcs = bcs
         self.solver_parameters = solver_parameters
-        self.su_diffusivity = su_diffusivity
+        self.su_advection = su_advection
 
         self.solution_space = solution.function_space()
         self.mesh = self.solution_space.mesh()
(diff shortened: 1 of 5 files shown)


Patol75:
I think I disagree with these changes but am happy to be overruled. For the `Equation` instance to include a diffusivity within its attributes, a diffusion term has to be provided. In other words, the problem at hand contains physical diffusion. My understanding of streamline-upwind (SU) advection is that it is a mathematical strategy to tackle the stabilisation of pure advection problems. It is not clear to me that the SU artificial diffusion in that case has anything to do with a physical diffusive process.

Given the above, I am quite unsure about SU diffusion being desirable if physical diffusion is present, and if it is not, then `Equation` should not know of any diffusion attribute.

Patol75 (commented):
I think I disagree with these changes but am happy to be overruled. For the `Equation` instance to include a diffusivity within its attributes, a diffusion term has to be provided. In other words, the problem at hand contains physical diffusion. My understanding of streamline-upwind (SU) advection is that it is a mathematical strategy to tackle the stabilisation of pure advection problems. It is not clear to me that the SU artificial diffusion in that case has anything to do with a physical diffusive process.

Given the above, I am quite unsure about SU diffusion being desirable if physical diffusion is present, and if it is not, then `Equation` should not know of any diffusion attribute.

w3168:
Hi @Patol75 thanks for looking at this!  When I was implementing it before I was looking at this book chapter
[FiniteElementMethodsforFlowProblems2003Donea.pdf](https://github.com/user-attachments/files/18877806/FiniteElementMethodsforFlowProblems2003Donea.pdf)

I think the definition of the artificual su diffusivity is tied to the physical diffusion specifed because we are calculating a grid Peclet number based on the real diffusion. If the grid peclet number is less than  1 then the su diffusivity added will be negligble. It only kicks in if the grid resolution is too coarse for the given flow and real diffusivity (hence high grid peclet number).

I agree though this kind of scheme is purely for robustness so maybe it is a good idea to be able to play with it... I guess from what I've seen (and maybe how we implemented it!) it is incredibly diffusive, which has motivated @drhodrid switch to using DG more often for the energy equation...

Patol75:
Hmm... But then, how does it make sense if you consider a pure advection problem?

w3168:
so I think if you have no diffusivity the grid peclet number will always be infinite. so the beta(Pe) function tends to 1.  And then nubar is just calculted to set grid Pe =1.  At the moment we've implemented this if the diffusion provided is zero then we calculate the grid peclet number with a (hopefully) small diffusivity of 1e-12 to avoid the divide by zero.

Patol75:
Having looked at the document, I think what you say makes sense. However, why have we limited ourselves to SU and not implemented SUPG (Section 2.4.1)? Figure 2.13 shows quite a difference between the two.

sghelichkhani:
> Having looked at the document, I think what you say makes sense. However, why have we limited ourselves to SU and not implemented SUPG (Section 2.4.1)? Figure 2.13 shows quite a difference between the two.

I think Steph did try to implement SUPG and there might be some code somewhere lying around for that. My recollection is that his test was showing more diffusive behavior for the test (!!!). It might be a good idea to get someone to experiment with SUPG at a later stage.

PR #171, opened 2025-02-20
Handle MixedFunctionSpace.mesh() being a MeshSequence

dsroberts:
Currently, a MixedFunctionSpace in firedrake can only be defined on a single mesh. The `_mesh` attribute of the mixed function space is set to the mesh object referenced in the first of the `FunctionSpace`'s passed to MixedFunctionSpace (see [here](https://github.com/firedrakeproject/firedrake/blob/a8fbe18ca607f0f79a5f7db336950b43f96a43b0/firedrake/functionspace.py#L290)). In the submesh branch, this requirement is dropped, and a new `MeshSequenceGeometry` object is created for the MixedFunctionSpace, regardless of whether the meshes are identical or not (see [here](https://github.com/firedrakeproject/firedrake/blob/18de7f0435f32c54d2b6c0491d3eb07d8aa3de68/firedrake/functionspace.py#L300)). The implication for g-adopt is that the `mesh.cartesian` property is no longer guaranteed to be present on a mesh passed to `upward_normal` via the `StokesSolver.__init__` method. Rather than storing `self.k` during `__init__`, I've replaced every reference to `self.k` with `upward_normal(fd.ufl_expr.extract_unique_domain(<function space>))`, which is guaranteed to return a `Mesh` object if passed a `FunctionSpace`, which it always is in g-adopt. The references are only in setup functions, so there shouldn't be much additional overhead. 

I've also made the `cartesian` parameter for the mesh optional since (as far as I can see) it was only used for `upward_normal` calculations. The assumption being that we're in working in polar coordinates unless told otherwise. The change is backward compatible, so `mesh.cartesian = False` is still valid. I wasn't around for the discussions related to the `cartesian` parameter, so if this idea was already dismissed, I'll revert the changes.

### DIFF
--- a/gadopt/stokes_integrators.py
+++ b/gadopt/stokes_integrators.py
@@ -238,8 +238,6 @@ class StokesSolver:
 
         self.solver_kwargs = kwargs
 
-        self.k = upward_normal(self.mesh)
-
         # Setup boundary conditions
         self.weak_bcs = {}
         self.strong_bcs = []
@@ -249,6 +247,18 @@ class StokesSolver:
         self.free_surface_dt = free_surface_dt
         self.free_surface_theta = free_surface_theta  # theta = 0.5 gives a second order accurate integration scheme in time
         self.free_surface = False
+
+        # Detect if solver should be direct, iterative or whatever the user has chosen
+        solver_auto_is_direct = None
+        if solver_parameters is not None:
+            if isinstance(self.mesh, fd.MeshSequence):
+                solver_auto_is_direct = self.mesh.topological_dimension() and all(
+                    hasattr(m, "cartesian") and m.cartesian for m in self.mesh
+                )
+        else:
+            solver_auto_is_direct = (
+                self.mesh.topological_dimension() == 2 and hasattr(self.mesh, "cartesian") and self.mesh.cartesian
+            )
 
         for id, bc in bcs.items():
             weak_bc = {}
@@ -331,7 +341,7 @@ class StokesSolver:
                         raise ValueError(
                             f"Solver type '{solver_parameters}' not implemented."
                         )
-            elif self.mesh.topological_dimension() == 2 and self.mesh.cartesian:
+            elif solver_auto_is_direct:
                 self.solver_parameters.update(direct_stokes_solver_parameters)
             else:
                 self.solver_parameters.update(iterative_stokes_solver_parameters)
@@ -357,7 +367,7 @@ class StokesSolver:
 
     def setup_equation_attributes(self):
         stress = self.approximation.stress(self.u)
-        source = self.approximation.buoyancy(self.p, self.T) * self.k
+        source = self.approximation.buoyancy(self.p, self.T) * upward_normal(fd.ufl_expr.extract_unique_domain(self.p))
         self.eqs_attrs = [{"p": self.p, "stress": stress, "source": source}, {"u": self.u, "rho_mass": self.rho_mass}]
 
     def setup_free_surface(self):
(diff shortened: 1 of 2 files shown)


angus-g:
> I've also made the cartesian parameter for the mesh optional since (as far as I can see) it was only used for upward_normal calculations.

I think the idea here was to prevent accidentally working in the wrong coordinate system by enforcing something to be chose. Nothing "breaks" if you have it wrong, but all your results will be nonsense, so it can be an expensive mistake to make.

dsroberts:
> > I've also made the cartesian parameter for the mesh optional since (as far as I can see) it was only used for upward_normal calculations.
> 
> but all your results will be nonsense, so it can be an expensive mistake to make.
> 
Yes, that's a good point. This is one of those security features that, on the surface, can be annoying but then it saves you once and all is forgiven. I'll revert that bit and we can discuss further another time.

gadopt/stokes_integrators.py
```diff
@@ -250,6 +248,14 @@ def __init__(
         self.free_surface_theta = free_surface_theta  # theta = 0.5 gives a second order accurate integration scheme in time
         self.free_surface = False
 
+        # Detect if solver should be direct, iterative or whatever the user has chosen
+        solver_auto_is_direct = None
+        if isinstance(self.mesh, fd.MeshGeometry):
+            solver_auto_is_direct = self.mesh.topological_dimension() == 2 and self.mesh.cartesian
+        else:
+            # A MeshSequence is a tuple of MeshGeometry's
+            solver_auto_is_direct = self.mesh.topological_dimension() and all(m.cartesian for m in self.mesh)
```
angus-g, line 257:
```suggestion
            solver_auto_is_direct = self.mesh.topological_dimension() == 2 and all(m.cartesian for m in self.mesh)
```

Is this missing?

gadopt/stokes_integrators.py
```diff
@@ -250,6 +248,14 @@ def __init__(
         self.free_surface_theta = free_surface_theta  # theta = 0.5 gives a second order accurate integration scheme in time
         self.free_surface = False
 
+        # Detect if solver should be direct, iterative or whatever the user has chosen
+        solver_auto_is_direct = None
+        if isinstance(self.mesh, fd.MeshGeometry):
+            solver_auto_is_direct = self.mesh.topological_dimension() == 2 and self.mesh.cartesian
+        else:
+            # A MeshSequence is a tuple of MeshGeometry's
+            solver_auto_is_direct = self.mesh.topological_dimension() == 2 and all(m.cartesian for m in self.mesh)
```
angus-g, line 257: Great, everything seems reasonable now. I'm just wondering about the `all` check here. Given I haven't tried using a `MeshSequence`, what does it look like to set up the g-adopt compatible mesh now?

dsroberts: I don't know what happens to g-adopt when we set up a `MixedFunctionSpace` on multiple different meshes, I think that's a bridge we'll need to cross when we get to it. I used `all` here as that made the logic backwards-compatible with the current single-mesh `MixedFunctionSpace`

angus-g: I guess it will make itself pretty clear if you don't set it on all the meshes, so that's fine for the moment.

angus-g (approved):
Without a concrete example using `MeshSequence`, this seems to make sense. I expect there could still be some tweaks needed depending on how everything interacts, but at least it won't get directly in the way of trying to using submeshes?

dsroberts:
> Without a concrete example using `MeshSequence`, this seems to make sense. I expect there could still be some tweaks needed depending on how everything interacts, but at least it won't get directly in the way of trying to using submeshes?

Yeah, this isn't done yet. Another issue has come up around this with `MeshSequence`. Will figure it out.

PR #173, opened 2025-02-20
Switch CI to use Firedrake pip install

angus-g:
- [x] switch rendering to use pip container too
- [x] will need to switch container tag when it's officially released

### DIFF
--- a/.github/workflows/render.yml
+++ b/.github/workflows/render.yml
@@ -15,8 +15,8 @@ jobs:
   render:
     runs-on: self-hosted
     container:
-      image: firedrakeproject/firedrake:latest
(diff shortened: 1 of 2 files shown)


angus-g:
Looks like Firedrake actually broke the existing images, but at least the new tag is in place, so we should be ready to switch over to this.

Render run: https://github.com/g-adopt/g-adopt/actions/runs/13560441984

Test run: https://github.com/g-adopt/g-adopt/actions/runs/13560441984

Assuming these are successful, we're probably good to go?

sghelichkhani (approved):
Needed to make the CI work again. Thanks @angus-g .

PR #178, opened 2025-02-28
Allow running demos without task-spooler

### LINKED ISSUE #167: Add timeout to test cases in CI
With the recent issue of a deadlock causing tests to hang until the full 6-hour window for Actions is reached (https://github.com/g-adopt/g-adopt/actions/runs/13303073225), it seems like we should pre-emptively kill cases that run over time. This means that we can more easily identify which cases (if any) are taking excessively long or completely hung, and we can also still get their full output, which could be helpful for figuring out what's going on.

I suppose it would be ideal to enforce this at the `tsp` level, otherwise we'll have to make sure that every command is safeguarded with whatever mechanism we choose.

### LINKED ISSUE #175: Allow running tests without task-spooler
Should also remove `/usr/bin/time` from our run commands. Would be nice to be able to set the run command from a configuration makefile, instead of having it in every single directory.

- [ ] remove `/usr/bin/time`
- [ ] allow running without task-spooler
- [ ] switch to @dsroberts improved task-spooler implementation

angus-g:
Also gets rid of the requirement on `/usr/bin/time` which has different behaviour on BSD (i.e. macOS).

- [x] also fix up `tests` directory
- [x] fix "Output timing information" step of CI

Closes #175.
Closes #167.

# How it works
This adds a `rules.mk` in the base directory, which has methods for running tests in `BATCH_MODE` (i.e. CI) or interactively, and uses `mpiexec` or not, depending on the case. Both modes will print some timing information. The important section is the `run-python` canned recipe which depends on the following variables: `$ncpus`, `$category`, `$PETSC_FLAGS`, `$exec_args` and expects the Python script to be the first dependency (as `$<`). There's also a pattern rule which uses the canned recipe. For example:

```makefile
params.log : category := mantle_convection
params.log: run-base_case
```

The `check` rules all use a relative path to find the test fixture, so they will work with `make -C` from wherever.

The `analytical_comparisons` test still requires task-spooler.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,8 +53,7 @@ jobs:
           S3CONFIG: ${{ secrets.S3CONFIG }}
         run: |
           sudo apt update
-          sudo apt install -y task-spooler time s3cmd unzip git-lfs gmsh
-          tsp -S 48
+          sudo apt install -y time s3cmd unzip git-lfs gmsh gcc cmake libsqlite3-dev
           mkdir -p $HOME/.s3cfg
           echo "${S3CONFIG}" > $HOME/.s3cfg/config
           chmod 600 $HOME/.s3cfg/config
@@ -69,6 +68,23 @@ jobs:
       - name: Fetch LFS artifacts
         run: |
           git lfs checkout
+
+      - name: Fetch task spooler
+        uses: actions/checkout@v4
+        with:
+          repository: dsroberts/tsp_for_hpc
+          path: ./tsp
+
+      - name: Build task spooler
+        uses: threeal/cmake-action@v2
+        with:
+          source-dir: ./tsp
+          build-dir: ./tsp_build
+          options: CGROUPV2=ON
+
+      - name: Put task spooler on PATH
+        run: |
+          mv tsp_build/tsp-hpc /home/firedrake/.local/bin/tsp
 
       - name: Install Python dependencies
         run: |
@@ -107,7 +123,6 @@ jobs:
           make -j test
         env:
           GADOPT_LOGLEVEL: INFO
-          TS_MAXFINISHED: 100
           BATCH_MODE: 1
 
       - name: Pytest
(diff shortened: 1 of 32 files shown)


angus-g:
I originally thought it was cleaner to set up the `run-%` rule, but then had to use the canned recipe in a few special cases. So maybe I should just use that everywhere, like:

```make
params.log: base_case.py
    $(run-python)
```

instead of:

```make
params.log: run-base_case
```

angus-g:
Yeah, that's fair. I will factor back to the canned recipe in all cases.

angus-g:
Looks like the "Output timing information" step on the CI is breaking on some of the cases, I'll look into that.

angus-g:
Should be good now.

dsroberts:
It looks like the uncategorised tests/demos are taking the `run_regular` path through the makefile, is that intentional?
```
analytical delta_cylindrical_freeslip_dpc took 3048 seconds
analytical delta_cylindrical_zeroslip took 3203 seconds
[multi_material]tosi_2015 took 
```
I can modify `tsp` so that your `sed` command will pull out the runtime. I'll also make it output the job id as you're using it in this workflow.

angus-g:
> It looks like the uncategorised tests/demos are taking the `run_regular` path through the makefile, is that intentional?

The analytical cases should be using `run_regular` because they have the weird "run lots of cases for convergence" thing going on. Multi-material should be using `run_batch` though (but I modified the `desc` variable to describe the case for those).

I missed you adding that change for compatibility, thanks (and then I asked for the time summary anyway, oops!) I realise that using the summary straight from tsp will miss the prettified `desc` variables, and maybe it's nicer to know queue + runtime anyway?

In general, I still think it is nice to have tsp print the timing summary. Maybe gated behind a flag so the default remains compatible, and we go back to the manual timekeeping + `sed` here? Very happy for feedback though!

angus-g:
Yeah, something like that would work well, I think.

angus-g:
> I can give `tsp` some time printing options, so `id=$$(tsp -f ... ); echo "$(2) took $$( tsp --print-total-time $id )"` ?

@dsroberts Is this available? I'd like to get this in so we can get the test into upstream Firedrake, if possible.

dsroberts:
Yep, done now. Let me know if there are any issues.

angus-g:
Thanks! Assuming the run passes, I think this will be good for review now?

rules.mk
```diff
@@ -0,0 +1,100 @@
...
+# Execution wrapper
+# -----------------
+#
+# When running in batch mode, we want an additional wrap inside the
+# `timeout` utility to enforce a 2 hour time limit on any given task.
+# This ensures that CI jobs don't exceed the 6 hour time limit on
+# GitHub Actions so they can be torn down gracefully and provide
+# output information.
+
+exec_wrapper = $(if $(BATCH_MODE),timeout 2h $(exec_cmd),$(exec_cmd))
```
dsroberts, line 77: Making `timeout` the thing `tsp` runs breaks OpenMPI binding as `tsp` always expects some variant `mpirun` or `mpiexec` to launch MPI. What I can do in `tsp` is have it scan the rest of its command line arguments for something that looks like `mpirun` or `mpiexec` if its first argument is `timeout` (or any other job control program we want to use one day). Looks like it doesn't really affect the CI, but I'm running a job on Gadi right now with `$BATCH_MODE` enabled and its helpfully decided to place 2 MPI ranks on the same core for some reason...

dsroberts (commented):
This is great, I just have one thought on the `analytical_comparisons` makefile, but other than that, no real issues. Being able to jump into any test or demo on any system and have `make` and `make check` just work is awesome!

rules.mk
angus-g: Since this is only a batch mode thing, can `tsp` enforce the timeout itself? Does it keep a control process around, or just fork into mpi and disappear?

dsroberts: For clean-up purposes, it has to. However, it's doing the simplest implementation of `init` possible, just waiting on every child process until there aren't any more, at which point it reports the job has finished and exits. I'm starting to think that given all the stuff we want out of `tsp`, maybe we turn it into a library and build a few utility programs on top of it that can be run alongside. `tsp-memprof`, `tsp-timeout`... Keeps `tsp` itself simple and allows building extensions on top of the existing framework. The job running and job reporting side of `tsp` are disparate enough that they could already be 2 separate applications.

.github/workflows/test.yml
```diff
@@ -70,6 +69,23 @@ jobs:
         run: |
           git lfs checkout
 
+      - name: Fetch task spooler
+        uses: actions/checkout@v4
+        with:
+          repository: dsroberts/tsp_for_hpc
+          path: ./tsp
```
dsroberts, line 76:
```suggestion
          path: ./tsp
          ref: v1.0.0
```
I've dropped a release for tsp, this way any further experimental changes don't break our CI

rules.mk
angus-g: Yeah, that sounds like a good way to deal with some of the requirements. For the moment, should we just drop the timeout and bring it in as a tsp extension? There are hard timeouts (actions, PBS job) that will save us anyway.

dsroberts: I think its best to leave `timeout` there. That way the output is retained in the database for analysis afterwards. MPI binding is generally off by default, so what happens in that case is that OpenMPI will respect the affinity group set by `tsp`, but the individual MPI processes will be free to move around within that group. The benefits of `timeout` in the CI outweigh the very minor loss in performance that not having MPI process binding causes.

dsroberts (approved):
LGTM.

PR #181, opened 2025-03-05
Correct storage flags on analytical longtest run

angus-g:
We (I) had been pointing at the old module files, oops!


sghelichkhani (approved):
Thanks Angus!

PR #182, opened 2025-03-06
Use new tsp in github actions

dsroberts:
Update github actions to use the new task spooler implementation. No memory profiling yet. I apologise in advance for how spammy this will probably get.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -53,7 +53,7 @@ jobs:
           S3CONFIG: ${{ secrets.S3CONFIG }}
         run: |
           sudo apt update
-          sudo apt install -y time s3cmd unzip git-lfs gmsh gcc cmake libsqlite3-dev
+          sudo apt install -y time s3cmd unzip git-lfs gmsh gcc cmake libsqlite3-dev tree
           mkdir -p $HOME/.s3cfg
           echo "${S3CONFIG}" > $HOME/.s3cfg/config
           chmod 600 $HOME/.s3cfg/config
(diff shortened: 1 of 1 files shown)


angus-g:
Looks like it's expecting the cgroup v1 hierarchy, but the system uses cgroup v2 (aka the `CRAY` cpp def)

.github/workflows/test.yml
```diff
@@ -53,8 +53,7 @@ jobs:
           S3CONFIG: ${{ secrets.S3CONFIG }}
         run: |
           sudo apt update
-          sudo apt install -y task-spooler time s3cmd unzip git-lfs gmsh
-          tsp -S 48
+          sudo apt install -y time s3cmd unzip git-lfs gmsh gcc cmake libsqlite3-dev tree
```
angus-g, line 56: `tree` was only from your testing, right?

angus-g:
Looks like there were a few warnings like the following:

```
--------------------------------------------------------------------------
WARNING: Open MPI tried to bind a process but failed.  This is a
warning only; your job will continue, though performance may
be degraded.

  Local host:        92237a818412
  Application name:  /__w/g-adopt/g-adopt/.venv/bin/python3
  Error message:     failed to bind memory
  Location:          ../../../../../../orte/mca/rtc/hwloc/rtc_hwloc.c:447

--------------------------------------------------------------------------
[92237a818412:02506] 3 more processes have sent help message help-orte-odls-default.txt / memory not bound
[92237a818412:02506] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

I expect it's probably not worth chasing too far, but maybe coming about now with the CPU pinning and the NUMA effects thereof?

angus-g (approved):
Pretty much drop-in, nice! I will tweak #175 to base on this.

dsroberts:
Given you've implemented this in #178, I'll close this PR.

PR #183, opened 2025-03-07
Set comm on restored CheckpointedROLVector

angus-g:
Should fix an error when checkpointing is enabled after restarting a checkpointed optimisation.


sghelichkhani:
Great! Thanks Angus. I confirm this fixes the issue on our test. Let's see if it fixes the issue on our big job.

sghelichkhani (approved):

PR #185, opened 2025-03-10
Update `LevelSetSolver`

Patol75:
Extracts changes from gh-128 related only to `LevelSetSolver`.


gadopt/level_set_tools.py
```diff
@@ -29,22 +30,6 @@
 ]
```
dsroberts, line 32:
```suggestion
# Default parameters for level-set advection
adv_params_default = {
    "time_integrator": eSSPRKs10p3,
    "bcs": {},
    "subcycles": 1,
    "solver_params": {
        "pc_type": "bjacobi",
        "sub_pc_type": "ilu",
    },
}
# Default parameters used to set up level-set reinitialisation
reini_params_default = {
    "timestep": 0.02,
    "time_integrator": eSSPRKs3p3,
    "steps": 1,
    "solver_params": {
        "pc_type": "bjacobi",
        "sub_pc_type": "ilu",
    },
}

```

gadopt/level_set_tools.py
```diff
@@ -152,190 +137,257 @@ def reinitialisation_term(
...
+            self.reini_kwargs["time_integrator"] = eSSPRKs3p3
+        if "solver_params" not in self.reini_kwargs:
+            self.reini_kwargs["solver_params"] = {
+                "pc_type": "bjacobi",
+                "sub_pc_type": "ilu",
+            }
+        if "steps" not in self.reini_kwargs:
+            self.reini_kwargs["steps"] = 1
+        if "frequency" not in self.reini_kwargs and self.mesh.cartesian:
+            self.reini_kwargs["frequency"] = self.reinitialisation_frequency()
```
dsroberts, line 258:
```suggestion
        if isinstance(adv_kwargs, dict):
            self.advection = True
            self.set_default_advection_args(adv_kwargs)

        if isinstance(reini_kwargs, dict):
            self.reinitialisation = True
            self.set_default_reinitialisation_args(reini_kwargs)

        self._solvers_ready = False

    def set_default_advection_args(self, adv_kwargs: dict[str, Any]) -> None:
        """Set default values of optional arguments if not provided."""
        if "u" not in adv_kwargs:
            raise KeyError("'u' must be present in adv_kwargs")
        if "timestep" not in adv_kwargs:
            raise KeyError("'timestep' must be present in adv_kwargs")
        solver_params = adv_kwargs["solver_params"] if "solver_params" in adv_kwargs else {}
        self.adv_kwargs = adv_params_default | adv_kwargs
        self.adv_kwargs["solver_params"] = adv_params_default["solver_params"] | solver_params

    def set_default_reinitialisation_args(self, reini_kwargs: dict[str, Any]) -> None:
        """Set default values of optional parameters if not provided."""
        if "epsilon" not in reini_kwargs:
            raise KeyError("'epsilon' must be present in reini_kwargs")
        solver_params = reini_kwargs["solver_params"] if "solver_params" in reini_kwargs else {}
        self.reini_kwargs = reini_params_default | reini_kwargs
        self.reini_kwargs["solver_params"] = reini_params_default["solver_params"] | solver_params
        if "frequency" not in reini_kwargs and self.mesh.cartesian:
            self.reini_kwargs["frequency"] = self.reinitialisation_frequency()
```

gadopt/level_set_tools.py
dsroberts: This change goes with the one below. I think its worth bringing the default settings up here similarly to the solver parameters in `stokes_integrators.py` and `transport_solver.py` (or make them a static property of the class). This way if there are any new default parameters required they can be added here instead of extending the `if "thing" not in...` chain in the `set_default_*_args()` functions below

gadopt/level_set_tools.py
dsroberts: This goes with the change above. Rather than defining which keys the `adv_kwargs` and `reini_kwargs` class members should have in these functions, they're defined by the union of the default settings above and whatever the user provides when instantiating. I think this is a bit nicer as if you need to add new parameters to these dicts, you don't have to modify these functions to do so, just the defaults at the top of the file. I've also added some error checking here rather than relying on any `KeyError` that will occur in e.g. `set_up_solvers()`. 

Also, I don't know if this is intended, but `reini_kwargs["frequency"]` will not be set if `self.mesh.cartesian` is False and the user doesn't provide it. Should this key be mandatory if `self.mesh.cartesian` is False?

dsroberts (commented):
I have some thoughts on the implementation of default settings for `adv_kwargs` and `reini_kwargs`. I haven't put the proposed changes through CI yet, so they may not work as-is. The rest of the changes look good to me.

gadopt/level_set_tools.py
Patol75: Thank you, Dale. These changes make the code more readable!

gadopt/level_set_tools.py
Patol75: Thank you for pointing out the non-cartesian case. I have not experimented with simulations in that case, but true, I also did not provide a fall-back value. This is now fixed so such simulations can still run.

gadopt/level_set_tools.py
```diff
@@ -214,49 +230,26 @@ def __init__(
...
 
         if isinstance(adv_kwargs, dict):
-            self.advection = True
-            self.adv_kwargs = adv_kwargs
+            if not all(param in adv_kwargs for param in ["u", "timestep"]):
+                raise KeyError("'u' and 'timestep' must be present in 'adv_kwargs'")
 
-            self.set_default_advection_args()
+            self.advection = True
+            self.adv_kwargs = adv_params_default | adv_kwargs
```
dsroberts, line 442: Here, and on line 246, the `solver_params` key is not treated separately, meaning that anything a user places in `adv_kwargs['solver_params']` will completely clobber the default solver params dict. In my initial suggestion, `adv_kwargs['solver_params']` would add to the default solver params, only overwriting keys that matched. If its more appropriate to replace the `solver_params` dict rather than add to it, I'm happy to approve this PR.

Patol75: Thank you for raising this. I think it is fine, given that we are only setting `pc_type` and `sub_pc_type`. Something to consider is that someone could want to specify a different `pc_type` and not provide `sub_pc_type`. That could lead to incompatibility issues. Happy to go ahead.

dsroberts (approved):

PR #186, opened 2025-03-10
Level-set initialisation API

Patol75:
Extracts changes from https://github.com/g-adopt/g-adopt/pull/128 related only to level-set initialisation.


Patol75:
The commit above demonstrates why CI returned a failure for the Schmalholz benchmark. These changes are helpful to reach slab detachment, as visualised on ParaView. In that regard, are we happy to update that benchmark here or should we wait for a later PR?

I have not investigated why the Schmeling case fails, but given the new `xfail` status, CI should return a green tick.

gadopt/level_set_tools.py
```diff
@@ -126,6 +129,218 @@ def thermal_diffusivity(cls, *args, **kwargs):
...
+
+    Returns:
+      A NumPy array holding the conservative level-set node values
+    """
+    if isinstance(epsilon, fd.Function):
+        epsilon = epsilon.dat.data
+
+    return (1 + np.tanh(np.asarray(signed_distance) / 2 / epsilon)) / 2
+
+
```
dsroberts, line 343: What's the rationale for having this as a separate function? From what I can see in the demos, the `signed_distance` list is neither used nor modified before being passed into this function. I also think its a little awkward to have to pass `level_set` into `signed_distance`, and then have to go and modify `level_set.dat.data[]` directly afterwards. It seems to me you could add `epsilon` to the list of args into `signed_distance`, have it do the final `(1 + np.tanh...)` bit and modify `level_set` in-place. Then this function wouldn't be needed any more.

dsroberts (commented):
Only one issue from my point of view. I think there is potential to streamline the user interface a little more. I recall we talked about this once, but I don't remember the outcome of that discussion.

gadopt/level_set_tools.py
Patol75: Thank you, that makes a lot of sense. Implementing something along these lines.

tests/multi_material/run_benchmark.py
```diff
@@ -119,26 +119,26 @@ def write_output(output_file):
     Simulation.initialise_temperature(temperature)
 
     # Set up function spaces and functions used in the level-set approach
-    func_space_ls = fd.FunctionSpace(mesh, "DQ", Simulation.level_set_func_space_deg)
+    func_space_ls = fd.FunctionSpace(mesh, "DQ", 2)
```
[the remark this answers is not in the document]
Patol75: Probably not. Thank you.

gadopt/level_set_tools.py
```diff
@@ -126,6 +131,207 @@ def thermal_diffusivity(cls, *args, **kwargs):
...
+    level_set: fd.Function,
+    epsilon: float | fd.Function,
+    /,
+    interface_geometry: str,
+    interface_coordinates: list[list[float]] | list[list[float], float] | None = None,
+    *,
+    interface_callable: Callable | str | None = None,
+    interface_args: tuple[Any] | None = None,
+    boundary_coordinates: list[list[float]] | np.ndarray | None = None,
+) -> list:
```
dsroberts, line 164: This new function has no return value, so type annotation should be updated accordingly.

dsroberts (approved):
Looks good to me.

PR #189, opened 2025-03-11
Pack boundary conditions and solver options in `StokesSolver` methods

Patol75:
This PR contains the first set of changes transitioning the current Stokes integrators module towards the one proposed in gh-158. The end goal is to inherit solver classes from an abstract base class, and the methods defined here will become part of the abstract base class.

### DIFF
--- a/gadopt/stokes_integrators.py
+++ b/gadopt/stokes_integrators.py
@@ -295,9 +295,6 @@ class StokesSolver:
 
     def set_boundary_conditions(self) -> None:
         """Sets strong and weak boundary conditions."""
-        self.strong_bcs = []
-        self.weak_bcs = {}
-
         bc_map = {"u": self.solution_space.sub(0)}
         if self.mesh.cartesian:
             bc_map["ux"] = bc_map["u"].sub(0)
--- a/gadopt/stokes_integrators.py
+++ b/gadopt/stokes_integrators.py
@@ -5,6 +5,7 @@ instantiate the `StokesSolver` class by providing relevant parameters and call t
 
 """
 
+from collections import defaultdict
 from numbers import Number
 from typing import Optional
 
@@ -224,13 +225,15 @@ class StokesSolver:
         free_surface_theta: float = 0.5,
         **kwargs,
     ):
-        self.Z = z.function_space()
-        self.mesh = self.Z.mesh()
-        self.test = fd.TestFunctions(self.Z)
+        self.solution_space = z.function_space()
+        self.mesh = self.solution_space.mesh()
+        self.test = fd.TestFunctions(self.solution_space)
         self.solution = z
         self.T = T
         self.approximation = approximation
+        self.bcs = bcs
         self.quad_degree = quad_degree
+        self.solver_parameters = solver_parameters
 
         self.J = J
         self.constant_jacobian = constant_jacobian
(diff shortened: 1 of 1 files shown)


gadopt/stokes_integrators.py
```diff
@@ -308,52 +288,102 @@ def __init__(
...
+                        self.free_surface = True
                     case _:
-                        raise ValueError(
-                            f"Solver type '{solver_parameters}' not implemented."
-                        )
-            elif self.mesh.topological_dimension() == 2 and self.mesh.cartesian:
-                self.solver_parameters.update(direct_stokes_solver_parameters)
-            else:
-                self.solver_parameters.update(iterative_stokes_solver_parameters)
+                        weak_bc[bc_type] += val
```
w3168, line 330: Does this work because it is a default dict?

Patol75: Correct

gadopt/stokes_integrators.py
```diff
@@ -308,52 +288,102 @@ def __init__(
...
-                            iterative_stokes_solver_parameters
+        self.set_solver_options()
+
+        # solver object is set up later to permit editing default solver parameters specified above
+        self._solver_setup = False
+
+    def set_boundary_conditions(self) -> None:
+        """Sets strong and weak boundary conditions."""
+        self.strong_bcs = []
+        self.weak_bcs = {}
```
dsroberts, line 299: These variables are initialised twice. Here and on L247-248. My preference would be to keep the initialisation in `__init__` and remove it from here.

dsroberts (commented):
Looks good. I only have one minor comment.

gadopt/stokes_integrators.py
angus-g: I don't really understand why this needs to be a default dict? There can only be one `bc_type` key with a given value for a given set of boundary conditions... Am I missing something?

Patol75: When one sets up a free surface, the weak boundary condition ends up being `normal_stress`. In some cases, one could set up a free surface and also specify a load on the free surface. In that case, we would end up with two values for the `normal_stress` weak boundary condition.

dsroberts: It made sense to me as its accounting for whether the boundary condition is present or not. If there are no bcs, it remains an empty dict. I think its nicer than setting it to zero at the start and seeing if its still zero at the end, or explicitly checking on the existence of a key.

angus-g: Okay, I guess I missed the context of this planning for a future change to the free surface/weak bc interaction. As it is, this change makes no difference (the check + add you're both talking about is still present), but it will be changed later.

Patol75: Sorry about the missing context; one current example in master where what I described above happens is [here](https://github.com/g-adopt/g-adopt/blob/3ebbda08b30a22d1c4942af643ea1342bfff95b2/demos/glacial_isostatic_adjustment/base_case/base_case.py#L416).

angus-g: Ah yeah, I understand the motivation for doing it, I'm just saying that in this PR, it is still unnecessary to use the defaultdict because the code in `setup_free_surface` that accesses the weak BCs is unmodified, which is why I was confused:

https://github.com/g-adopt/g-adopt/blob/9c0579542007f73ba3df432f233619f5cd8db0ea/gadopt/stokes_integrators.py#L425-L431

Patol75: Very true! The joys of pulling changes apart. :D

Patol75:
Any objection to merging this one in?

angus-g (approved):
I think we're good to go here!

PR #195, opened 2025-03-21
Adapt to Firedrake root containers

angus-g:
The Firedrake container now runs as root by default, which makes git complain about safe directories. Set all directories as safe, remove the `$HOME` hack (because Firedrake was installed to `$HOME/.local`, now in `/opt`) and install tsp directly into `/usr/local`. Hopefully gets CI running again...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,7 +86,7 @@ jobs:
 
       - name: Put task spooler on PATH
         run: |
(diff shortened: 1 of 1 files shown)


dsroberts (approved):
Looks good to me. A reminder for everyone (including myself) that any new PRs will need to have incorporated this change to run the tests correctly.

PR #196, opened 2025-03-23
Adapt to Firedrake root containers some more

dsroberts:
There are some additional settings in the Firedrake build/test github action that are needed to get root containers working here as well.


dsroberts:
OK, I think I see what's going on here. The in-container root user needs to be remapped to whatever the firedrake user was before in order to access /__w, which is mapped in from the host. @angus-g I think this is up to you...

angus-g:
Should be working now!

PR #201, opened 2025-04-02
Remove unnecessary global scope from test_spiegelman

angus-g:
The conf dictionary is always copied before modification, so it doesn't need to be global in the test scope. Will fix our newly developed lint failure.


sghelichkhani (approved):
I was just wondering where that linting error started to develop that I saw this! Thanks Angus!

PR #202, opened 2025-04-03
Update free surface setup in `StokesSolver` using new `set_equations` and `set_form` methods

Patol75:
This PR attempts to streamline the free surface implementation in `StokesSolver`. In doing so, it introduces two new methods (`set_equations` and `set_form`) that deal with setting up `Equation` instances and calculating the weak form based on the theta scheme.

### DIFF
--- a/demos/mantle_convection/free_surface/free_surface.py
+++ b/demos/mantle_convection/free_surface/free_surface.py
@@ -128,7 +128,7 @@ Bfs = 10.0  # Free surface buoyancy number
 
 stokes_bcs = {
     boundary.bottom: {"uy": 0},
-    boundary.top: {"free_surface": {"eta_index": 2, "RaFS": Ra * Bfs}},
+    boundary.top: {"free_surface": {"RaFS": Ra * Bfs}},
     boundary.left: {"ux": 0},
     boundary.right: {"ux": 0},
 }
--- a/demos/mantle_convection/free_surface/free_surface.py
+++ b/demos/mantle_convection/free_surface/free_surface.py
@@ -124,11 +124,11 @@ T.interpolate((1.0 - X[1]) + (0.05 * cos(pi * X[0]) * sin(pi * X[1])))
 # Typically, when the exterior density is negligbile compared with the interior density (i.e. air vs rock!) and with $\alpha$ = 3x10$^{-5}$ K$^{-1}$ and $\Delta T$ = 3000 K, $B_{fs}$ is about 10. For dimensional simulations you can specify the density contrast across the free surface with *delta_rho_fs* through the free surface dictionary, as well as providing $\rho_0$ and $g$ through the approximation object in the usual way.
 
 # +
-Bfs = 10.  # Free surface buoyancy number
+Bfs = 10.0  # Free surface buoyancy number
 
 stokes_bcs = {
     boundary.bottom: {"uy": 0},
-    boundary.top: {"free_surface": {"RaFS": Ra*Bfs}},
+    boundary.top: {"free_surface": {"eta_index": 2, "RaFS": Ra * Bfs}},
     boundary.left: {"ux": 0},
     boundary.right: {"ux": 0},
 }
@@ -171,7 +171,7 @@ energy_solver = EnergySolver(
 )
 
 stokes_solver = StokesSolver(
-    z, T, approximation, bcs=stokes_bcs, constant_jacobian=False, free_surface_dt=delta_t
+    z, T, approximation, bcs=stokes_bcs, constant_jacobian=False, coupled_tstep=delta_t
 )
 
 # -
--- a/gadopt/approximations.py
+++ b/gadopt/approximations.py
@@ -241,13 +241,13 @@ class BoussinesqApproximation(BaseApproximation):
         return self.rho * self.H
 
     def free_surface_terms(
-        self, p, T, eta, theta_fs, *, variable_rho_fs=True, RaFS=1, delta_rho_fs=1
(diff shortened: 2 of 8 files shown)


Patol75:
I very slightly broke the expected errors for the iterative free surface tests. This is annoying because they are not broken on `Patol75/backend_changes`. Let's hope I can figure out why.

w3168:
> I very slightly broke the expected errors for the iterative free surface tests. This is annoying because they are not broken on `Patol75/backend_changes`. Let's hope I can figure out why.

I havent completely tracked it down yet but I think it might be that the scaling factor -theta. I think you possibly include it twice because in your branch you have taken the theta out of the prefactor term returned by approximations, but in master this is still in.

I got the same values as before by commenting out the scaling factor in stokes_integrators (and the assert). But I guess you want to shuffle around where the theta factor gets defined? 

Are we sure that the theta scaling really should be an assert btw? Seeing as though we didnt get second order convergence for the more general case with the internal variable perhaps we shouldnt hard code something like this in now...

Patol75:
Bang on! Oh man, it would have taken me ages to spot this. Thank you!

With the current changes, you can just remove `theta_fs` from the approximation and recover the expected behaviour. I guess my downfall was to leave `approximations.py` untouched!

For the scaling and assert, let me think about it a bit more, and I will come back to you.

Patol75:
> Are we sure that the theta scaling really should be an assert btw? Seeing as though we didnt get second order convergence for the more general case with the internal variable perhaps we shouldnt hard code something like this in now...

My understanding was that, to correctly apply the theta scheme, you need to ensure that all equations in the coupled system that have a mass term are scaled by `-theta`. Doing this separates the theta scheme from the actual definition of equation terms, which I feel is desirable.

w3168:
> > Are we sure that the theta scaling really should be an assert btw? Seeing as though we didnt get second order convergence for the more general case with the internal variable perhaps we shouldnt hard code something like this in now...
> 
> My understanding was that, to correctly apply the theta scheme, you need to ensure that all equations in the coupled system that have a mass term are scaled by `-theta`. Doing this separates the theta scheme from the actual definition of equation terms, which I feel is desirable.

I agree that it makes sense based on how we understood the free surface implementation. Hopefully we can use this to get to the bottom of why the internal variable didnt show second order convergence

Patol75:
@w3168 @angus-g We should try to merge this PR next. I have resolved conflicts and am now waiting on CI. If the green tick stays, are you happy for the changes to go in?

w3168:
> @w3168 @angus-g We should try to merge this PR next. I have resolved conflicts and am now waiting on CI. If the green tick stays, are you happy for the changes to go in?

I'm still happy if it passes the CI!

demos/mantle_convection/free_surface/free_surface.py
```diff
@@ -124,11 +124,11 @@
 # Typically, when the exterior density is negligbile compared with the interior density (i.e. air vs rock!) and with $\alpha$ = 3x10$^{-5}$ K$^{-1}$ and $\Delta T$ = 3000 K, $B_{fs}$ is about 10. For dimensional simulations you can specify the density contrast across the free surface with *delta_rho_fs* through the free surface dictionary, as well as providing $\rho_0$ and $g$ through the approximation object in the usual way.
 
 # +
-Bfs = 10.  # Free surface buoyancy number
+Bfs = 10.0  # Free surface buoyancy number
 
 stokes_bcs = {
     boundary.bottom: {"uy": 0},
-    boundary.top: {"free_surface": {"RaFS": Ra*Bfs}},
+    boundary.top: {"free_surface": {"eta_index": 2, "RaFS": Ra * Bfs}},
```
angus-g, line 131: I haven't had time to look at this, but I'm not sure I like the ergonomics of this `eta_index` mechanism... Let me have a proper look and think first though.

drhodrid:
In the interests of getting things moving again, @angus-g are you happy to look at this again? @w3168 has approved, but there is on outstanding item between yourself and @Patol75.

demos/mantle_convection/free_surface/free_surface.py
angus-g: Thanks! I disagree that it's fragile to rely on insertion order: it's an explicit feature of Python 3.7+ and we're well past that (we should probably nuke the comments about that...) It is true that you need to be a bit more careful setting things up with multiple free surface BCs, true.

With that in mind, my preference would be not to introduce `eta_index` here. I definitely agree that *something* is needed, particularly when internal variables are introduced, but I don't think it's time just yet.

Patol75: No wait, you are missing the point. I am not talking about the dictionary insertion order (I agree with your conclusion on that). I am talking about the fact that the insertion order in the dictionary has to match the order of the functions in the mixed space.

Patol75: Perhaps another way to deal with that is to add a boundary attribute to each free-surface function in the same way that we add the cartesian attribute to the mesh?

angus-g: They're both good ideas, and possibilities for sure. My preference is to make sure we think about this properly, so we get a clean and ergonomic interface. It's not in scope for this PR in particular.

Patol75: Given that there is more than one possibility now, I agree that the way forward is to leave this change out and open an issue mentioning the problem and potential solutions.

Patol75:
Provided CI returns a green tick, this PR should be good to go.

PR #203, opened 2025-04-03
Add unit test for solver parameter consistency

angus-g:
In response to a recent Firedrake regression where solver parameters weren't being passed through to blocks on the tape. This is a bit brute force, but we ensure that our artificial "gadopt_test" option is processed by the options database when we play the tape forward by evaluating the reduced functional.

I think maybe the implementation here is a bit ugly and certainly happy to take a different approach!

### DIFF
--- a/tests/unit/data/solver_params.py
+++ b/tests/unit/data/solver_params.py
@@ -1,0 +1,39 @@
+from firedrake import *
+from firedrake.adjoint import *
+
+continue_annotation()
+
+solver_parameters = {
+    "snes_type": "newtonls",
+    "snes_converged_reason": None,
+    "snes_max_it": 50,
+    "ksp_type": "cg",
+    "pc_type": "hypre",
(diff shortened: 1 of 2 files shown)


Patol75:
A suggestion without much solid basis to it: could you potentially get that information using one of the [callback arguments](https://pyadjoint.org/documentation/pyadjoint_api.html#pyadjoint.ReducedFunctional) to `ReducedFunctional`?

sghelichkhani:
@Patol75 Not ReducedFunctional, but you can get it from the corresponding block on tape.

My idea originally was to make sure `get_working_tape().get_blocks()[1,3].forward_kwargs` are the same as solver_parameters. But maybe it's better to do it this way. The point is that we would still catch issues if the solver is set, but not being used somehow. 

Thanks @angus-g!

sghelichkhani (approved):
Well done! Thanks

PR #204, opened 2025-04-03
Update `setup_solver` in `StokesSolver` to make `nullspace` arguments explicit

Patol75:
This PR makes nullspace arguments to variational solvers explicit in `StokesSolver`, i.e. they are now their own documented argument instead of being silently injected through `kwargs`.

### DIFF
--- a/demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
+++ b/demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
@@ -442,10 +442,20 @@ Z_near_nullspace = create_stokes_nullspace(Z, closed=False, rotational=True, tra
 # along with the approximation, timestep and boundary conditions.
 #
 
-stokes_solver = ViscoelasticStokesSolver(z, stress_old, displacement, approximation,
-                                         dt, bcs=stokes_bcs, constant_jacobian=True,
-                                         nullspace=Z_nullspace, transpose_nullspace=Z_nullspace,
-                                         near_nullspace=Z_near_nullspace)
+stokes_solver = ViscoelasticStokesSolver(
+    z,
+    stress_old,
+    displacement,
+    approximation,
+    dt,
+    bcs=stokes_bcs,
+    constant_jacobian=True,
+    nullspace={
+        "nullspace": Z_nullspace,
+        "transpose_nullspace": Z_nullspace,
+        "near_nullspace": Z_near_nullspace,
+    },
+)
 
 # We next set up our output, in VTK format. This format can be read by programs like pyvista and Paraview.
 
--- a/gadopt/stokes_integrators.py
+++ b/gadopt/stokes_integrators.py
@@ -205,6 +205,8 @@ class StokesSolver:
       free_surface_theta: Timestepping prefactor for free surface equation, where
                           theta = 0: Forward Euler, theta = 0.5: Crank-Nicolson (default),
                           or theta = 1: Backward Euler
+      nullspace:
+        Dictionary of nullspace options, including transpose and near nullspaces
 
     """
 
@@ -222,21 +224,20 @@ class StokesSolver:
         constant_jacobian: bool = False,
         free_surface_dt: Optional[float] = None,
         free_surface_theta: float = 0.5,
-        **kwargs,
+        nullspace: dict[str, fd.MixedVectorSpaceBasis] = {},
     ):
-        self.Z = z.function_space()
-        self.mesh = self.Z.mesh()
-        self.test = fd.TestFunctions(self.Z)
+        self.solution_space = z.function_space()
+        self.mesh = self.solution_space.mesh()
+        self.test = fd.TestFunctions(self.solution_space)
         self.solution = z
         self.T = T
         self.approximation = approximation
         self.quad_degree = quad_degree
+        self.nullspace = nullspace
 
         self.J = J
         self.constant_jacobian = constant_jacobian
         self.linear = not depends_on(self.approximation.mu, self.solution)
-
-        self.solver_kwargs = kwargs
 
         self.k = upward_normal(self.mesh)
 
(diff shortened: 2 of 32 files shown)


dsroberts:
Very in favour of this. It took me half a day to figure out that the nullspaces being silently passed from the top-level demo down to NonlinearVariationalSolver. I'm sure it would have been a no-brainer to anyone familiar with these sorts of physical problems, but to a smooth-brained software engineer like me that can't see past the code, this will be very helpful.

Patol75:
I am pretty sure I had the same experience months ago. I remember asking myself: where does the nullspace go? :D

w3168:
I agree this is very useful! I think passing the nullspace/transpose/nearnullspace etc as a dict is nice, but... would it be better though to keep the api the same?

Patol75:
The only problem I have with passing all nullspace arguments independently is that it adds three arguments to the solver instead of one. That solver already has a number of arguments, so it is perhaps not the best to make the list grow significantly.

dsroberts:
@Patol75 but they already have to be passed to `StokesSolver`. Doesn't change anything from the users point-of-view. I don't think there is a limit on the number of arguments a constructor can take...

Patol75:
True. I do not have a strong opinion then.

Patol75:
Now that `StokesSolver` does not accept additional keyword arguments, CI picked up that we forgot to remove the `mu` argument in one of the `unit` tests.

@angus-g @drhodrid @sghelichkhani Any opinions on the API for null spaces? Dictionary or individual arguments?

gadopt/stokes_integrators.py
```diff
@@ -205,6 +205,8 @@ class StokesSolver:
       free_surface_theta: Timestepping prefactor for free surface equation, where
                           theta = 0: Forward Euler, theta = 0.5: Crank-Nicolson (default),
                           or theta = 1: Backward Euler
+      nullspace:
+        Dictionary of nullspace options, including transpose and near nullspaces
```
angus-g, line 209: I do like that this is explicit now. I'm as yet undecided on the API change, but I guess this kind of technically *reduces* the number of arguments to init, which I think is better.

On this comment, maybe explicitly state the supported nullspaces (`nullspace`, `transpose_nullspace`, `near_nullspace`) and that they can be created with `create_stokes_nullspace`? Or not :shrug:, I don't know the discoverability of this vs. starting with a demo...

Maybe (for later) we could have an abstraction over the interaction `create_stokes_nullspace` and this parameter. Just thinking out loud now!

Patol75: I agree with the need for a more detailed docstring. As an example, [this](https://github.com/g-adopt/g-adopt/blob/aaea809421a3d7cc350888eec9a1047d885763ca/gadopt/stokes_integrators.py#L126-L175) is what I intend to suggest for the coming ABC. Adding another table for null spaces seems a reasonable option.

Regarding `create_stokes_nullspace`, I thought a few times about hiding it from the user, but I never got anywhere I liked. The main issue lies with compressing the arguments that function needs, but I feel like it is not necessarily a great idea.

Patol75:
The last CI failure is unrelated and will be fixed once the updated level-set API is merged.

demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -442,10 +442,20 @@ def setup_heterogenous_viscosity(viscosity):
...
-                                         near_nullspace=Z_near_nullspace)
+stokes_solver = ViscoelasticStokesSolver(
+    z,
+    stress_old,
+    displacement,
+    approximation,
+    dt,
+    bcs=stokes_bcs,
+    constant_jacobian=True,
+    nullspace={
```
drhodrid, line 453: One comment from me here which I think is important: I don't like the idea of combining nullspaces and near nullspaces in the same dictionary, as they are very different things (one has a physical meaning, the other is strictly used in geometric multigrid). They should be separate.

Patol75: I would say null spaces have a mathematical meaning: they represent the kernel of the operator or operator's transpose. Sure, you can then deduce some physical implications based on the problem at hand, but they remain quite mathematical in essence. For the near null space, yes, the concept is not exactly the same: we are looking for additional vector spaces that approximate the kernel of the operator. In other words, additional modes for which the operator's image can be considered sufficiently small to influence kernel-sensitive operators. I may be wrong, but that feels close enough to me to pack these three options together.

sghelichkhani: Just chiming in here to say that for practical reason I believe they should separate too. Nullspaces are intrinsic to a PDE system, near-nullspace are just related to one specific preconditioner.

Patol75: All sensible here. Looking at Firedrake's source code for [`NonlinearVariationalSolver`](https://www.firedrakeproject.org/_modules/firedrake/variational_solver.html), they do separate all the null space arguments when instantiating the solver, but if you look at the docstring and the `set_nullspace` method, they tend to group them. So from my perspective, it makes little sense to have a group of 2 and a single one, so either we stay with 3 separate arguments or we accept grouping them all. Happy either way, as long as they are explicitly referenced, which is currently not the case.

drhodrid: Ok sounds like have a plan. No grouping, but explicitly referenced! Go go go!

Patol75:
Provided CI returns a green tick, this PR should be good to go.

gadopt/stokes_integrators.py
```diff
@@ -322,6 +334,9 @@ def set_solver_options(self) -> None:
         self.appctx = {
             "mu": self.approximation.mu / self.approximation.rho_continuity()
         }
+        if self.free_surface_map:
+            self.appctx["free_surface"] = self.free_surface_map
+            self.appctx["ds"] = self.equations[-1].ds
```
angus-g, line 339: Why did this move?

Patol75: True, maybe too early. When using an ABC, there is no reason to have this condition for all solvers when it is only used by `StokesSolver`. As a result, it makes more sense to have it in solver options than setup, as it is easier to overwrite that former method. Should I revert for now?

angus-g: It would be good to keep the PR within its scope, and not include this. If/when it's needed it can go in.

angus-g (approved):

PR #212, opened 2025-04-26
Diffusion Constant in pygplates connector

sghelichkhani:
Diffusion is a constant that is used in non-dimensionalising time, and velocities when using pygplates Connector. This PR allows setting kappa. The default value is kept 1.0 so that we not need an interface change or any other change to demos.


drhodrid (approved):
This looks good to me - provided CI goes green, this is good to go.

PR #213, opened 2025-04-29
Update expected test results for PETSc version bump

angus-g:
Firedrake recently bumped to PETSc 3.23.0, which has introduced a few test failures.

The first of these is the `implicit-cylindrical` variant of the free surface tests. This has got pretty tight tolerances and has shifted just slightly outside these. I've updated the expected results:

dt factor | 2 | 1 | 0.5 | 0.25
---- | ---- | --- | ---- | ----
old | 3.86852488e-10 |8.91383896e-11 | 2.11316015e-11 | 5.08922311e-12
new | 3.86852477e-10 | 8.91384378e-11 | 2.11355594e-11 | 5.08755227e-12
diff (abs) | 1.08353995e-17 | -4.82486944e-17 | -3.95786403e-15 |  1.67084368e-15
diff (rel) | 2.80091245e-08 | -5.41278507e-07 | -1.87295980e-04 |  3.28310166e-04

The unresolved problem at this point is with the Schmeling (2008) multi-material benchmark. This seems to have destabilised somewhat. The test asserts:

```python3
abs(output_time[slab_tip_depth >= 600).min() - 43) <= 0.1
```

From some of my various testing runs today, all the values are significantly above 0.1:

```
0.20074226785175853
0.19789304945984298
0.24344459596927948
0.19521094028348074
```

I might need @Patol75's input for this test case!


w3168:
Thanks Angus for checking the free surface part!

Patol75:
Hey @angus-g,
Given all the upcoming changes to the level-set part of G-ADOPT'S API (see open PRs), all the tests are going to be modified at least slightly. As a result, what we do here is not very important from my perspective. Bumping this 0.1 tolerance to 1.0 seems easy enough to get a green tick.

angus-g:
I believe I have figured out the root cause of this: the change to use more solvers from system packages on the Firedrake side (https://github.com/firedrakeproject/firedrake/commit/c48cadfb503ddfcb9ff5a7ffa28605e2fc5dc95e). Nothing substantial has changed from a PETSc standpoint. As shown below though, everything (except metis) is an older version than previously!

package | PETSc-installed version | Ubuntu-installed version
--- | --- | ---
MUMPS | 5.7.3 | 5.6.2
metis | v5.1.0-p12 | 5.1.0
PTScotch | 7.0.6 | 7.0.4
scalapack | 2.2.2 | 2.2.1
SuiteSparse | 7.10.2 | 7.6.1
superLU | 7.0.0 | 6.0.1

Testing each of these in isolation, using the system version of PTscotch seems to be the issue. I guess this means that we don't have to worry about anything fundamentally changing in PETSc, and it's just some partitioning instability caused by the different versions. So maybe the way forward here is to merge this PR, and accept that OS updates could cause answer changes outside of the Firedrake/PETSc ecosystem?

drhodrid:
Hi Angus,

I suspect it is worth first making the Firedrake dev’s aware of the impact of this change. I very much doubt their intention was to revert back to older versions (some significantly so!) of several of these libraries by default… unless, of course I am missing something…

Happy for this to be merged for now, but let’s add it to the discussion items for next week’s dev meeting.

Best wishes,

Rhod

On 8 May 2025, at 7:14 am, Angus Gibson ***@***.***> wrote:

[https://avatars.githubusercontent.com/u/11085886?s=20&v=4]angus-g left a comment (g-adopt/g-adopt#213)<https://github.com/g-adopt/g-adopt/pull/213#issuecomment-2861895882>

I believe I have figured out the root cause of this: the change to use more solvers from system packages on the Firedrake side ***@***.***<https://github.com/firedrakeproject/firedrake/commit/c48cadfb503ddfcb9ff5a7ffa28605e2fc5dc95e>). Nothing substantial has changed from a PETSc standpoint. As shown below though, everything (except metis) is an older version than previously!

package PETSc-installed version Ubuntu-installed version
MUMPS   5.7.3   5.6.2
metis   v5.1.0-p12      5.1.0
PTScotch        7.0.6   7.0.4
scalapack       2.2.2   2.2.1
SuiteSparse     7.10.2  7.6.1
superLU 7.0.0   6.0.1

Testing each of these in isolation, using the system version of PTscotch seems to be the issue. I guess this means that we don't have to worry about anything fundamentally changing in PETSc, and it's just some partitioning instability caused by the different versions. So maybe the way forward here is to merge this PR, and accept that OS updates could cause answer changes outside of the Firedrake/PETSc ecosystem?

Patol75:
I have a feeling that was their intention. When you decide to go with system packages, you know you will build against earlier package versions. One question here is why our numbers did not change in the first place when those packages moved to newer versions. Or did they?

drhodrid:
In the interests of getting us going again on pull requests, this needs to go in.

drhodrid (approved):

PR #214, opened 2025-04-29
Fix render workflow

angus-g:
There were a couple of outdated constructs in the render workflow that don't apply with the Firedrake root container. The real issue here is that I _think_ we need to explicitly specify that we have a `gadopt.gplates` subpackage, because we're not using setuptools' automatic package discovery. It's kind of weird that this only sometimes shows up on the render job, and it doesn't ever cause an issue on the testing job. I think this should probably wait for (or be combined with) #213 so that we don't get spurious test failures.

### DIFF
--- a/.github/workflows/render.yml
+++ b/.github/workflows/render.yml
@@ -55,8 +55,6 @@ jobs:
 
       - name: Set up for network namespacing
         run: |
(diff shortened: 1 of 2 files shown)


sghelichkhani (approved):
Thanks @angus-g! Good to have demos being built again!

Issue #215, opened 2025-05-13
Inconsistent partition leading to failures on free surface test


angus-g:
I thought we could just fix this in #213, but it seems like maybe the change to PTScotch is causing inconsistent partitioning changes, giving us intermittent failures? An example failure is
```
actual:
array([3.86852477e-10, 8.91384378e-11, 2.11363394e-11, 5.08755227e-12])

expected:
array([3.86852477e-10, 8.91384378e-11, 2.11355594e-11, 5.08755227e-12])

relative error:
array([0.000000e+00, 0.000000e+00, 3.690463e-05, 0.000000e+00])
```

So it's just one of the dt factors that's causing trouble. Is there any direction you'd like to go with this @w3168?

w3168:
I wonder if we make the iterative solver tolerances tighter just for this case it might still be consistent with the expect values (to rtol=1e-6...)

angus-g:
Looks like perhaps not: https://github.com/g-adopt/g-adopt/actions/runs/14988028878. I might dig into the log files and see if there's an iteration difference or anything obvious like that.

w3168:
yeah... Is there any chance that by making the tolerances tighter we now have a new expected set and this should always converge to this regardless of the partitioning? (and if not do we need to refine the tolerance again...)

w3168:
fixed by #216

PR #216, opened 2025-05-14
Use tighter tolerances for implicit cylindrical free surface

angus-g:
As pointed out in #215, the test has a fairly tight tolerance, but the default iterative solver tolerances are looser than this. Maybe this is a way around the partitioning issue? This did run consistently in three successive tests: [one](https://github.com/g-adopt/g-adopt/actions/runs/14988028878), [two](https://github.com/g-adopt/g-adopt/actions/runs/15007221068), [three](https://github.com/g-adopt/g-adopt/actions/runs/15007221835)


drhodrid:
Looks good to me!

w3168 (approved):

PR #220, opened 2025-05-27
Silence memory binding warnings in CI

dsroberts:
These warnings are meaningless as a) no memory binding policy is set, so as far as I can tell its failing to not set memory binding and b) memory binding is useless on this system as it only has a single NUMA node. The warning is spurious and therefore can be safely silenced.


dsroberts:
Didn't work. Will keep investigating

PR #221, opened 2025-05-28
Add memory profiling to CI runs

dsroberts:
Set tsp version via. repository variable, start memprof function when jobs start.


angus-g:
Nice, seems to just work. For day-to-day, is it worth adding something like a Max RSS column to the GH run summary output? I had a quick glance at the tsp code and it's not an immediately straightforward change that I can see.

I'll look into doing a last-N workflow summary now that everything we should need is in the DB artifact.

dsroberts:
OK, I'm happy with how the updated tsp is operating. Once we've confirmed memory usage from this last run is present and sensible, this is good to merge.

PR #222, opened 2025-06-03
Tests for adjoint with changing Dirichlet boundary condition

sghelichkhani:
This introduces a changing BC test to the adjoint test in a box. In addition we also introduce a None scheduler (not using any scheduling in adjoint) and disk storage scheduling for the same case (going from 4cases x 1scheduler = 4total to  5cases x 3scheduler=15total cases). 

**Note** With the current bug in pyadjoint, the additional tests should fail on firedrake 2025.04.post0

### DIFF
--- a/tests/adjoint/Makefile
+++ b/tests/adjoint/Makefile
@@ -1,7 +1,7 @@
 include ../../rules.mk
 
-cases := damping smoothing Tobs uobs
-scheduler := fullmemory
+cases := damping smoothing Tobs uobs uimposed
+scheduler := noscheduler fullmemory fullstorage
 
 cases_schedulers := $(foreach case,$(cases),$(foreach sched,$(scheduler),$(case)_$(sched)))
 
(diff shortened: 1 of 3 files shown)


sghelichkhani:
@drhodrid @angus-g Not sure what the best strategy here would be. To merge these and see when the `2025.04.post1` lands, or wait until that lands and then merge!

angus-g:
> @drhodrid @angus-g Not sure what the best strategy here would be. To merge these and see when the `2025.04.post1` lands, or wait until that lands and then merge!

Given this will explicitly fail until a there's a new release, it doesn't really make sense to merge it yet. Of course, it would be good to demonstrate that it _does_ fail on 2024.04.post0 but pass on master?

I've submitted https://github.com/firedrakeproject/firedrake/pull/4351 to see if we can backport the changes to the next release. Otherwise, there'll hopefully be a new release soon when Firedrake updates to pull in the latest PETSc changes.

angus-g:
Awesome, thanks @connorjward!

angus-g (approved):

PR #225, opened 2025-06-03
Base class for coupled momentum solvers

Patol75:
Work in progress

### DIFF
--- a/gadopt/stokes_integrators.py
+++ b/gadopt/stokes_integrators.py
@@ -5,9 +5,10 @@ instantiate the `StokesSolver` class by providing relevant parameters and call t
 
 """
 
+import abc
 from collections import defaultdict
 from numbers import Number
-from typing import Optional
+from typing import Any, Optional
 from warnings import warn
 
 import firedrake as fd
@@ -57,7 +58,7 @@ iterative_stokes_solver_parameters = {
         "Mp_ksp_ksp_rtol": 1e-5,
         "Mp_ksp_ksp_type": "cg",
         "Mp_ksp_pc_type": "sor",
-    }
+    },
 }
 """Default iterative solver parameters for solution of stokes system.
 
@@ -198,68 +199,107 @@ def create_stokes_nullspace(
     return fd.MixedVectorSpaceBasis(Z, null_space)
 
 
-class StokesSolver:
-    """Solves the Stokes system in place.
+class MetaPostInit(abc.ABCMeta):
+    """Calls the implemented `prepare_solver` method after `__init__` returns.
 
-    Note: Null space options can be generated using `create_stokes_nullspace`.
-
-    Arguments:
-      solution: Firedrake function representing mixed Stokes system
-      T: Firedrake function representing temperature
-      approximation: Approximation describing system of equations
-      bcs: Dictionary of identifier-value pairs specifying boundary conditions
-      quad_degree: Quadrature degree. Default value is `2p + 1`, where
-                   p is the polynomial degree of the trial space
-      solver_parameters: Either a dictionary of PETSc solver parameters or a string
-                         specifying a default set of parameters defined in G-ADOPT
-      J: Firedrake function representing the Jacobian of the system
-      constant_jacobian: Whether the Jacobian of the system is constant
-      coupled_tstep: Timestep for advancing free surface equation
-      theta: Timestepping prefactor for free surface equation, where
-             theta = 0: Forward Euler, theta = 0.5: Crank-Nicolson (default),
-             or theta = 1: Backward Euler
-      nullspace:
-        A `MixedVectorSpaceBasis` for the operator's kernel
-      transpose_nullspace:
-        A `MixedVectorSpaceBasis` for the kernel of the operator's transpose
-      near_nullspace:
-        A `MixedVectorSpaceBasis` for the operator's smallest eigenmodes (e.g. rigid
-        body modes)
-
+    The implemented behaviour allows any subclass `__init__` method to first call its
+    parent class's `__init__` through super(), then execute its own code, and finally
+    call `prepare_solver`. The latter call is automatic and does not require any
+    attention from the developer or user.
     """
 
-    name = "Stokes"
+    def __call__(cls, *args, **kwargs):
+        class_instance = super().__call__(*args, **kwargs)
+        class_instance.prepare_solver()
+
+        return class_instance
+
+
+class CoupledMomentumBase(abc.ABC, metaclass=MetaPostInit):
+    """Solver for a system involving mass and momentum conservation.
+
+    ### Valid keys for boundary conditions
+    |   Condition   |  Type  |                 Description                  |
+    | :------------ | :----- | :------------------------------------------: |
+    | u             | Strong | Solution                                     |
+    | ux            | Strong | Solution along the first Cartesian axis      |
+    | uy            | Strong | Solution along the second Cartesian axis     |
+    | uz            | Strong | Solution along the third Cartesian axis      |
+    | un            | Weak   | Solution along the normal to the boundary    |
+    | stress        | Weak   | Traction across the boundary                 |
+    | normal_stress | Weak   | Stress component normal to the boundary      |
+    | free_surface  | Weak   | Free-surface characteristics of the boundary |
+
+    ### Valid keys describing the free surface boundary:
+    |     Argument     | Required |                  Description                   |
+    | :--------------- | :------: | :--------------------------------------------: |
+    | eta_index        | Yes/No   | Function index in mixed space formulation      |
+    | rho_ext          | No       | Exterior density along the free surface        |
+    | Ra_fs            | No       | Rayleigh number at the free surface            |
+    | include_buoyancy | No       | Whether the interior density includes buoyancy |
+
+    ### Classic theta values for coupled implicit time integration
+    | Theta |        Scheme         |
+    | :---- | :-------------------: |
+    | 0.5   | Crank-Nicolson method |
+    | 1.0   | Backward Euler method |
+
+    Args:
+      solution:
+        Firedrake function representing the field over the mixed Stokes space
+      approximation:
+        G-ADOPT approximation defining terms in the system of equations
+      coupled_tstep:
+        Float quantifying the time step used in a coupled time integration
+      theta:
+        Float quantifying the implicit contribution in a coupled time integration
+      bcs:
+        Dictionary specifying boundary conditions (identifier, type, and value)
+      quad_degree:
+        Integer denoting the quadrature degree
+      solver_parameters:
+        Dictionary of PETSc solver options or string matching a default set thereof
+      J:
+        Firedrake function representing the Jacobian of the mixed Stokes system
+      constant_jacobian:
+        Boolean specifying whether the Jacobian of the system is constant
+      nullspace:
+        Dictionary of nullspace options, including transpose and near nullspaces
+    """
+
+    name = "CoupledMomentum"
 
     def __init__(
         self,
         solution: fd.Function,
-        T: fd.Function,
         approximation: BaseApproximation,
-        bcs: dict[int, dict[str, Number]] = {},
-        quad_degree: int = 6,
-        solver_parameters: Optional[dict[str, str | Number] | str] = None,
-        J: Optional[fd.Function] = None,
-        constant_jacobian: bool = False,
-        coupled_tstep: Optional[float] = None,
+        /,
+        *,
+        coupled_tstep: float | None = None,
         theta: float = 0.5,
+        bcs: dict[int, dict[str, Any]] = {},
+        quad_degree: int = 6,
+        solver_parameters: dict[str, str | Number] | str | None = None,
+        J: fd.Function | None = None,
+        constant_jacobian: bool = False,
         nullspace: fd.MixedVectorSpaceBasis = None,
         transpose_nullspace: fd.MixedVectorSpaceBasis = None,
         near_nullspace: fd.MixedVectorSpaceBasis = None,
-    ):
-        self.solution_space = solution.function_space()
-        self.mesh = self.solution_space.mesh()
-        self.tests = fd.TestFunctions(self.solution_space)
+    ) -> None:
         self.solution = solution
-        self.T = T
         self.approximation = approximation
+        self.coupled_tstep = coupled_tstep
+        self.theta = theta
         self.bcs = bcs
         self.quad_degree = quad_degree
         self.solver_parameters = solver_parameters
-        self.coupled_tstep = coupled_tstep
-        self.theta = theta
+        self.J = J
+        self.constant_jacobian = constant_jacobian
         self.nullspace = nullspace
         self.transpose_nullspace = transpose_nullspace
         self.near_nullspace = near_nullspace
+
+        self.eta_ind = 2
 
         self.solution_old = solution.copy(deepcopy=True)
         self.solution_split = fd.split(solution)
@@ -269,41 +309,28 @@ class StokesSolver:
             for sol, sol_old in zip(self.solution_split, self.solution_old_split)
         ]
 
-        self.J = J
-        self.constant_jacobian = constant_jacobian
-        self.linear = not depends_on(self.approximation.mu, self.solution)
-
-        self.k = upward_normal(self.mesh)
-
-        # Setup boundary conditions
-        self.weak_bcs = {}
-        self.strong_bcs = []
-
-        # Free surface parameters
-        self.free_surface_map = {}
-        self.eta_ind = 2
-        self.buoyancy_fs = [None] * len(self.solution_split)
-
-        self.set_boundary_conditions()
-
-        # eta is a list of 0, 1 or multiple free surface fields
-        self.u, self.p, *self.eta = fd.split(self.solution)
-
-        self.rho_mass = self.approximation.rho_continuity()
+        self.solution_space = solution.function_space()
+        self.mesh = self.solution_space.mesh()
+        self.tests = fd.TestFunctions(self.solution_space)
 
         self.equations = []
         self.F = 0.0  # Weak form of the system
 
-        self.set_equations()
-        self.set_form()
-
-        self.set_solver_options()
-
         # Solver object is set up later to permit editing default solver options.
         self._solver_ready = False
 
+    def prepare_solver(self) -> None:
+        """Runs methods that set up arguments for the variational problem and solver."""
+        self.set_boundary_conditions()
+        self.set_equations()
+        self.set_form()
+        self.set_solver_options()
+
     def set_boundary_conditions(self) -> None:
         """Sets strong and weak boundary conditions."""
+        self.strong_bcs = []
+        self.weak_bcs = {}
+
         bc_map = {"u": self.solution_space.sub(0)}
         if self.mesh.cartesian:
             bc_map["ux"] = bc_map["u"].sub(0)
@@ -329,6 +356,43 @@ class StokesSolver:
                         weak_bc[bc_type] += val
 
             self.weak_bcs[bc_id] = weak_bc
+
+    def set_free_surface_boundary(
+        self, params_fs: dict[str, int | bool], bc_id: int
+    ) -> fd.ufl.algebra.Product | fd.ufl.algebra.Sum:
+        """Sets the given boundary as a free surface.
+
+        This method calculates the normal stress at the free surface boundary. In the
+        coupled approach, it also sets a zero-interior strong condition away from that
+        boundary and populates the `free_surface_map` dictionary used to calculate the
+        free-surface contribution to the momentum weak form.
+
+        Args:
+          params_fs:
+            Dictionary holding information about the free surface boundary
+          bc_id:
+            Integer representing the index of the mesh boundary
+
+        Returns:
+          UFL expression for the normal stress at the free surface boundary
+        """
+
+    @abc.abstractmethod
+    def set_equations(self):
+        """Sets Equation instances for each equation in the system.
+
+        Equations must be ordered like solutions in the mixed space.
+        """
+
+    def set_form(self) -> None:
+        """Sets the weak form including linear and bilinear terms."""
+        for equation, solution, solution_old in zip(
+            self.equations, self.solution_split, self.solution_old_split
+        ):
+            if equation.mass_term:
+                assert equation.scaling_factor == -self.theta
+                self.F += equation.mass((solution - solution_old) / self.coupled_tstep)
+            self.F -= equation.residual(solution)
 
     def set_solver_options(self) -> None:
         """Sets PETSc solver parameters."""
@@ -369,99 +433,8 @@ class StokesSolver:
             elif INFO >= log_level:
                 self.solver_parameters["fieldsplit_1"]["ksp_converged_reason"] = None
 
-                if self.free_surface_map:
-                    # Gather pressure and free surface fields for Schur complement solve
-                    fields_ind = ",".join(map(str, range(1, len(self.solution_split))))
-                    self.solver_parameters.update(
-                        {
-                            "pc_fieldsplit_0_fields": "0",
-                            "pc_fieldsplit_1_fields": fields_ind,
-                        }
-                    )
-                    # Update mass inverse preconditioner
-                    self.solver_parameters["fieldsplit_1"].update(
-                        {"pc_python_type": "gadopt.FreeSurfaceMassInvPC"}
-                    )
-
-    def set_free_surface_boundary(
-        self, params_fs: dict[str, int | bool], bc_id: int
-    ) -> fd.ufl.algebra.Product | fd.ufl.algebra.Sum:
-        # Associate the free-surface index with the boundary id
-        # Note: This assumes that ordering of the free-surface boundary conditions is
-        # the same as that of free-surface functions in the mixed space.
-        self.free_surface_map[bc_id] = self.eta_ind
-
-        # Set internal degrees of freedom to zero to prevent singular matrix
-        self.strong_bcs.append(InteriorBC(self.solution_space[self.eta_ind], 0, bc_id))
-
-        normal_stress, self.buoyancy_fs[self.eta_ind] = (
-            self.approximation.free_surface_terms(
-                self.solution_split[1],
-                self.T,
-                self.solution_theta_split[self.eta_ind],
-                **params_fs,
-            )
-        )
-
-        return normal_stress
-
-    def set_equations(self) -> None:
-        u, p = self.solution_split[:2]
-        stress = self.approximation.stress(self.u)
-        source = self.approximation.buoyancy(self.p, self.T) * self.k
-        eqs_attrs = [
-            {"p": p, "stress": stress, "source": source},
-            {"u": u, "rho_mass": self.approximation.rho_continuity()},
-        ]
-
-        for i in range(len(residual_terms_stokes)):
-            self.equations.append(
-                Equation(
-                    self.tests[i],
-                    self.solution_space[i],
-                    residual_terms_stokes[i],
-                    eq_attrs=eqs_attrs[i],
-                    approximation=self.approximation,
-                    bcs=self.weak_bcs,
-                    quad_degree=self.quad_degree,
-                )
-            )
-
-        for bc_id, eta_ind in self.free_surface_map.items():
-            eq_attrs = {
-                "boundary_id": bc_id,
-                "buoyancy_scale": self.buoyancy_fs[eta_ind],
-                "u": u,
-            }
-
-            self.equations.append(
-                Equation(
-                    self.tests[eta_ind],
-                    self.solution_space[eta_ind],
-                    free_surface_term,
-                    mass_term=mass_term_fs,
-                    eq_attrs=eq_attrs,
-                    quad_degree=self.quad_degree,
-                    scaling_factor=-self.theta,
-                )
-            )
-
-    def set_form(self) -> None:
-        """Sets the weak form including linear and bilinear terms."""
-        for equation, solution, solution_old in zip(
-            self.equations, self.solution_split, self.solution_old_split
-        ):
-            if equation.mass_term:
-                assert equation.scaling_factor == -self.theta
-                self.F += equation.mass((solution - solution_old) / self.coupled_tstep)
-            self.F -= equation.residual(solution)
-
     def setup_solver(self) -> None:
         """Sets up the solver."""
-        if self.free_surface_map:
-            self.appctx["free_surface"] = self.free_surface_map
-            self.appctx["ds"] = self.equations[-1].ds
-
         if self.constant_jacobian:
             trial = fd.TrialFunction(self.solution_space)
             F = fd.replace(self.F, {self.solution: trial})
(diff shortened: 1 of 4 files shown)


gadopt/stokes_integrators.py
```diff
@@ -352,67 +430,153 @@ def set_solver_options(self) -> None:
...
+        if self.solver_parameters_update is not None:
+            for key, value in self.solver_parameters_update.items():
+                if isinstance(value, dict):
+                    self.solver_parameters[key] |= value
+                else:
+                    self.solver_parameters[key] = value
+
+    def set_solver(self) -> None:
+        """Sets up the Firedrake variational problem and solver."""
+        if self.constant_jacobian:
```
Patol75, line 460: Given that `constant_jacobian` can be `False` in `LinearVariationalProblem`, is it reasonable to use the value of `self.constant_jacobian` as a condition to determine between linear and non-linear problems?

Patol75: But we still wanted a linear solver in that case, right?

gadopt/stokes_integrators.py
```diff
@@ -352,67 +430,153 @@ def set_solver_options(self) -> None:
...
+            self.solver_parameters |= iterative_stokes_solver_parameters
 
-        # Extra options for iterative solvers
+        # Extra monitoring options for iterative solvers
         if self.solver_parameters.get("pc_type") == "fieldsplit":
             if DEBUG >= log_level:
                 self.solver_parameters["fieldsplit_0"]["ksp_converged_reason"] = None
                 self.solver_parameters["fieldsplit_1"]["ksp_monitor"] = None
             elif INFO >= log_level:
                 self.solver_parameters["fieldsplit_1"]["ksp_converged_reason"] = None
```
Patol75, line 449: I am wondering how useful these log levels are. Are we using them at all?

gadopt/stokes_integrators.py
w3168: yeah we still want a linear solver

gadopt/stokes_integrators.py
dsroberts: One of the things lower down on my to-do list is to expand the logging interface. I'm using it in the solver initialisation thing I'm working on.

Patol75: Cool, I will leave this block be then.

gadopt/free_surface_equation.py
```diff
@@ -23,7 +23,7 @@ def free_surface_term(
     eq: Equation, trial: fd.Argument | fd.ufl.indexed.Indexed | fd.Function
 ) -> fd.Form:
     r"""Free Surface term: u \dot n"""
-    F = -eq.buoyancy_scale * eq.test * fd.dot(eq.u, eq.n) * eq.ds(eq.boundary_id)
```
w3168, line 26: Sorry to be annoying but do you need to rename this in this PR?

Patol75: You are not! I will revert that change.

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
...
-
-        # eta is a list of 0, 1 or multiple free surface fields
-        self.u, self.p, *self.eta = fd.split(self.solution)
+        self.tests = fd.TestFunctions(self.solution_space)
 
         self.rho_mass = self.approximation.rho_continuity()
+        if hasattr(self.approximation, "mu"):
+            self.is_linear = not depends_on(self.approximation.mu, self.solution)
+        else:
+            self.is_linear = True
```
Patol75, line 342: Is this useful?

gadopt/stokes_integrators.py
```diff
@@ -352,67 +438,162 @@ def set_solver_options(self) -> None:
...
+                    self.solver_parameters[key] |= value
+                else:
+                    self.solver_parameters[key] = value
+
+    def set_solver(self) -> None:
+        """Sets up the Firedrake variational problem and solver."""
+        if self.forcing_term is not None:
+            self.F += self.forcing_term
+
+        if self.is_linear:
```
Patol75, line 471: Can someone confirm this should be reverted to `if self.constant_jacobian:` despite the fact that `LinearVariationalProblem` accepts `constant_jacobian=False`?

gadopt/stokes_integrators.py
```diff
@@ -352,67 +438,162 @@ def set_solver_options(self) -> None:
...
+            self.solution_split[1],
+            self.T,
+            self.solution_theta_split[self.eta_ind],
+            **params_fs,
+        )
+        # Associate the free-surface index with the boundary id and buoyancy term
         # Note: This assumes that ordering of the free-surface boundary conditions is
         # the same as that of free-surface functions in the mixed space.
-        self.free_surface_map[bc_id] = self.eta_ind
+        self.free_surface_map[bc_id] = [self.eta_ind, buoyancy]
```
Patol75, line 584: Not sure if a good idea to pack these two together.

Patol75 (commented):
Reminder to address solver parameters being deleted in PDE constrained optimisation demos.

gadopt/stokes_integrators.py
dsroberts: I what you've done here should stay, `approximation` is annotated as a `BaseApproximation` object, which doesn't guarantee the existence of a `mu` class member. Whilst all of its subclasses currently define a `mu` parameter, maybe one day there will be one that doesn't? Out of scope for this PR, but we should probably address other cases like this where class members that may not exist in abstract base classes are referenced. I don't think there are too many.

gadopt/stokes_integrators.py
```diff
@@ -603,64 +740,64 @@ def ala_right_nullspace(
...
-        )
+        self.stress_old = stress_old  # Deviatoric stress from previous time step
+        self.displacement = displacement  # Total displacement
 
-        scale_mu = fd.Constant(1e10)  # this is a scaling factor roughly size of mantle maxwell time to make sure that solve converges with strong bcs in parallel...
-        self.F = (1 / scale_mu)*self.F
+        # Replace approximation's viscosity with effective viscosity
+        approximation.mu = approximation.effective_viscosity(self.dt)
+        # Scaling factor for the previous stress
+        self.stress_scale = self.approximation.prefactor_prestress(self.dt)
```
dsroberts, line 800: Scientifically this is probably a silly question, but what happens if `approximation` is not a `SmallDisplacementViscoelasticApproximation`? Are there other viscoelastic approximations that will need to be accounted for one day? On a technical, pedantic point, `SmallDisplacementViscoelasticApproximation` doesn't actually subclass `BaseApproximation`, so any approximation that matches the type annotation won't have the `effective_viscosity` or `prefactor_prestress` functions.

gadopt/stokes_integrators.py
```diff
@@ -669,12 +806,9 @@ def set_free_surface_boundary(
         # `incremental displacement' (u) that we are solving for. Then, calculate the
         # free surface stress term. This is also referred to as the Hydrostatic
         # Prestress advection term in the GIA literature.
-        u, p = self.solution_split[:2]
+        u, p = self.solution_split
```
dsroberts, line 809: Is `self.solution_split` guaranteed to only have 2 subfunctions here? Will that always be true? This seems like an odd change as it removes a safeguard against potential future issues.

gadopt/stokes_integrators.py
```diff
@@ -325,23 +377,57 @@ def set_boundary_conditions(self) -> None:
...
+        Equations must be ordered like solutions in the mixed space.
+        """
+
+    def set_form(self) -> None:
+        """Sets the weak form including linear and bilinear terms."""
+        for equation, solution, solution_old in zip(
+            self.equations, self.solution_split, self.solution_old_split
+        ):
+            if equation.mass_term:
+                assert equation.scaling_factor == -self.theta
```
dsroberts, line 418: I don't think `assert` belongs in production code. If this is a fatal error, tell the user why. If it isn't, remove it.

gadopt/stokes_integrators.py
```diff
@@ -699,17 +833,25 @@ def set_equations(self) -> None:
                     approximation=self.approximation,
                     bcs=self.weak_bcs,
                     quad_degree=self.quad_degree,
+                    # Scaling factor roughly size of mantle Maxwell time to make sure
+                    # that solve converges with strong bcs in parallel
+                    scaling_factor=1e-10,
```
dsroberts, line 838: This is very different to 0.5, and I'm sure your comment explains it adequately, but could this cause issues? I guess its OK as `ViscoelasticStokesSolver` doesn't ever have a `mass_term` so the `assert`/fatal error above won't trip? Will that break if a user introduces a forcing with a mass term? Is that a dumb thing to do? Will a user try it anyway?

dsroberts (commented):
A couple of minor things and a few questions from me. Looks good though, and I'm definitely in favour of more consistency between the momentum and transport solvers. This is a step in the right direction in that regard.

gadopt/stokes_integrators.py
```diff
@@ -325,23 +377,57 @@ def set_boundary_conditions(self) -> None:
...
+                self.F += equation.mass((solution - solution_old) / self.dt)
+            self.F -= equation.residual(solution)
+
+    def set_petsc_options(self) -> None:
+        """Sets PETSc solver options."""
         # Application context for the inverse mass matrix preconditioner
-        self.appctx = {
-            "mu": self.approximation.mu / self.approximation.rho_continuity()
-        }
+        self.appctx = {"mu": self.approximation.mu / self.rho_mass}
```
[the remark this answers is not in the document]
Patol75: Good question. I think @w3168 will have an idea.

gadopt/stokes_integrators.py
Patol75: Yes, for this solver, the mixed space should only contain two spaces.

gadopt/stokes_integrators.py
Patol75: Fair, I will have to put a `raise` then.

gadopt/stokes_integrators.py
Patol75: So the idea was to ensure that the scaling factor was `theta` when the system includes a mass term. The reason for this is that we hard-code the timestepping scheme for the mass term because we are not coupling mass and momentum with temperature. In this case, there is no mass term, and @w3168 introduced a scaling factor to ensure some stability. I think the instability is similar to #247, and we currently do not have another fix for it.

gadopt/stokes_integrators.py
```diff
@@ -325,23 +377,57 @@ def set_boundary_conditions(self) -> None:
...
                     case _:
                         weak_bc[bc_type] += val
 
             self.weak_bcs[bc_id] = weak_bc
 
-    def set_solver_options(self) -> None:
-        """Sets PETSc solver parameters."""
+    def set_free_surface_boundary(
+        self, params_fs: dict[str, int | bool], bc_id: int
+    ) -> fd.ufl.algebra.Product | fd.ufl.algebra.Sum:
```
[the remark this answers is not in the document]
dsroberts: Well, here is what happens if a solver doesn't define `set_free_surface_boundary` but a user attempts to set a `free_surface` boundary condition:
```
  File "/home/dale/repos/g-adopts/review/demos/glacial_isostatic_adjustment/base_case/base_case.py", line 434, in <module>
    stokes_solver = ViscoelasticStokesSolver(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/dale/repos/g-adopts/review/gadopt/stokes_integrators.py", line 224, in __call__
    class_instance.prepare_solver()
  File "/home/dale/repos/g-adopts/review/gadopt/stokes_integrators.py", line 349, in prepare_solver
    self.set_boundary_conditions()
  File "/home/dale/repos/g-adopts/review/gadopt/stokes_integrators.py", line 377, in set_boundary_conditions
    weak_bc["normal_stress"] += self.set_free_surface_boundary(
TypeError: unsupported operand type(s) for +=: 'Product' and 'NoneType'
```
I get that its a silly thing to do scientifically, but that doesn't mean it won't happen by accident at some point. Maybe `SolverBase.set_free_surface_boundary` can raise an exception and tell the user "This solver does not implement a free surface boundary, but a free surface boundary has been requested".

gadopt/stokes_integrators.py
dsroberts: So change the type annotation for now? At least that way a user new to G-ADOPT will get a correct message from their linter. Right now we get:
```
Argument of type "SmallDisplacementViscoelasticApproximation" cannot be assigned to parameter "approximation" of type "BaseApproximation" in function "__init__"
```
Which is exactly the opposite of what should happen.

gadopt/stokes_integrators.py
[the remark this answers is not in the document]
sghelichkhani: Why do we need to provide that freedom of not having a `mu`? Where does it become useful?

Just looking at this I am thinking we require `mu` in Approximation. At least for the viscous systems that we are discussing here, we do need a viscosity (zero viscosity breaks all the assumptions)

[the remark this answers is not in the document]
dsroberts: @sghelichkhani because it is allowed by `BaseApproximation`. If it doesn't make sense to account for an undefined `mu`, then `mu` should not be allowed to be undefined. In that case I think the correct approach would be to set `mu` as a `@property` in `BaseApproximation`, thus forcing every subclass to have it defined, like `compressible` and `Tbar`. @w3168's new approximations don't subclass `BaseApproximation` so this should be OK, though maybe it means `BaseApproximation` needs renaming?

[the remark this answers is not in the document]
sghelichkhani: Since this is all different solvers for the Stokes system, it is very unreasonable to assume `mu` can be non-existent. Unless we want to have a solver for *everything*, which I would argue itself is unreasonable.

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
...
-             theta = 0: Forward Euler, theta = 0.5: Crank-Nicolson (default),
-             or theta = 1: Backward Euler
+    def __call__(cls, *args, **kwargs):
+        class_instance = super().__call__(*args, **kwargs)
+        class_instance.prepare_solver()
+
+        return class_instance
+
+
+class SolverBase(abc.ABC, metaclass=MetaPostInit):
```
sghelichkhani, line 229: It probably makes sense to call this specifically `StokesSolverBase`? I feel calling it `solver` is not informative.

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
...
+
+    ### Valid keys for boundary conditions
+    |   Condition   |  Type  |                 Description                  |
+    | :------------ | :----- | :------------------------------------------: |
+    | u             | Strong | Solution                                     |
+    | ux            | Strong | Solution along the first Cartesian axis      |
+    | uy            | Strong | Solution along the second Cartesian axis     |
+    | uz            | Strong | Solution along the third Cartesian axis      |
+    | un            | Weak   | Solution along the normal to the boundary    |
+    | stress        | Weak   | Traction across the boundary                 |
```
sghelichkhani, line 240: Does that actually work for any chid classs? Like can we impose normal stresses for a viscous flow?

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
...
+    | free_surface  | Weak   | Free-surface characteristics of the boundary |
+
+    ### Valid keys describing the free surface boundary:
+    |     Argument    | Required |                   Description                    |
+    | :-------------- | :------: | :----------------------------------------------: |
+    | delta_rho_fs    | Yes (d)  | Density contrast along the free surface          |
+    | RaFS            | Yes (nd) | Rayleigh number (free-surface density contrast)  |
+    | variable_rho_fs | No       | Account for buoyancy effects on interior density |
+
+    ### Classic theta values for coupled implicit time integration
```
sghelichkhani, line 251: It is important here to mention why we have time-integration (Stokes with no free surface does not have time). So say this is for cases with a free-surface case only.

gadopt/stokes_integrators.py
sghelichkhani: I think we have discussed this on the meeting? So I think this does not need to change?

w3168: Isn't this the example of the free surface with adaptive dt? i.e. the probablem is linear so you probably could just set up a linear variational problem but pass constant_jacobian=False.  Presumably this is currently happening for the free surface demo?

gadopt/stokes_integrators.py
```diff
@@ -352,67 +438,161 @@ def set_solver_options(self) -> None:
...
+      forcing_term:
+        Firedrake form specifying an additional term contributing to the residual
+      bcs:
+        Dictionary specifying boundary conditions (identifier, type, and value)
+      quad_degree:
+        Integer denoting the quadrature degree
+      solver_parameters:
+        Dictionary of PETSc solver options or string matching a default set thereof
+      solver_parameters_update:
+        Dictionary of PETSc solver options used to update the default G-ADOPT options
```
sghelichkhani, line 535: Was there a problem with the old behvaiour? Presumably this is your way or overwriting some of the parameters, and not all of them? It's very reasonable to assume that you need to specify the `parameters` once.

gadopt/stokes_integrators.py
Patol75: If you pass `constant_jacobian=False`, which is the default, you end up setting up a `NonLinearVariationalProblem`.

gadopt/stokes_integrators.py
Patol75: That is fair. Then, let's assume that `mu` is an attribute of `approximation` here, and if this logic ever breaks, then we will have to revisit.

gadopt/stokes_integrators.py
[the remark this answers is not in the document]
Patol75: And `eta`, which is another usual name for it, is now used to denote the free surface.
For the viscosity in general, I think it makes more sense to assume the approximation has an attribute called `mu`. If we impose its presence in the approximation, then someone writing an approximation that does not include a viscosity would be left quite puzzled as to why such a choice was made.

sghelichkhani (changes requested):
Some comments. More in depth will folloow

gadopt/stokes_integrators.py
```diff
@@ -58,42 +63,34 @@
...
-adjusted by accessing the solver_parameter dictionary, for example:
-
-```
-   stokes_solver.solver_parameters['fieldsplit_0']['ksp_converged_reason'] = None
-   stokes_solver.solver_parameters['fieldsplit_0']['ksp_rtol'] = 1e-3
-   stokes_solver.solver_parameters['fieldsplit_0']['assembled_pc_gamg_threshold'] = -1
-   stokes_solver.solver_parameters['fieldsplit_1']['ksp_converged_reason'] = None
-   stokes_solver.solver_parameters['fieldsplit_1']['ksp_rtol'] = 1e-2
-```
+"""Default iterative solver parameters for solution of Stokes system.
```
sghelichkhani, line 68: I appreciate you trying to improve the docstring. But then it becomes impossible to focus on all these little things you have changed. As you said, incremental steps, rather than changing everything.

https://user-images.githubusercontent.com/217029/181068444-9a21be12-9ed3-42cb-95d3-594e75d6192c.png

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
     return fd.MixedVectorSpaceBasis(Z, null_space)
 
 
-class StokesSolver:
-    """Solves the Stokes system in place.
+class MetaPostInit(abc.ABCMeta):
```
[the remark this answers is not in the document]
sghelichkhani: So the plan is to get this in, and then remove it? I sometimes think it might be better to go with #271 first. Just a thought!

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
...
+    | RaFS            | Yes (nd) | Rayleigh number (free-surface density contrast)  |
+    | variable_rho_fs | No       | Account for buoyancy effects on interior density |
+
+    ### Classic theta values for coupled implicit time integration
+    | Theta |        Scheme         |
+    | :---- | :-------------------: |
+    | 0.5   | Crank-Nicolson method |
+    | 1.0   | Backward Euler method |
+
+    Args:
```
sghelichkhani, line 257: I don't know if what I am saying is standard. But in a docstring I rather first see what each argument is, and further down see the details regarding each. Here we start by boundary conditions, without first saying what all the arguments are.

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
...
+      solution:
+        Firedrake function representing the field over the mixed Stokes space
+      approximation:
+        G-ADOPT approximation defining terms in the system of equations
+      dt:
```
sghelichkhani, line 262: Maybe mention for instantaneous problems this is redundant?

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
...
         self,
         solution: fd.Function,
-        T: fd.Function,
         approximation: BaseApproximation,
-        bcs: dict[int, dict[str, Number]] = {},
+        /,
+        *,
+        dt: float | None = None,
+        theta: float = 0.5,
+        forcing_term: fd.Form | None = None,
```
sghelichkhani, line 299: I am not really sure sure about this. I feel it's a regression in the clarity fo the code. Maybe for now we can call it `non_thermal_buoyancy`? or something more pointing to why we have an additional term rather than `T`? Like a user might look at this and think they need to provide the thermal buoyancy here? 

My ideal solution is to drop this here, and address it in a seperate PR where we address the Stokes RHS.

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
...
+        bcs: dict[int | str, dict[str, Any]] = {},
```
sghelichkhani, line 300: Thanks for fixing this.

gadopt/stokes_integrators.py
```diff
@@ -58,42 +63,34 @@
...
+The `fieldsplit_1` entries contain solver options for the Schur complement solve itself.
+For preconditioning, we approximate the Schur complement matrix with a mass matrix
+scaled by viscosity, with the viscosity sourced from the approximation object provided
+to Stokes solver. Since this preconditioner step involves an iterative solve, the Krylov
+method used for the Schur complement needs to be of flexible type, and we use the
+Flexible Generalized Minimal Residual (`fgmres`) method by default.
 
 Note:
-  G-ADOPT defaults to iterative solvers in 3-D.
+  G-ADOPT will use the above default iterative solver parameters if the argument
```
sghelichkhani, line 87: I think it makes sense to drop this update_parameters method as Dale will be changing this.

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
...
       transpose_nullspace:
         A `MixedVectorSpaceBasis` for the kernel of the operator's transpose
       near_nullspace:
         A `MixedVectorSpaceBasis` for the operator's smallest eigenmodes (e.g. rigid
         body modes)
-
     """
 
-    name = "Stokes"
+    name = "Solver"
```
sghelichkhani, line 289:
```suggestion
    name = "MomentumSolver"
```

gadopt/stokes_integrators.py
sghelichkhani: Thanks Thomas. Does it then make sense to introduce `viscosity` or whatever we agree on next week as an abstract property method for `BaseApproximation`?

like: 
```python
    @property
    @abc.abstractmethod
    def mu(self) -> Function | Number:
        """Dynamic viscosity parameter.
        pass
```

gadopt/stokes_integrators.py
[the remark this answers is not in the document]
sghelichkhani: Since these solvers are specifically for Stokes, we are going to have a free surface implementation anyway. So I think we are fine with the suggested approach.

gadopt/stokes_integrators.py
```diff
@@ -58,42 +63,34 @@
...
-viscosity provided through the optional mu keyword argument to Stokes
-solver. Since this preconditioner step involves an iterative solve,
-the Krylov method used for the Schur complement needs to be of
-flexible type, and we use FGMRES by default.
-
-We note that our default solver parameters can be augmented or
-adjusted by accessing the solver_parameter dictionary, for example:
-
-```
-   stokes_solver.solver_parameters['fieldsplit_0']['ksp_converged_reason'] = None
```
sghelichkhani, line 88: I actually think this old bit of showing an example was very useful.

gadopt/stokes_integrators.py
```diff
@@ -325,23 +377,57 @@ def set_boundary_conditions(self) -> None:
...
+        ):
+            if equation.mass_term:
+                assert equation.scaling_factor == -self.theta
+                self.F += equation.mass((solution - solution_old) / self.dt)
+            self.F -= equation.residual(solution)
+
+    def set_petsc_options(self) -> None:
+        """Sets PETSc solver options."""
         # Application context for the inverse mass matrix preconditioner
-        self.appctx = {
```
sghelichkhani, line 337: Is the change from `rho_continuity`->`rho_mass` necessary? Both could be valid names, but why the change?

gadopt/stokes_integrators.py
```diff
@@ -352,67 +438,161 @@ def set_solver_options(self) -> None:
...
-                    # Update mass inverse preconditioner
-                    self.solver_parameters["fieldsplit_1"].update(
-                        {"pc_python_type": "gadopt.FreeSurfaceMassInvPC"}
-                    )
+        if self.solver_parameters_update is not None:
+            for key, value in self.solver_parameters_update.items():
+                if isinstance(value, dict):
+                    self.solver_parameters[key] |= value
+                else:
+                    self.solver_parameters[key] = value
```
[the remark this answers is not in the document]
sghelichkhani: @Patol75 Would be nice if you try to revert the part related to `solver_parameters_update` so it makes it more clear what will go in as part of this PR.

gadopt/stokes_integrators.py
```diff
@@ -352,67 +438,161 @@ def set_solver_options(self) -> None:
...
     def set_free_surface_boundary(
         self, params_fs: dict[str, int | bool], bc_id: int
     ) -> Expr:
-        # Associate the free-surface index with the boundary id
+        # Set internal degrees of freedom to zero to prevent singular matrix
+        self.strong_bcs.append(
+            InteriorBC(self.solution_space[self.eta_ind], 0.0, bc_id)
+        )
+
+        normal_stress, buoyancy = self.approximation.free_surface_terms(
```
sghelichkhani, line 574: `buoyancy`? Are you talking about surface load? Or is actually right hand side buoyance to Stokes? Avoid using `buoyancy`

gadopt/stokes_integrators.py
```diff
@@ -603,64 +740,64 @@ def ala_right_nullspace(
     return p
 
 
-class ViscoelasticStokesSolver(StokesSolver):
+class ViscoelasticStokesSolver(SolverBase):
```
sghelichkhani, line 743: If we end up just having one StokesSolver, this would end up not being exposed,

sghelichkhani (changes requested):
Thanks @Patol75 . I have a left a few other comments. For now I think the main feedback is that it would be nice to remove the changes that will be anyway superseded by #271. For example those related to `solver_parameters_update` are IMO not a good solution. I think it would be OK to leave the viscoelastic solver as is for now as it will be again superceded by Will's implementation.

gadopt/stokes_integrators.py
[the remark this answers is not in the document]
w3168: I think this is specifically to do with the incompressible stokes mantle convection right? i.e. we need it in the inverse mass matrix for pressure? 

I just realised ```mu``` = viscosity as a keyword is actually hard baked into firedrake e.g. https://www.firedrakeproject.org/_modules/firedrake/preconditioners/massinv.html#MassInvPC 

Anyway I don't think we want this in the internal variable solver parameters at the moment because there is no pressure enforcing divergence freeness and hence no schur complement fieldsplit. 

Maybe with #271 this will be separated out and be clearer?

gadopt/stokes_integrators.py
w3168: Sorry I got a bit lost trying to work out what this has changed. Why did you put them together?

gadopt/stokes_integrators.py
w3168: The short answer is that the internal variable solver free surface is quite different to this.

At the moment the _free surface_ is quite different again in the internal variable branch. Because we are solving for displacement and not incremental displacement it makes more sense to me to keep it closer to the advection of hydrostatic prestress term that it derives from. Also with compressibility we need to explicitly write out a body integral term after integrating the hydrostatic prestress term by parts. For the incompressible case it just gets incorporated into the pressure.

gadopt/stokes_integrators.py
w3168: It's like Thomas mentioned this is the old code before we nondimensionalised it, without the scaling factor there were some funny edge effects with strong bcs. 

Since we nondimensionalised the internal variable equations I havent needed a scaling factor (except in the nonlinear case to actually make the starting residaul larger but lets forget about that for now!)

gadopt/stokes_integrators.py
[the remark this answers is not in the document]
dsroberts: Yeah, let's revert this. It'll make your API change clearer and reduce the size of the PR overall.

gadopt/stokes_integrators.py
```diff
@@ -1,13 +1,18 @@
-r"""This module provides a fine-tuned solver class for the Stokes system of conservation
-equations and a function to automatically set the associated null spaces. Users
-instantiate the `StokesSolver` class by providing relevant parameters and call the
-`solve` method to request a solver update.
-
+"""Solver classes targetting systems dealing with momentum conservation.
+
+This module provides a fine-tuned abstract base class, `SolverBase`, from which
+efficient numerical solvers can be derived, such as `StokesSolver` for the Stokes system
```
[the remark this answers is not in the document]
Patol75: For now, let's have a base class that solvers inherit from. I will keep Maxwell in as well, given that removal of this solver is out of scope here.

gadopt/stokes_integrators.py
Patol75: I will blame GitHub here. It should honestly do a better job at highlighting differences in docstrings. Essentially, I have ensured a line length of 89 characters and consistency between the different solver parameter descriptions. I figured it was important to make sure all objects connected to the solver abc were consistently documented.

gadopt/stokes_integrators.py
Patol75: Fair, I changed that in #113, but happy to revert.

gadopt/stokes_integrators.py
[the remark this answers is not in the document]
Patol75: I think the idea is that, once #271 is in, the content of `prepare_solver` can be dumped at the beginning of `set_solver`. So happy to proceed this way.

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
...
+
+
+class SolverBase(abc.ABC, metaclass=MetaPostInit):
+    """Solver for a system involving mass and momentum conservation.
+
+    ### Valid keys for boundary conditions
+    |   Condition   |  Type  |                 Description                  |
+    | :------------ | :----- | :------------------------------------------: |
+    | u             | Strong | Solution                                     |
+    | ux            | Strong | Solution along the first Cartesian axis      |
```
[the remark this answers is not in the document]
Patol75: I see your point, but then in 2-D some people like to think of their axes as xz instead of xy, and there could perhaps be a confusion. Is it this bad to have it using first, second, and third?

gadopt/stokes_integrators.py
Patol75: So currently, all these boundary conditions are implemented in `momentum_equation`. Given all solvers here use the same terms, then these weak boundary conditions should work, yes.

gadopt/stokes_integrators.py
Patol75: I think that makes sense, thank you.

gadopt/stokes_integrators.py
Patol75: I think it is reasonable addition here, which does not preclude an eventual future change. I have renamed the argument to `additional_forcing_term`, which makes its intent clearer, hopefully.

gadopt/stokes_integrators.py
```diff
@@ -199,112 +210,153 @@ def create_stokes_nullspace(
...
+        *,
+        dt: float | None = None,
+        theta: float = 0.5,
+        forcing_term: fd.Form | None = None,
+        bcs: dict[int | str, dict[str, Any]] = {},
         quad_degree: int = 6,
-        solver_parameters: Optional[dict[str, str | Number] | str] = None,
-        J: Optional[fd.Function] = None,
+        solver_parameters: dict[str, str | float] | str | None = None,
+        solver_parameters_update: dict[str, str | float] | str | None = None,
```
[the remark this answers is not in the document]
Patol75: I would not call it unorthodox because that is essentially what #271 will enable, but I see the point to drop it here.

gadopt/stokes_integrators.py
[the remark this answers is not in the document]
Patol75: Reverting

gadopt/stokes_integrators.py
Patol75: The problem with the current behaviour is the fact that the solver is not fully ready after class instantiation. `solve` should not have to conclude initialisation before performing the solve.

gadopt/stokes_integrators.py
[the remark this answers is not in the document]
sghelichkhani: hmm ... But in all meshes 2d problems are xy, right?
Of course nothing that bad to keep it this way, just reads weird!

gadopt/stokes_integrators.py
Patol75: I will go ahead with the simple approach, and we can revisit in #268.

gadopt/stokes_integrators.py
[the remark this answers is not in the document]
Patol75: I went ahead with raising an exception.

gadopt/stokes_integrators.py
Patol75: That is the buoyancy term that arises in the free-surface equation. It is a `\Delta\rho * g`, similar to the momentun source term.

gadopt/stokes_integrators.py
Patol75: Because I wanted to avoid having to track too many attributes. Maybe it is OK like this.

gadopt/stokes_integrators.py
Patol75: For now, it will stay, but let's see where we end up after #268 and #271.

gadopt/stokes_integrators.py
Patol75: Let's wait until #268 to see how things lay out. For now, just changing the annotation.

gadopt/stokes_integrators.py
Patol75: Leaving as-is given this formulation only incorporates two variables in the mixed space.

gadopt/stokes_integrators.py
Patol75: Leaving as-is until further development (probably in #268?)

w3168 (commented):
Thanks very much @Patol75 - well done this is a lot of work! Personally I'm very happy with this as is. It's definitely a step in the right direction so I would suggest we merge this and then work on #271 and then #268.

gadopt/stokes_integrators.py
```diff
@@ -322,21 +391,62 @@ def set_boundary_conditions(self) -> None:
...
-                        self.eta_ind += 1
                     case _:
                         weak_bc[bc_type] += val
 
             self.weak_bcs[bc_id] = weak_bc
 
-    def set_solver_options(self) -> None:
-        """Sets PETSc solver parameters."""
+    def set_free_surface_boundary(
+        self, params_fs: dict[str, int | bool], bc_id: int
```
sghelichkhani, line 407: `bc_id` can be a string, right?

gadopt/stokes_integrators.py
```diff
@@ -199,26 +230,48 @@ def create_stokes_nullspace(
...
+      theta:
+        Float defining the theta scheme parameter used in a coupled time integration
+      additional_forcing_term:
+        Firedrake form specifying an additional term contributing to the residual
+      bcs:
+        Dictionary specifying boundary conditions (identifier, type, and value)
+      quad_degree:
+        Integer denoting the quadrature degree
+      solver_parameters:
+        Dictionary of PETSc solver options or string matching a default set thereof
```
sghelichkhani, line 268:
```suggestion
        Dictionary of PETSc solver options or string matching one of the default sets
```

sghelichkhani (approved):
Thanks Thomas! Letting go of the `solver_parameters_update` issue for now as you mentioned, and a couple of small issues I am approving this to be merged. 

For the future, I think it is really important to maintain the scope of a PR to make the review process much easier. I actually agree with the majority of your stylistic changes here, but it just made the review process very inefficient as I needed pay attention to many little things. Big fan of multiple smaller PRs rather than a big one.  

Well done!

PR #233, opened 2025-06-05
Remove unnecessary halo accesses in diagnostics

Patol75:
As discussed in gh-187, halo accesses are not required for diagnostic calculations. This PR remove such accesses in the `GeodynamicalDiagnostics` class.

### DIFF
--- a/gadopt/diagnostics.py
+++ b/gadopt/diagnostics.py
@@ -83,15 +83,15 @@ class GeodynamicalDiagnostics:
         return assemble(self.T * self.dx) / self.domain_volume
 
     def T_min(self):
-        T_data = self.T.dat.data_ro_with_halos[:]
+        T_data = self.T.dat.data_ro
         return self.T.comm.allreduce(T_data.min(), MPI.MIN)
 
     def T_max(self):
-        T_data = self.T.dat.data_ro_with_halos[:]
+        T_data = self.T.dat.data_ro
         return self.T.comm.allreduce(T_data.max(), MPI.MAX)
 
     def ux_max(self, boundary_id=None) -> float:
-        ux_data = self.u.dat.data_ro_with_halos[:, 0]
+        ux_data = self.u.dat.data_ro[:, 0]
 
         if boundary_id:
             bcu = DirichletBC(self.u.function_space(), 0, boundary_id)


gadopt/diagnostics.py
```diff
@@ -83,15 +83,15 @@ def T_avg(self):
         return assemble(self.T * self.dx) / self.domain_volume
 
     def T_min(self):
-        T_data = self.T.dat.data_ro_with_halos[:]
+        T_data = self.T.dat.data_ro
```
sghelichkhani, line 86: I might be missing the background to this change, Why are you not copying the array and referencing it here? Why not `self.T.dat.data_ro[:]`?

sghelichkhani (commented):
@Patol75 Could you maybe provide a background to this change? Are we just being pedantic, or is there an actual advantage to doing it this way?

Patol75:
Sure, the background is this [comment](https://github.com/g-adopt/g-adopt/pull/187#discussion_r2127774146) earlier today. From that point of view, there should not be a need to involve halos in diagnostic calculations, and so I thought it was reasonable to update the code to ensure best practice.

angus-g:
I think it could be useful to seek advice from the Firedrake (pyop2) folk. I agree that it's more performant to not force a halo exchange when every rank is looking at its own data (and performing some global collective anyway). Certainly by naming alone, you'd expect that we should be using the non-halo versions.

However, from a brief glance (at e.g. the plotting code), there are times when the halos are included when maybe you'd expect them to be excluded. So maybe there's some deeper reason, particularly when it comes to element representation...

gadopt/diagnostics.py
```diff
@@ -83,15 +83,15 @@ def T_avg(self):
...
         return self.T.comm.allreduce(T_data.min(), MPI.MIN)
 
     def T_max(self):
-        T_data = self.T.dat.data_ro_with_halos[:]
+        T_data = self.T.dat.data_ro
         return self.T.comm.allreduce(T_data.max(), MPI.MAX)
 
     def ux_max(self, boundary_id=None) -> float:
-        ux_data = self.u.dat.data_ro_with_halos[:, 0]
+        ux_data = self.u.dat.data_ro[:, 0]
```
angus-g, line 94: This one doesn't work:

```
  File "/__w/g-adopt/g-adopt/demos/mantle_convection/2d_compressible_TALA/2d_compressible_TALA.py", line 246, in <module>
    f"{gd.u_rms()} {gd.u_rms_top()} {gd.ux_max(boundary.top)} {gd.Nu_bottom()} "
                                     ^^^^^^^^^^^^^^^^^^^^^^^
  File "/__w/g-adopt/g-adopt/.venv/lib/python3.12/site-packages/gadopt/diagnostics.py", line 98, in ux_max
    ux_data = ux_data[bcu.nodes]
              ~~~~~~~^^^^^^^^^^^
IndexError: index 1622 is out of bounds for axis 0 with size 1620
```

dsroberts:
I agree with @angus-g. Like I said in the comment, I'm approaching this from MPI first principles. If there are additional firedrake shenanigans going on, it would be good to know. I didn't realise that `data_ro_with_halos` forces a halo exchange, so everything I said about halos being out of date can be ignored. Looking at the pyop2 routines, they do very similar things. The halos are tacked onto the end of the arrays, so `data_ro` just gives you `_data[:self.dataset.size]`, and `data_ro_with_halos` returns the whole array. If it is as straightforward as that, then yes, we should absolutely switch to `data_ro`.

@sghelichkhani The major advantage is that it prevents a halo exchange, which is a costly MPI operation. Unlike an `allreduce`, there is no opportunity for the MPI implementation to optimise point-to-point comms, so for a quadrilateral 3D mesh, you're looking at 6 distinct send/recv pairs per rank, or 12,480 extraneous MPI messages per timestep in a 2,080 core job. Another (and much less consequential) reason is that you're reducing the size of the array you have to do the reductions over, so a few microseconds to save there I guess.

gadopt/diagnostics.py
sghelichkhani: Yes, `DirichletBC.nodes` includes the halo nodes, which I think makes sense to avoid inconsistency between halos when assigning boundary conditions by changing arrays. In pygplates connector we have the same.

dsroberts:
@sghelichkhani hard to say without profiling. I believe @drhodrid's 4160 core 3D spherical gplates jobs have the `T_min` and `T_max` diagnostics enabled, so it'd be interesting to do a comparison between master and this branch (next quarter when we have more SU). Even if the benefit is negligible now, it is still worth getting rid of extraneous MPI operations now before they can become a problem later on.

Patol75:
Just confirming that bringing back halos access for boundary data allows CI to pass.

PR #236, opened 2025-06-11
Update memprof plot action to use updated memprof_plotter

dsroberts:
I've updated memprof_plotter to use `PyGithub`. Its a bit smarter now, it picks the `n` most recent successful runs that have artefacts. Since it doesn't use `gh` at all any more, the action here needs to be updated. There is also a test action over at g-adopt/memprof_plotter, we should all be able to see the artefacts from that action.


angus-g (approved):

angus-g:
Cool, just retriggered CI to see if we can fluke a pass on the flaky tests to get this in...

PR #242, opened 2025-06-21
Clarify and generalise `assign_level_set_values`

Patol75:
Through interacting with Fanny, who has started to use `G-ADOPT` to implement multi-material, thermomechanical subduction models, I realised that some items of the recently updated level-set API were not satisfactory. This PR deals with `assign_level_set_values`. The first item of interest is the function's documentation, which was either incomplete or unclear. The second item is the way level-set values (essentially, 0 or 1) are assigned, particularly for interfaces represented as curves. The current behaviour only allows for mathematical functions, but implicit curve equations should be handled. Finally, the whole idea behind  `assign_level_set_values` was to hide the `Shapely` API from the user. However, there are cases for which the simple scenarios currently implemented are not sufficient to generate the desired interface. In these cases, the user has to interact with `Shapely` directly and should be able to provide their `Shapely` object to `G-ADOPT`'s API. Such a scenario is not currently handled, and this PR addresses this gap.

For reference, this is the code Fanny and I wrote for the subduction case:
```python
# Material interface
plate_extremity_coords = (0.0, domain_dims[1])
trench_coords = (domain_dims[0] / 2, domain_dims[1])
weak_layer_thickness = 2e4
ann_outer_radius = 2.5e5
ann_centre = (trench_coords[0], trench_coords[1] - ann_outer_radius)
slab_tip_angle = np.deg2rad(77.0)

surface_plate_coords = [
    plate_extremity_coords,
    trench_coords,
    (trench_coords[0], trench_coords[1] - weak_layer_thickness),
    (plate_extremity_coords[0], plate_extremity_coords[1] - weak_layer_thickness),
    plate_extremity_coords,
]
surface_plate = sl.Polygon(surface_plate_coords)

ann_outer_circle = sl.Point(ann_centre).buffer(ann_outer_radius)
ann_inner_circle = sl.Point(ann_centre).buffer(ann_outer_radius - weak_layer_thickness)
annulus = ann_outer_circle.difference(ann_inner_circle)

slab_tip_triangle_coords = [
    trench_coords,
    (trench_coords[0] + ann_outer_radius * np.tan(slab_tip_angle), trench_coords[1]),
    ann_centre,
    trench_coords,
]
slab_tip_triangle = sl.Polygon(slab_tip_triangle_coords)

weak_layer_polygon = surface_plate.union(annulus.intersection(slab_tip_triangle))
weak_layer_polygon_coords = weak_layer_polygon.exterior.coords._coords
([plate_extremity_index],) = np.all(
    weak_layer_polygon_coords == plate_extremity_coords, axis=1
).nonzero()
weak_layer_interface_coords = weak_layer_polygon_coords[:plate_extremity_index]
weak_layer_interface = sl.LineString(weak_layer_interface_coords)

# Initialise level set
epsilon = interface_thickness(K)
assign_level_set_values(
    psi,
    epsilon,
    interface_geometry="shapely",
    interface=weak_layer_interface,
    boundary_coordinates=weak_layer_polygon_coords[plate_extremity_index:],
)
```

### DIFF
--- a/gadopt/level_set_tools.py
+++ b/gadopt/level_set_tools.py
@@ -77,35 +77,48 @@ def assign_level_set_values(
     epsilon: float | fd.Function,
     /,
     interface_geometry: str,
-    interface_coordinates: list[list[float]] | list[list[float], float] | None = None,
     *,
+    interface: sl.LineString | sl.Polygon | None = None,
+    interface_coordinates: list[list[float]] | list[list[float], float] | None = None,
     interface_callable: Callable | str | None = None,
     interface_args: tuple[Any] | None = None,
-    boundary_coordinates: list[list[float]] | np.ndarray | None = None,
+    boundary_coordinates: list[list[float]] | None = None,
 ):
     """Updates level-set field given interface thickness and signed-distance function.
 
     Generates signed-distance function values at level-set nodes and overwrites
     level-set data according to the conservative level-set method using the provided
-    interface thickness. By convention, the 1-side of the conservative level set is set
-    above the curve or inside the polygon or circle.
+    interface thickness. By convention, the 1-side of the conservative level set lies
+    inside the geometrical object outlined by the interface alone or extended with
+    domain boundary segments (to form a closed loop).
 
-    Three scenarios are currently implemented to generate the signed-distance function:
-    - The material interface is described by a mathematical function y = f(x). In this
-      case, `interface_geometry` should be `curve` and `interface_callable` must be
-      provided along with any `interface_args` to implement the aforementioned
-      mathematical function.
-    - The material interface is a polygon, and `interface_geometry` takes the value
-      `polygon`. In this case, `interface_coordinates` must exclude the polygon sides
-      that do not act as a material interface and coincide with domain boundaries. The
-      coordinates of these sides should be provided using the `boundary_coordinates`
-      argument such that the concatenation of the two coordinate objects describes a
-      closed polygonal chain.
+    This function uses `Shapely` to generate a geometrical representation of the
+    interface. It handles simple base scenarios for which the interface can be described
+    by a plane curve, a polygonal chain (open or closed), or a circle. In these cases, a
+    mathematical description of the latter geometrical objects is sufficient to generate
+    the interface. For more complex objects, users should set up their own geometrical
+    representation via `Shapely` and provide it to this function.
+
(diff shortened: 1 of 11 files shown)


gadopt/level_set_tools.py
```diff
@@ -77,35 +77,48 @@ def assign_level_set_values(
     epsilon: float | fd.Function,
     /,
     interface_geometry: str,
-    interface_coordinates: list[list[float]] | list[list[float], float] | None = None,
     *,
+    interface: sl.LineString | sl.Polygon | None = None,
+    interface_coordinates: list[list[float]] | list[list[float], float] | None = None,
```
dsroberts, line 82:
```suggestion
    interface_coordinates: list[list[float]] | tuple[float, float] | None = None,
```
If I've read your docs right, this should be the type annotation for the (centre, radius) case for `interface_coordinates`.

gadopt/level_set_tools.py
```diff
@@ -77,35 +77,48 @@ def assign_level_set_values(
...
     interface_callable: Callable | str | None = None,
     interface_args: tuple[Any] | None = None,
```
dsroberts, line 84:
```suggestion
    interface_args: tuple[Any, ...] | None = None,
```
While we're on pedantic type checking, `interface_args` has unknown length, so it should be annotated with `...`.

dsroberts (commented):
Given the expanded scope of the `assign_level_set_values` function, there needs to be a whole lot of error checking to make sure incompatible arguments are not passed to it. There are a few more cases other than the one highlighted below where this needs to happen. Just the ones I've spotted

- `interface_coordinates` must be provided when `interface_geometry == "circle"`
- if `interface_callable` is provided when `interface_geometry == "circle"`, the provided coordinates get trashed.
- if `interface_geometry == "shapely"` and `interface` is not a `Polygon`, `boundary_coordinates` must be provided.

This is all in your docs, but we should 'encourage' users to look there rather than draw their own conclusions after some numpy or firedrake exception pops up. I don't know about you, but I'll have a much easier time tracking down a mistake I've made if I see this exception:
```
ValueError: "curve" interface geometry has been selected but "boundary_coordinates" were not supplied
```
compared to this one:
```
ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 1 dimensions. The detected shape was (3,) + inhomogeneous part.
```

Patol75:
image

PR #243, opened 2025-06-23
Fix compatibility issues between level-set API and extruded meshes

Patol75:
The current level-set API is incompatible with extruded meshes. There are two issues, as far as I can tell. The first one lies with `interface_thickness`. `epsilon.interpolate(scale * fd.MinCellEdgeLength(level_set_space.mesh()))` will fail on an extruded mesh. I propose a fix using `cell_sizes`. The second issue lies with `set_gradient_solver`. That solver does not involve the G-ADOPT API (i.e. `Equation`), and so it will fail on an extruded mesh because of the surface integral term. I propose a fix following what is done in `Equation`. I also note that this latter issue disappears if gh-157 gets merged.


gadopt/level_set_tools.py
```diff
@@ -67,7 +67,13 @@ def interface_thickness(
       A Firedrake function holding the interface thickness values
     """
     epsilon = fd.Function(level_set_space, name="Interface thickness")
-    epsilon.interpolate(scale * fd.MinCellEdgeLength(level_set_space.mesh()))
+
+    if level_set_space.extruded:
+        scale = scale or 0.25
```
drhodrid, line 72: Why are these values different (0.25 for extruded versus 0.35 for other)? Sorry if it's obvious... but it wasn't to me from quick thinking.

Patol75: Because `0.25 * sqrt(2)` is roughly 0.35. See also my comment in the main thread.

dsroberts:
Oh, OK. Well, the code looks fine and I trust you know what you're doing but @drhodrid may need more convincing. I'll leave final approval to him.

gadopt/level_set_tools.py
angus-g: I think magic numbers like this deserve a comment so the history isn't lost to a GitHub thread!

Patol75:
I ran the compositional buoyancy demo (van Keken et al., 1997) in parallel (2 cores) on a 32x32-elements grid twice, using the following meshes:
- `RectangleMesh(*mesh_elements, *domain_dims, quadrilateral=True)`
- `ExtrudedMesh(IntervalMesh(mesh_elements[0], domain_dims[0]), layers=mesh_elements[1], layer_height=domain_dims[1] / mesh_elements[1]),`

I get the following results for `rms velocity` and `entrainment` at the end of the simulation (non-dimensional time of `2000.0`):
- `0.00020218872615183922` `0.8593749999999982`
- `0.00019981326699852354` `0.8544921874999982`

Patol75:
<details> 
  <summary>Screenshots</summary>

image
image
compare_end
</details>

Yes, I believe nodal distributions are identical. The mismatch is to be expected because interface thickness values are not calculated exactly in the same way. I looked into the extruded mesh object, but it was not obvious to me how to extract the minimum cell edge length information. We could always use cell sizes in both scenarios, but I think the minimum cell edge length is more appropriate. The reason why I am saying this is the scenario where the normal to a material interface is aligned with the edge that has the minimum length.

drhodrid:
Hmmmm interesting. I wonder if part of the differences are due to the fact that those models are under-resolved and the level-set goes a little funky. If you run the exact cases twice, how big are the differences to the numbers you provide? Do the differences reduce at higher resolution?

Patol75:
After chatting to @drhodrid, there might be a way using the logic from [extend_function_to_3d](https://github.com/g-adopt/g-adopt/blob/2ef1be7aa2448284eae4ed1f2e3d5d1bb231da0f/gadopt/utility.py#L207).

Consider the following code and visual output:
```python
import gmsh

from gadopt import *
from gadopt.utility import extend_function_to_3d

gmsh.initialize()
gmsh.model.add("mesh")

points = []
for x in range(4):
    points.append(gmsh.model.geo.addPoint(x**2, 0.0, 0.0, (x + 1) / 5))

lines = []
for i in range(3):
    lines.append(gmsh.model.geo.addLine(points[i], points[i + 1]))

gmsh.model.geo.synchronize()

gmsh.model.mesh.generate(1)

gmsh.write("mesh.msh")
gmsh.finalize()

mesh_1d = Mesh("mesh.msh")
func_space_1d = FunctionSpace(mesh_1d, "CG", 1)
func_1d = Function(func_space_1d, name="Function (1D)")

func_1d.interpolate(MinCellEdgeLength(mesh_1d))

layer_height = 0.629
mesh_2d = ExtrudedMesh(mesh_1d, 3, layer_height=layer_height)
func_space_2d = FunctionSpace(mesh_2d, "CG", 1)
func_2d = Function(func_space_2d, name="Function (2D)")

func_extended = extend_function_to_3d(func_1d, mesh_2d)
func_extended.interpolate(
    conditional(func_extended < layer_height, func_extended, layer_height)
)

output = VTKFile("test.pvd")
output.write(func_2d, func_extended)
```
image

This successfully recovers a 2-D pseudo-representation of `MinCellEdgeLength` on an extruded mesh. My main concern is about accessing the value of `layer_height` when only having the extruded mesh object. Any idea?

Patol75:
@angus-g @sghelichkhani You mentioned having a solution for the above issue of recovering `layer_height`. Would you mind sharing it so I can finalise this PR? Thank you.

angus-g:
Hmm, I was wrong about `layer_height` being stored directly. You'd probably have to recover that from the coordinates, which starts to get tricky... Arguably this is probably something that's useful Firedrake/UFL functionality though.

sghelichkhani:
> Hmm, I was wrong about `layer_height` being stored directly. You'd probably have to recover that from the coordinates, which starts to get tricky... Arguably this is probably something that's useful Firedrake/UFL functionality though.

Considering that we store `layers` on the Extruded mesh, does it make sense to store layer_heights as well? Maybe this could be an improvement on FD side?

sghelichkhani:
> image

Then using the coordinates it is! We still have `layers`, which would give you the layer height.

Patol75:
Having made progress on this, it is not getting very much anywhere, sadly. Whilst the above strategy is sometimes viable, problems such as 
```python
>>> from gadopt import *
>>> base_mesh = CircleManifoldMesh(10, degree=2)
>>> func_space = FunctionSpace(base_mesh, "CG", 1)
>>> func = Function(func_space)
>>> func.interpolate(MinCellEdgeLength(base_mesh))
/home/thomas/Software/venv-firedrake/lib/python3.12/site-packages/ufl/algorithms/apply_geometry_lowering.py:343: UserWarning: Only know how to compute cell edge lengths of P1 or Q1 cell.
  warnings.warn("Only know how to compute cell edge lengths of P1 or Q1 cell.")
[...]
NotImplementedError: Cannot handle geometric quantity type: <class 'ufl.geometry.MinCellEdgeLength'>
```
prevent any general applicability.

It seems to me that there are only 2 ways forward:
- accept and document differences between extruded and non-extruded scenarios
- use `cell_sizes` in both scenarios, even though it is far from ideal when cells' aspect ratio becomes large

Patol75:
Given that this issue is dragging on and no satisfactory solution has been proposed, I have decided to reinstate `cell_sizes` as the default strategy and add a boolean to permit the use of `MinCellEdgeLength`. If everyone is happy with that choice, then we should be able to merge, provided a CI green tick.

gadopt/level_set_tools.py
```diff
@@ -53,36 +53,46 @@
...
+    default strategy implemented here is to approximate the local minimum edge length
+    as the quotient of the `cell_sizes` mesh attribute and the square root of the mesh's
+    geometric dimension. If `MinCellEdgeLength` is desired instead, a value of `True`
+    must be provided for the `min_cell_edge_length` argument. The local minimum edge
+    length is then multiplied by a scaling factor, with a default value of 0.35 that can
+    be modified via the `scale` argument.
+
+    **Note**: The two different formulations implemented here are not equivalent and
+    will result in measurable changes, for example, between comparable extruded and
+    non-extruded meshes.
```
drhodrid, line 80: This last comment is fine, but should you also add a brief comment stating that cell_sizes yields the same behavior on both extruded and non-extruded meshes, for clarity?

sghelichkhani: Also if `MinCellEdgeLength` does not work on `ExtrudedMesh`, one should raise an error if the mesh is extruded.

sghelichkhani (changes requested):

gadopt/level_set_tools.py
```diff
@@ -490,21 +519,25 @@ def set_gradient_solver(self) -> None:
...
         trial = fd.TrialFunction(gradient_space)
 
         bilinear_form = fd.inner(test, trial) * fd.dx(domain=self.mesh)
         ibp_element = -self.solution * fd.div(test) * fd.dx(domain=self.mesh)
-        ibp_boundary = (
-            self.solution
-            * fd.dot(test, fd.FacetNormal(self.mesh))
-            * fd.ds(domain=self.mesh)
-        )
+        ibp_boundary = self.solution * fd.dot(test, fd.FacetNormal(self.mesh)) * ds
```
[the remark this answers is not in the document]
Patol75: It made its way a few lines above. But I do not think any of these `domain` arguments are needed, as they can all be inferred from the integrands. I removed them and am checking if CI thinks it is fine.

Patol75:
Thank you for the reviews, which I have addressed. Regarding testing `cell_sizes`, this is currently not the case. My idea was to include a `2d_cylindrical` demo for multi-material in a separate PR, and so `cell_sizes` would be tested there. Does that work?

drhodrid:
Thanks @Patol75 - this is now fine IMO. Please add a test of cell sizes when you update the testing suite.

drhodrid (approved):

sghelichkhani (approved):
Looking good!

PR #244, opened 2025-06-23
Trigger github_status rebuild when new memory profiling plots are created

dsroberts:
More or less the same as "Trigger documentation build" step in render.yml.

### DIFF
--- a/.github/workflows/memprof_plot.yml
+++ b/.github/workflows/memprof_plot.yml
@@ -26,3 +26,12 @@ jobs:
         with:
           name: memprof_plots
           path: memprof_plots
+      - name: Trigger gadopt_status page build
+        env:
+          GH_TOKEN: ${{ secrets.GADOPT_STATUS_DISPATCH_TOKEN }}
+        run: |
+          gh api --method POST \
+          -H "Accept: application/vnd.github+json" \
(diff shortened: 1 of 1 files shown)


.github/workflows/memprof_plot.yml
```diff
@@ -25,4 +25,13 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: memprof_plots
-          path: memprof_plots
\ No newline at end of file
+          path: memprof_plots
+      - name: Trigger gadopt_status page build
+        env:
+          GH_TOKEN: ${{ secrets.GADOPT_STATUS_DISPATCH_TOKEN }}
```
angus-g, line 31: Should we maybe just use the same token for both dispatches? A little bit less fuss for updating... It's annoying that they have to be tied to user accounts, rather than organisation-wide. But I guess as long as somebody generates one with the right permissions it can go in the secret.

dsroberts: Makes sense. The existing `DISPATCH_TOKEN` expires in less than a month. Could make a new one now with permissions on both repos. Do you want to handle that?

angus-g: Feel free to go ahead and replace it so you can update the PR with the right secret path.

PR #245, opened 2025-06-23
Fix Makefile `all` target preventing tests to run

Patol75:
For some reason, CI gave a green tick, but now nothing runs anymore. Proposed changes should fix the situation


dsroberts (approved):
LGTM

PR #246, opened 2025-06-25
Switch to DG 2 discretisation for temperature in adjoint demo. 

drhodrid:
Whilst the demo uses Direct solves, it is important that it also works with iterative solvers. This switch improves robustness of iterative solvers in the optimisation procedure: some of the guesses provided by ROL are a little oscillatory which proves problematic for the CG schemes.

No need for a review yet -- I just want to first ensure this passes CI, prior to tidying things up and pondering a few other aspects.


sghelichkhani (approved):
LGTM

PR #251, opened 2025-06-26
Pin VTK for pyvista compatibility

angus-g:
We can unpin when https://github.com/pyvista/pyvista/issues/7587 is closed (which should be pretty soon really). Should allow render jobs to complete again.


angus-g:
Render run is working again: https://github.com/g-adopt/g-adopt/actions/runs/15891464196

dsroberts (approved):
LGTM

PR #256, opened 2025-06-27
Fix `get_boundary_ids` and `test_boundary_detection`

### LINKED ISSUE #253: `get_boundary_ids` returns incorrect tags for `BoxMesh`
Say you try to run the 3-D Cartesian demo using a `BoxMesh` (`mesh = BoxMesh(nx, ny, nz, a, b, c, hexahedral=True)`) instead of an [extruded `RectangleMesh`](https://github.com/g-adopt/g-adopt/blob/e22fffc21d0608cf08e6911473720c66bdf0bb63/demos/mantle_convection/3d_cartesian/3d_cartesian.py#L43-L44), then you would encounter `DIVERGED_LINEAR_SOLVE` (my worst nightmare these last few days). Fortunately, thanks to @w3168 sharp thinking, this one should not be too hard to handle. The culprit is the incorrect assignment of boundary tags via `get_boundary_ids`. If one sets these tags manually as described [here](https://www.firedrakeproject.org/firedrake.html#firedrake.utility_meshes.BoxMesh), then everything runs as usual.

dsroberts:
Correct axis convention to z is depth for non-extruded 3D meshes in `get_boundary_ids`. Correct this in `test_boundary_detection` as well. Update paths to `mesh.geo` files in the `multi_material` tests.

Closes #253.

### DIFF
--- a/gadopt/utility.py
+++ b/gadopt/utility.py
@@ -512,13 +512,13 @@ def get_boundary_ids(mesh) -> SimpleNamespace:
     # in its own mesh creation functions.
 
     if mesh.topology_dm.hasLabel("Face Sets"):
-        axis_extremes_order = [["left", "right"], ["bottom", "top"], ["front", "back"]]
+        axis_extremes_order = [["left", "right"], ["bottom", "top"]]
         dim = mesh.geometric_dimension()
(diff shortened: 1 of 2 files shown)


Patol75 (approved):
Provided CI returns a green tick, this is good to go! Thank you, @dsroberts!

PR #257, opened 2025-06-27
Fix warnings produced through unit tests

Patol75:
Current unit-test execution produces warnings, namely:
- `UserWarning: Discontinuous Lagrange element requested on quadrilateral, creating DQ element.`
- `UserWarning: Discontinuous Lagrange element requested on interval * interval, creating DQ element.`
- `FutureWarning: The use of 'interpolate' to perform the numerical interpolation is deprecated.`
- `UserWarning: PyOP2.Global has no comm, this is likely to break in parallel!`

This PR proposes changes to prevent warning occurrences.

### DIFF
--- a/gadopt/limiter.py
+++ b/gadopt/limiter.py
@@ -8,6 +8,7 @@ from firedrake import VertexBasedLimiter, FunctionSpace, TrialFunction, LinearSo
 from firedrake import max_value, min_value, Function
 from firedrake import TensorProductElement, VectorElement, HDivElement, MixedElement, EnrichedElement, FiniteElement
 from firedrake.functionspaceimpl import WithGeometry
+from ufl import triangle
 import numpy as np
 from pyop2.profiling import timed_region, timed_function, timed_stage  # NOQA
 from pyop2 import op2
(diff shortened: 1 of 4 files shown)


dsroberts:
I'm not sure if I'm missing something but of the warnings you've listed, I can only find this one: 
```
UserWarning: Discontinuous Lagrange element requested on quadrilateral, creating DQ element
```
in any of our CI logs, including in the logs from this run. I checked regression test logs from master from a month ago, and only the viscoplastic DG case shows this warning. The rest of the warnings come straight from G-ADOPT. I thought maybe this was a firedrake version thing, but the older logs I checked used Firedrake 2025.4.1. I had sqlite spit out all logs sorted by command for both runs, diffed them and tried to find any instance of 'warning' that didn't appear in both sets of logs. The only additional warnings appeared in the most recent runs from the level set tests:
```
/__w/g-adopt/g-adopt/.venv/lib/python3.12/site-packages/gadopt/level_set_tools.py:835: RuntimeWarning: invalid value encountered in log
```
These seem like sensible changes. but I can't see what, if anything, they've done? Are there any key differences between our CI setup and your installation that might contribute to the extra warnings you see?

angus-g:
I think you'll see them in the "Pytest" step of the usual `test.yml` workflow. The interpolation one in particular is probably important, because they have now implemented that deprecation in master, as far as I'm aware.

dsroberts:
Sigh... *unit* tests. Its right there in the title. Next time I'll read better.

dsroberts (approved):
My reading comprehension has improved and I can definitely see the warnings are gone now. Should the 'UserWarning: Discontinuous Lagrange element requested on quadrilateral, creating DQ element.' caused by [this line](https://github.com/g-adopt/g-adopt/blob/a5f8dd16f133543f4b3883989af2701b63313409/tests/viscoplastic_case_DG/viscoplastic_case_DG.py#L58) be fixed too in a separate PR?

Patol75:
Good question. I am always happy to get rid of warnings wherever possible. I am also generally for specifying `Q` and `DQ` elements instead of `CG` and `DG` when not using simplex meshes, but I also understand that people are more familiar with the latter. Anyway, no reason to extend that discussion here. Thank you for the review, @dsroberts.

PR #258, opened 2025-06-27
Allow material entrainment calculation on "non-Cartesian" meshes

Patol75:
`material_entrainment` is currently restricted to "Cartesian" meshes. It is only a simple change to allow non-Cartesian compatibility, as shown here. I tested and validated the updated functionality on a multi-material version of the 2D cylindrical demo.


Patol75:
Spot on, thank you!

dsroberts (approved):
Nice. Gotta love a PR that adds functionality and reduces code size at the same time.

PR #259, opened 2025-06-29
Testing for adjoint bugs to make sure we dont regress.

sghelichkhani:
We have two ongoing issues, and one from recent BC bug, that we'd like to keep an eye. I am adding those MFE as unit tests for us. 

Note that I am xfailing one of tests as the fix is still not in master.


sghelichkhani:
There is something about `SingleDiskStorageSchedule` that does not like to be used in a sequence of simulations. Possibly due to how disk checkpointing is handled! Although there is nothing obvious that tells me why we are getting the error! Another example of great documentation for the whole thing ... 

For now I have commented disk checkpointing out.

tests/unit/test_adjoints.py
```diff
@@ -0,0 +1,218 @@
+import pytest
+import pickle
+from firedrake import *
+from firedrake.adjoint import *
+from checkpoint_schedules import SingleMemoryStorageSchedule, SingleDiskStorageSchedule
+import numpy as np
+from pathlib import Path
+
+
+def tape_generation_staggered_solves(scheduler):
```
angus-g, line 10: Document what this case is trying to do?

tests/unit/test_adjoints.py
```diff
@@ -0,0 +1,218 @@
...
+
+    continue_annotation()
+
+    # get tape
+    tape = get_working_tape()
+    tape.clear_tape()
+    ret = {}
+
+    if scheduler is not None:
+        if isinstance(scheduler, SingleDiskStorageSchedule):
```
angus-g, line 20: I think we can use `schedule.uses_storage_type(StorageType.DISK)` to be a bit more generic.

tests/unit/test_adjoints.py
```diff
@@ -0,0 +1,218 @@
...
+            enable_disk_checkpointing()
+        tape.enable_checkpointing(scheduler)
+
+    mesh = checkpointable_mesh(UnitSquareMesh(1, 1)) if isinstance(scheduler, SingleDiskStorageSchedule) else UnitSquareMesh(1, 1)
```
angus-g, line 24:
```suggestion
    mesh = UnitSquareMesh(1, 1)
    if scheduler.uses_storage_type(StorageType.DISK):
        mesh = checkpointable_mesh(mesh)
```

tests/unit/test_adjoints.py
```diff
@@ -0,0 +1,218 @@
...
+    u_0 = Function(V).assign(1.0)
+    u = Function(V).assign(u_0)
+    r = Function(V)
+
+    for i in tape.timestepper(iter(range(10))):
+        if i % 3 == 0:
+            r.project(1.01 * u)
+        u.project(r * u)
+
+    J = assemble((u) ** 2 * dx)
```
angus-g, line 37:
```suggestion
    J = assemble(u ** 2 * dx)
```

tests/unit/test_adjoints.py
```diff
@@ -0,0 +1,218 @@
...
+
+    pause_annotation()
+
+    reduced_functional = ReducedFunctional(J, Control(u_0))
+
+    # Storing the diagnostics
+    ret["using scheduler"] = scheduler.__class__.__name__ if scheduler is not None else "None"
```
angus-g, line 44: I don't understand this entry, it just gets popped in the test method anyway?

tests/unit/test_adjoints.py
```diff
@@ -0,0 +1,218 @@
...
+    ret["J1"] = reduced_functional(u_0)
+    ret["dJdm1"] = reduced_functional.derivative().dat.data_ro.copy()
+    return ret
+
+
+def tape_generation_control_invariant_assign(scheduler):
+    """ See bellow test_control_invariant_assign for references """
```
angus-g, line 51: But below you have:

> See above tape_generation_control_invariant_assign for details.

:rofl:

I think this (and the other tape) generation function should document what block structure they're trying to achieve, and the test function can have the bug details.

tests/unit/test_adjoints.py
```diff
@@ -0,0 +1,218 @@
...
+    if isinstance(scheduler, SingleDiskStorageSchedule):
+        enable_disk_checkpointing()
+
+    if scheduler is not None:
+        tape.enable_checkpointing(scheduler)
+
+    mesh = UnitSquareMesh(1, 1)
+
+    if isinstance(scheduler, SingleDiskStorageSchedule):
+        mesh = checkpointable_mesh(mesh)
```
angus-g, line 65: Unify structure with above, or maybe even pull the whole prologue of the tape generation functions out and return a `mesh` compatible with the scheduler after enabling annotation and checkpointing.

tests/unit/test_adjoints.py
```diff
@@ -0,0 +1,218 @@
...
+    ret["residuals"] = residuals
+
+    ret["using scheduler"] = scheduler.__class__.__name__ if scheduler is not None else "None"
+    return ret
+
+
+def tape_generation_DirichletBCc(scheduler):
+    """
+    See bellow test_DirichletBCc for references
+    """
```
angus-g, line 132: As above (there isn't even a `test_DirichletBCc` "bellow")...

tests/unit/test_adjoints.py
```diff
@@ -0,0 +1,218 @@
...
+    This causes the value to be reconstructed incorrectly on the next reverse pass.
+
+    Reference: https://github.com/dolfin-adjoint/pyadjoint/issues/209
+    """
+
+    if isinstance(scheduler, SingleMemoryStorageSchedule) and func == tape_generation_control_invariant_assign:
+        pytest.xfail("pyadjoint issue #209 not yet fixed")
+
+    if isinstance(scheduler, SingleMemoryStorageSchedule) and func == tape_generation_staggered_solves:
+        pytest.xfail("pyadjoint issue #211 not yet fixed")
```
angus-g, line 209:
```suggestion
    if isinstance(scheduler, SingleMemoryStorageSchedule) and func is tape_generation_control_invariant_assign:
        pytest.xfail("pyadjoint issue #209 not yet fixed")

    if isinstance(scheduler, SingleMemoryStorageSchedule) and func is tape_generation_staggered_solves:
        pytest.xfail("pyadjoint issue #211 not yet fixed")
```

tests/unit/test_adjoints.py
```diff
@@ -0,0 +1,218 @@
...
+
+    reference_results = pickle.load(open(Path(__file__).parent.resolve() / "data" / ref_filename, "rb"))
```
angus-g, line 211: Should probably include the method for generating the reference data.

angus-g (changes requested):
Nice, good job writing these up. I'll have a look at why we can't switch between schedulers (I think it's to do with the "single dispatch" mechanism in pyadjoint checkpointing).

angus-g (approved):
Thanks! I think it's probably a good idea to get these in now. When we un-xfail the tests due to upstream fixes we can do any other needed cleanup.

Issue #174, opened 2025-02-26
Include Setonix in HPC testing


angus-g:
Either concurrent with Gadi, or separate?

dsroberts:
Given its a `workflow_dispatch` task, you can add something like this:
```
    inputs:
        system:
          description: "Select a system to test on"
          required: false
          type: choice
          options:
            - gadi
            - setonix
```
Then a job that defines a `matrix` based on that choice, if the user doesn't make a choice, run on both.

drhodrid:
Runs, most test pass, a few analytical tests fail which we will look at in a subsequent pull request.

PR #260, opened 2025-07-02
Integrate `gadopt_hpc_helper` module with longtests

dsroberts:
To address #174, I've come up with an optional add-on for G-ADOPT to assist on deployment on HPC clusters (Currently at https://github.com/dsroberts/gadopt_hpc_helper - will move to g-adopt). There is no need for desktop G-ADOPT users to ever use it, however, I plan to install it into our firedrake modules on Gadi and Setonix (and any other system we end up running on one day). The idea is that it detects what system you're on, and it has knowledge of what queues are available and their properties, what scheduler is in use and how to run parallel jobs. Two of the three entry points are demonstrated in this change, the `gadopt_hpcrun` script, which sort of replaces `qsub` or `mpiexec`, and the `gadopt_hpcrun_async` python function, which is the preferred option for running many batch jobs out of a single python script/make target like the analytical longtests. There is also a `gadopt_hpcrun` function that is just wraps `gadopt_hpcrun_async` in its own event loop so a user doesn't need to worry about dealing with `asyncio`. Given a new firedrake patch release is close, I'll install `gadopt_hpc_helper` into the next firedrake module, so the long tests won't work (and therefore this PR shouldn't be merged) until that's available on both Gadi and Setonix.

In the meantime, I'd appreciate any input on https://github.com/dsroberts/gadopt_hpc_helper, as its going to be the main driver behind the HPC testing now.


Patol75:
I had a quick peek and therefore cannot provide any in-depth comment, but it all looks very well-structured, well-written, and neat!

drhodrid:
This looks exciting! That's my only comment! Probably one for Angus!!

angus-g:
I think it's probably going to be more pleasant to work on things with this in place. I'd be happy with getting this in, then working on the tests later. The HPC tests have consistently "failed" for a while anyway.

drhodrid:
Agreed with @angus-g. Get this in, then we can work on getting things green again!

angus-g (approved):
Awesome, looks pretty reasonable to me! I'll have a look through the HPC helper too, but this is probably fine to go in.

dsroberts:
Thanks, I'll wait for your input on `gadopt_hpc_helper` then. That's moved over to https://github.com/g-adopt/gadopt_hpc_helper now.

angus-g:
I don't really know what kind of feedback you want on the helper. It obviously works, but it's pretty big and you've already designed most of it... If you're happy enough with it, might as well just squash this PR in, unless you are looking for anything specific?

dsroberts:
I guess I wanted to leave that question open. Between this, TSP and `memprof_plotter`, there are a few utilities that live outside of the main G-ADOPT repo that it now depends on. On the one hand, asking for review does create extra work for people, but on the other, saying 'trust me this code is fine' doesn't hold up when working on G-ADOPT, and therefore shouldn't be the standard for its utilities either.

I'm also conscious of the appearance of bias. Much of the code in `gadopt_hpc_helper` came from #194, which was (rightly) never merged into G-ADOPT. Its not the best look that I've come back with a solution to the problem #194 was meant to solve that's external to the G-ADOPT repo and therefore won't be subject to the same standard of review. That doesn't sit entirely right with me, I don't want this approach to be seen as a way to get around having your code reviewed. I mean, level sets could have been implemented as a separate, optional library that G-ADOPT could hook into. Just have G-ADOPT raise an exception if you try and do level set stuff and the `gadopt_level_sets` module isn't present.

Anyway, that's not a discussion for here, but it is something that'd be good to get everyone's thoughts on. I'm happy with `gadopt_hpc_helper` because its author has the same sets of preferences, biases and blind spots that I do...

PR #263, opened 2025-07-16
Fixing the surface integral for dynamic topography average

sghelichkhani:
Fixes an issue with removing the average value of dynamic topography.  Before the surface area was wrongly being calculated using ds(mesh), which gives the total surface of the mesh, whereas we are only interested in the specific `subdomain_id`. 

We now provide the mesh to `self.ds` so that surface area integral would work simply with `self.ds`.

### DIFF
--- a/gadopt/stokes_integrators.py
+++ b/gadopt/stokes_integrators.py
@@ -830,11 +830,11 @@ class BoundaryNormalStressSolver:
             - self.p * fd.Identity(self.dim)
         )
 
-        degree = self._kwargs.get("quad_degree", None)
+        ds_kwargs = {"domain": self.mesh, "degree": self._kwargs.get("quad_degree", None)}
         if self.mesh.extruded and self.subdomain_id in ["top", "bottom"]:
-            self.ds = {"top": fd.ds_t(degree=degree), "bottom": fd.ds_b(degree=degree)}.get(self.subdomain_id)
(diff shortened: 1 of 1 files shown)


gadopt/stokes_integrators.py
```diff
@@ -830,10 +830,11 @@ def setup_solver(self):
             - self.p * fd.Identity(self.dim)
         )
 
+        degree = self._kwargs.get("quad_degree", None)
         if self.mesh.extruded and self.subdomain_id in ["top", "bottom"]:
-            self.ds = {"top": fd.ds_t, "bottom": fd.ds_b}.get(self.subdomain_id)
+            self.ds = {"top": fd.ds_t(degree=degree), "bottom": fd.ds_b(degree=degree)}.get(self.subdomain_id)
         else:
-            self.ds = fd.ds(self.subdomain_id)
+            self.ds = fd.ds(self.subdomain_id, domain=self.mesh, degree=degree)
```
Patol75, line 837:
```python
ds_kwargs = {"domain": self.mesh, "degree": self._kwargs.get("quad_degree", None)}
if self.mesh.extruded and self.subdomain_id in ["top", "bottom"]:
    self.ds = (fd.ds_t if self.subdomain_id == "top" else fd.ds_b)(**ds_kwargs)
else:
    self.ds = fd.ds(self.subdomain_id, **ds_kwargs)
```

PR #267, opened 2025-07-30
Remove "exercise" tags from rendered output

### LINKED ISSUE #248: Add tag for tutorials to seperate exercises.
As above.

angus-g:
It could also be useful to use the "skip-execution" tag on these cells, so they can have stub/invalid code in them. I've written a bit of documentation, including these tags [on the wiki](https://github.com/g-adopt/g-adopt/wiki/Writing-tutorials).

Closes #248

### DIFF
--- a/.github/workflows/render.yml
+++ b/.github/workflows/render.yml
@@ -25,8 +25,8 @@ jobs:
     steps:
       - name: Install Git LFS
         run: |
-          sudo apt update
-          sudo apt install -y git-lfs
+          apt update
+          apt install -y git-lfs
       - uses: actions/checkout@v4
       - name: Install uv
         uses: astral-sh/setup-uv@v5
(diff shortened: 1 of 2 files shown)


w3168 (changes requested):
I agree it looks really neat - thanks! I've looked through it and checked that the changes work locally. i.e. adding Angus' example snippets from the wiki with the exercise tag 1) They show up in a notebook if you just open with jupytext or convert to a notebook. 2) The lines don't execute them when run as a python script, 2) they disappear if you run the additional jupyter nbconver ... 

@angus-g are you happy to merge? Sorry, I wasn't sure how to actually test this as part of the CI and look at the website that has been built. It would be nice to have an example exercise at the end of the base case so we can check that the whole workflow works....

w3168 (approved):

PR #268, opened 2025-07-30
Internal variable via substitue method

w3168:
Start of PR to merge in Internal variable method for viscoelasticity via the substitute method. 

Plan is to be rebased on #225 

Currently base case for stokes, free surface demo and tests should work, as well as some new simple maxwell (and burgers) viscoelastic and elastic loading cases, similar to the existing zhong viscoelastic tests.


w3168:
To do 
- merge with #225 to remove meta class and rejig how solver is setup in base class
- need to undo lots of linting changes
- Work out what to do with order of T and approximation. Do we really want to change all the demos? (if we do should be done in #225 anyway)


In later PR
- modify existing viscoelastic demos to use Internal variable (one thing will be separating out u and p nullspaces)

Patol75:
Thank you, @w3168. Still some work to do, as you pointed out, but we should not be too far! Perhaps we also need to ponder where specific code blocks end up or are called to make the modules as readable as possible.

dsroberts:
In the interest of moving things forward, I'm going to suggest the metaclass stays there for now. I have a pathway forwards that does remove it, but its a more substantial change than belongs in this branch.

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -14,191 +14,237 @@
...
+#     H  = \sum_i \left(\int_{K_i} -\nabla \cdot \boldsymbol{\phi}~   B_{\mu} \rho_0 g u_k  \, dx  + \int_{\partial K_i} \boldsymbol{\phi} \cdot \hat{\boldsymbol{n}}  B_{\mu} \rho_0 g u_k \, ds  \right).
 # \end{equation}
+#
+# Since we allow density and gravity discontinuities through the $DG0$ discretisation, we need to also account for the contribution of surface integrals along interior facets, denoted by $\Gamma$, using
+#
+# \begin{equation}
+#     H  = -\int_{\Omega} \nabla \cdot \boldsymbol{\phi} B_{\mu} \rho_0 g u_k \, dx  + \int_{\Gamma} \boldsymbol{\phi} \cdot B_{\mu} u_k [[\rho_0 g \hat{\boldsymbol{n}}]]  \, ds  + \int_{\partial \Omega_\textrm{top}} \boldsymbol{\phi} \cdot \hat{\boldsymbol{n}} B_{\mu} \rho_0 g u_k \, ds  ,
+# \end{equation}
+#
+# where the jump term is given by
```
drhodrid, line 214: I suspect its worth specifying somewhere above what you really mean by jump term here. I'm kinda clear but also unclear. Remember the tutorial readers will be from a range of backgrounds and they'll need to understand the significance of these jump terms.

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -249,88 +315,95 @@
...
 # association of degrees of freedom with the mesh and pointing to the
-# finite element basis. We will choose Q2-Q1 for the mixed incremental displacement-pressure similar to our mantle convection demos.
-# This is a Taylor-Hood element pair which has good properties for Stokes modelling. We also initialise a discontinuous tensor function
-# space that wil store our previous values of the deviatoric stress, as the gradient of the continous incremental displacement field will
-# be discontinuous.
+# finite element basis. We will choose Q2 for the displacement similar to the velocity field in our mantle convection demos.
+# We also initialise a discontinuous tensor function space that will store our previous values of the internal variable
+# as the gradient of the continous displacement field will be discontinuous. Given that we have purely radial variations in density, viscosity and shear modulus, we choose the DG0 space (i.e., constant within a finite element cell but discontinuous between cells),
 
 # Set up function spaces - currently using the bilinear Q2Q1 element pair:
```
drhodrid, line 322: No longer bilinear Q2Q1 right?

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -249,88 +315,95 @@
...
+u = Function(V, name="displacement")  # field to hold our displacement solution
+
+m = Function(S, name="Internal variable")  # Lagged value of internal variable at previous timestep
+m_list = [m]
 # -
 
 # We can output function space information, for example the number of degrees
-# of freedom (DOF).
+# of freedom (DOF) using `log`, a utility provided by
+# G-ADOPT. (N.b. `log` is equivalent to python's `print` function, except that it simplifies outputs when running simulations in parallel.)
```
drhodrid, line 356:
```suggestion
# G-ADOPT. (N.b. `log` is equivalent to python's `print` function, except that it simplifies outputs for parallel simulations.)
```

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -410,38 +484,41 @@ def initialise_background_field(field, background_values, vertical_tanh_width=20
...
 stokes_bcs = {
     boundary.bottom: {'uy': 0},
-    boundary.top: {'normal_stress': ice_load, 'free_surface': {'delta_rho_fs': density - exterior_density}},
+    boundary.top: {'free_surface': {'normal_stress': ice_load}},
     boundary.left: {'ux': 0},
     boundary.right: {'ux': 0},
 }
 
-gd = GeodynamicalDiagnostics(z, density, boundary.bottom, boundary.top)
+# gd = GeodynamicalDiagnostics(z, density, boundary.bottom, boundary.top)
```
drhodrid, line 494: Delete line?

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -457,30 +534,30 @@ def initialise_background_field(field, background_values, vertical_tanh_width=20
...
+            for i, m_i in enumerate(m_list):
+                checkpoint.save_function(m_i, name=f"internal_variable_{i}")
 
     # Log diagnostics:
-    plog.log_str(
-        f"{timestep} {float(time)} {float(dt)} "
-        f"{gd.u_rms()} {gd.u_rms_top()} {gd.ux_max(boundary.top)} "
-        f"{displacement.dat.data[:, 1].min()} {displacement.dat.data[:, 1].max()}"
-    )
+#    plog.log_str(
```
drhodrid, line 552: Uncomment or delete?

drhodrid (commented):
Great stuff Will. Some thoughts to go with my more detailed (minor) comments:
1. Provide a little more of a background at the top saying what exactly this case is, for context (i.e. compressible -- a 2D version of Weerderstein etc...)? 
2. Edit file such that line length is shorter. I can't remember the limit (7? characters?)?
3. Add some exercises -- my thoughts -- 1. make model incompressible and compare a specific diagnostics -- tell the users what they should expect to see and give them pointers on how to do it; 2. your idea around Cathles parameter -- can be done individually or as a group.
4. I liked the maths and description at the top -- and I'd keep it. But during the session in Hobart, we may need to request that attendees at least read the introduction to this tutorial beforehand -- such that they understand the physical equations, then we can start from the weak form on the day or something like that.

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -2,490 +2,739 @@
...
-#
-# This expression for the stress is similar to that relevant for mantle convection. We are solving for incremental
-# displacement instead of velocity, but the only difference between these two functions is the timestep multiplication factor.
-# We also have a modified viscosity based on the timestep and the Maxwell time. Finally, the stress history is included as the
-# last term is the deviatoric stress from the previous timestep multiplied by a prefactor involving the timestep and the Maxwell time.
-# Note that in the small dt limit (i.e. $dt$ << $\alpha$) the effective viscosity tends towards the shear modulus. i.e. we are solving the elastic equations.
+# where $\eta_i$ is the non-dimensional viscosity of the the internal variable
+# $\boldsymbol{m}_i$. Recombining the final system of equations is
+#
+# \begin{multline}
```
sghelichkhani, line 364: I am not sure if this is something just for my system, but I get  `ParseError` for multline. Are you sure this runs fine?

w3168: Hmm I think it works on my jupytext notebook but I can change this if it is not general!

gadopt/diagnostics.py
```diff
@@ -36,6 +38,7 @@ class GeodynamicalDiagnostics:
       T_min: Minimum temperature in domain
       T_max: Maximum temperature in domain
       ux_max: Maximum velocity (first component, optionally over a given boundary)
+      uk_min: Minimum velocity (vertical component, optionally over a given boundary)
```
Patol75: Is `uk` a reasonable choice here? I am wondering if it is obvious to everyone that it represents the vertical component. Could `uv` be better? Or should we be more explicit in the naming (e.g. `u_vert_min`)?

gadopt/diagnostics.py
```diff
@@ -51,7 +54,20 @@ def __init__(
     ):
         mesh = extract_unique_domain(z)
 
-        self.u, self.p = z.subfunctions[:2]
+        # Allows sub-iv solver (just disp no mixed space)
```
Patol75: Please write explicitly what `sub-iv` means.

tests/glacial_isostatic_adjustment/test_iv_ve_fs.py
Patol75: Same idea as above, a little more verbosity in the file name would be helpful.

gadopt/momentum_equation.py
```diff
@@ -172,6 +191,43 @@ def momentum_source_term(
...
+        if type(rho0.function_space()._mesh) is ExtrudedMeshTopology:
+            dS = dS_h
+        else:
+            dS = eq.dS
+        F = B_mu("+") * jump(rho0) * u_r("+") * g("+") * dot(eq.test("+"), eq.n("+")) * dS
+    if eq.approximation.compressible_adv_hyd_pre:
+        # Include body integral after i.b.p of hydrostatic prestress advection term
+        # Analytical solution from Cathles 2024 Eq 2b doesn't include prestress
+        # so we neglect this term but keep the free surface term that accounts for
+        # viscous feedback at isostatic equibrium
```
Patol75: `equilibrium`

gadopt/stokes_integrators.py
```diff
@@ -692,6 +710,120 @@ def solve(self) -> None:
...
+        self,
+        solution: fd.Function,
+        approximation: SmallDisplacementViscoelasticApproximation,
+        /,
+        *,
+        m_list: list,
+        dt: float,
+        **kwargs,
+    ) -> None:
+        self.m_list = m_list
```
Patol75: I do not think it is good practice to include in a variable/argument name its expected type.

gadopt/stokes_integrators.py
```diff
@@ -692,6 +710,120 @@ def solve(self) -> None:
...
+        approximation.mu = approximation.viscosity[0]/(approximation.maxwell_times[0]+dt)
+        super().__init__(solution, approximation, dt=dt, **kwargs)
+
+    def set_equations(self) -> None:
+        self.strain = self.approximation.deviatoric_strain(self.solution)
+
+        if len(self.m_list) != len(self.approximation.maxwell_times):
+            raise ValueError("Number of internal variables and corresponding Maxwell times must be consistent")
+
+        m_new_list = [self.update_m(m, alpha) for m, alpha in zip(self.m_list, self.approximation.maxwell_times)]
```
Patol75: `m_integrated` or `m_next`?
How about `maxwell_time` instead of `alpha`?

gadopt/approximations.py
```diff
@@ -413,12 +421,66 @@ def buoyancy(self, p, T):
...
+      shear_modulus: shear modulus
+      viscosity:     viscosity
+      g:             gravitational acceleration
+      B_mu:          Nondimensional number describing ratio of buoyancy to elastic
+                     shear strength used for nondimensionalisation.
+                     $$ B_{\\mu} = \frac{\bar{\rho} \bar{g} L}{\bar{\\mu}}$,
+                     where $\bar{\rho}$ is a characteristic density scale (kg / m^3),
+                     $\bar{g}$ is a characteristic gravity scale (m / s^2),
+                     $L$ is a characteristic length scale, often Mantle depth (m),
+                     $\\mu$ is a characteristic shear modulus (Pa).
```
Patol75: This number is quite similar in its definition to the [Cauchy number](https://www.wikiwand.com/en/articles/Cauchy_number). But that is not the same number, right?

gadopt/approximations.py
```diff
@@ -413,12 +421,66 @@ def buoyancy(self, p, T):
...
+        self.density = ensure_constant(density)
+        self.shear_modulus = ensure_constant(shear_modulus)
+        self.viscosity = ensure_constant(viscosity)
+        self.g = ensure_constant(g)
+        self.B_mu = ensure_constant(B_mu)
+
+    def buoyancy(self, displacement):
+        # Buoyancy term due to the advection of the background density field
+        # written on RHS of equations
+        return -self.B_mu * self.g * -inner(displacement, grad(self.density))
```
Patol75: Any reason to keep two `-` signs?

gadopt/approximations.py
```diff
@@ -466,15 +531,204 @@ def prefactor_prestress(self, dt):
...
+        if len(viscosity) != len(shear_modulus):
+            raise ValueError("Length of viscosity and shear modulus lists must be consistent")
+
+        super().__init__(density, shear_modulus, viscosity, **kwargs)
+        self.maxwell_times = [ensure_constant(visc / mu) for visc, mu in zip(self.viscosity, self.shear_modulus)]
+        self.mu0 = ensure_constant(sum(self.shear_modulus))
+
+    def div_u(self, u: Function) -> ufl.core.expr.Expr:
+        dim = len(u)
+        return div(u) * Identity(dim)
```
Patol75: I am not convinced this method is necessary. Rather, you could define `self.identity` and use it wherever necessary.

gadopt/approximations.py
w3168: Not really I guess in my head I was still thinking of the term as `-B_mu * g * rho_1`  where `rho_1 = -grad u \cdot grad rho_0`

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -2,490 +2,739 @@
...
+shear_modulus_scale = 1e11
+viscosity_scale = 1e21
+characteristic_maxwell_time = viscosity_scale / shear_modulus_scale
+gravity_scale = 9.81
+B_mu = Constant(density_scale * D * gravity_scale / shear_modulus_scale)
+
+density_values_tilde = np.array(density_values)/density_scale
+shear_modulus_values_tilde = np.array(shear_modulus_values)/shear_modulus_scale
+viscosity_values_tilde = np.array(viscosity_values)/viscosity_scale
+bulk_modulus_values_tilde = 2 * shear_modulus_values_tilde
```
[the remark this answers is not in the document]
w3168: added a short explanation - I think it's typical ish for mantle for bulk = 2x shear

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -2,490 +2,739 @@
...
 
-def initialise_background_field(field, background_values, vertical_tanh_width=20e3):
-    profile = background_values[0]
-    sharpness = 1 / vertical_tanh_width
-    depth = X[1] - D
-    for i in range(1, len(background_values)):
-        centre = radius_values[i] - radius_values[0]
-        mag = background_values[i] - background_values[i-1]
-        profile += step_func(depth, centre, mag, increasing=False, sharpness=sharpness)
+def initialise_background_field(field, background_values):
```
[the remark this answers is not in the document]
w3168: moved this to utility

gadopt/approximations.py
w3168: this was removed by Dale

gadopt/diagnostics.py
w3168: renamed to `uv_min` I think this is all going to get looked at again in #299 so I think this ok for now.

gadopt/diagnostics.py
w3168: expanded comment

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -2,491 +2,665 @@
...
-# The corresponding elastic constitutive equation is
+# The GIA community generally model the mantle as a Maxwell solid. The
+# conceptual picture is a spring and a dashpot connected together in series
+# ([Ranalli, 1995](https://link.springer.com/book/9780412546709)).
+# For this viscoelastic model the elastic and viscous stresses are the same
+# but the total displacements combine.
+#
+#
+# We follow the internal variable formulation adopted by Al-Attar and
+# Tromp (2014) and Crawford et al. (2017, 2018), in which viscoelastic
```
Patol75: Missing DOI links?

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -2,491 +2,665 @@
...
-# 'incremental displacement', thus recasting the problem in terms of $\textbf{u}_{inc}^n = \textbf{u}^n - \textbf{u}^{n-1}$,
-# where subscripts refer to time levels $t$ and $t - \Delta t$ respectively.
-# The incremental strain $\Delta \boldsymbol{\epsilon}$ is
+# One of the key differences with the mantle convection demos is that the
+# constitutive equation now depends on time. We discretise the internal
+# variable evolution equation in time using the implicit Backward Euler (BE)
+# scheme. This choice allows us to take timesteps larger than the characteristic
+#  Maxwell time without compromising numerical stability. Such flexibility is
+# particularly advantageous when the timescale of glacial loading is
+# substantially slower than the Maxwell time -- as is often the case in
```
Patol75: slower -> shorter

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -2,491 +2,665 @@
...
+# and the contributibution of the internal variable from the previous
+# timestep.
+#
+# Implicit terms involving the unknown displacement at the next time step,
+#  $\boldsymbol{u}^{n+1}$, have been collected on the left-hand side and
+# explicit terms known from the current time step are on the right-hand side.
+# Finding the new values of $\boldsymbol{u}^{n+1}$ requires solving a linear
+# system at each timestep, which is solved in Firedrake through PETSc's
+# comprehensive linear algebra library. For more details on the numerical
+# discretisation please refer to Scott et al. (2025).
```
Patol75: Missing DOI?

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -2,491 +2,665 @@
...
 
-# We next set up our output, in VTK format. This format can be read by programs like pyvista and Paraview.
+# We next set up our output, in VTK format. This format can be read by programs
+# like pyvista and Paraview. We also create a function to store the velocity
+# field at each timestep.
 
 # +
+# initialise velocity and old displacement functions for plotting
+velocity = Function(u, name="velocity")
+disp_old = Function(u, name="old_disp").assign(u)
```
Patol75: Do you need the `assign` part here?

gadopt/utility.py
```diff
@@ -7,13 +7,14 @@
...
 from firedrake import as_vector, SpatialCoordinate, Constant, max_value, min_value, dx, assemble, tanh
-from firedrake import op2, VectorElement, DirichletBC, utils, interpolate
+from firedrake import op2, VectorElement, DirichletBC, utils, interpolate, conditional
 from firedrake.ufl_expr import extract_unique_domain
 import ufl
 import time
 from ufl.corealg.traversal import traverse_unique_terminals
 from firedrake.petsc import PETSc
 from mpi4py import MPI
+from numbers import Number
```
Patol75: Is this necessary, or can we use `float` instead?

gadopt/utility.py
```diff
@@ -588,3 +589,49 @@ def get_boundary_ids(mesh) -> SimpleNamespace:
...
+
+        for i in range(nz):
+            layer_heights.append(dz)
+
+    return layer_heights
+
+
+def initialise_background_field(field, background_values, X, radii, shift=0):
+    '''Initialises discontinuous field with sharp jumps at rheological boundaries'''
+    for i in range(0, len(background_values)):
```
Patol75: `for i in range(len(background_values)):`

gadopt/diagnostics.py
```diff
@@ -102,3 +118,14 @@ def ux_max(self, boundary_id=None) -> float:
             ux_data = self.u.dat.data_ro[:, 0]
 
         return self.u.comm.allreduce(ux_data.max(), MPI.MAX)
+
+    def uv_min(self, boundary_id=None) -> float:
+        "Minimum value of vertical component of velocity/displacement"
+        self.uv.interpolate(vertical_component(self.u))
+        if boundary_id:
+            bcu = DirichletBC(self.uv.function_space(), 0, boundary_id)
+            uv_data = self.uv.dat.data_ro_with_halos[bcu.nodes]
```
Patol75: This could run into the same issue as #319.

Patol75 (changes requested):
PR is in great shape; well done, everyone. Some additional comments and suggestions.

gadopt/diagnostics.py
w3168: doesn't adding 

```
uv_data.min(initial=np.inf)
```

fix this? I think this was the fix suggedted in #299 ...

gadopt/approximations.py
```diff
@@ -413,12 +421,66 @@ def buoyancy(self, p, T):
...
 
 class SmallDisplacementViscoelasticApproximation:
-    """Expressions for the small displacement viscoelastic approximation.
+    """Base class for viscoelasticity assuming small displacements.
+
+    By assuming a small displacement with respect to mantle depth
+    we can linearise the problem, assuming a perturbation away from a reference state.
+
+    For background and derivation of the formulation please see the equations and
+    references provided in Scott et al 2025.
```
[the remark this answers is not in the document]
w3168: It's got the reference underneath so probably ok?

gadopt/approximations.py
```diff
@@ -413,12 +421,66 @@ def buoyancy(self, p, T):
...
-    away from a reference state. We assume a Maxwell viscoelastic rheology, i.e.
-    stress is the same but viscous and elastic strains combine linearly. We follow
-    the approach by Zhong et al. 2003 redefining the problem in terms of incremental
+    def rho_continuity(self):
+        # StokesSolverBase in stokes_integrators.py currently requires rho_continuity
+        # from approximation. This is only strictly necessary for quasi-compressible
+        # approximations in Mantle convection e.g. TALA and ALA, but is used in setting up
+        # the continuity equation for the IncompressibleMaxwellApproximation to use similar
+        # code for the incompressible viscous mantle convection in StokesSolverBase.
+        return 1
```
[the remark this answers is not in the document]
w3168: I think let's leave this for now. The gia demos in hobart will all be direct so we don't need to separate out the solver strategy of `InternalVariableSolver` compared with the `StokesSolverBase` just yet

gadopt/approximations.py
```diff
@@ -413,12 +421,66 @@ def buoyancy(self, p, T):
...
+
+
+class IncompressibleMaxwellApproximation(SmallDisplacementViscoelasticApproximation):
+    """Incompressible maxwell rheology via the incremental displacement formulation.
+
+    This class implements incompressible Maxwell rheology similar to the approach
+    by Zhong et al. 2003. The linearised problem is cast in terms of incremental
     displacement, i.e. velocity * dt where dt is the timestep. This produces a mixed
     stokes system for incremental displacement and pressure which can be solved in
```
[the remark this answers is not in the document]
w3168: done

gadopt/approximations.py
```diff
@@ -466,15 +531,204 @@ def prefactor_prestress(self, dt):
...
+    arranged in parallel. This formulation provides a compact, flexible and convenient
+    means to incorporate transient rheology into viscoelastic deformation models:
+    using a single internal variable is equivalent to a simple Maxwell material;
+    two correspond to a Burgers model with two characteristic relaxation frequencies;
+    and using a series of internal variables permits approximation of a continuous
+    range of relaxation timescales for more complicated rheologies.
+
+    This class implements the substiution method where the time-dependent internal
+    variable equation is substituted into the momentum equation assuming a Backward
+    Euler time discretation. Therefore, the displacement field is the only unknown.
```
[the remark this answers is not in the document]
w3168: done

gadopt/utility.py
w3168: done

gadopt/utility.py
w3168: done

gadopt/utility.py
```diff
@@ -588,3 +589,49 @@ def get_boundary_ids(mesh) -> SimpleNamespace:
...
+        field.interpolate(
+            conditional(vertical_component(X) + shift >= radii[i+1],
+                        conditional(vertical_component(X) + shift <= radii[i],
+                                    background_values[i], field), field))
```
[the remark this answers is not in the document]
w3168: I made the first one just `>` and the tests still pass!

demos/glacial_isostatic_adjustment/base_case/base_case.py
```diff
@@ -2,491 +2,665 @@
...
+# system at each timestep, which is solved in Firedrake through PETSc's
+# comprehensive linear algebra library. For more details on the numerical
+# discretisation please refer to Scott et al. (2025).
+#
 
 # This example
 # -------------
-# We will simulate a viscoelastic loading and unloading problem based on a 2D version of the test case presented in [Weerdesteijn et al. (2023)](https://doi.org/10.1029/2022GC010813).
+# We will simulate a viscoelastic loading and unloading problem based on a 2D
+# version of the test case presented in Weerdesteijn et al. (2023).
```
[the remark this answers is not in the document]
w3168: added back in

demos/glacial_isostatic_adjustment/base_case/base_case.py
w3168: removed

gadopt/__init__.py
```diff
@@ -4,8 +4,11 @@
...
     AnelasticLiquidApproximation,
     BoussinesqApproximation,
+    QuasiCompressibleInternalVariableApproximation,
+    CompressibleInternalVariableApproximation,
     ExtendedBoussinesqApproximation,
-    SmallDisplacementViscoelasticApproximation,
+    IncompressibleMaxwellApproximation,
+    MaxwellApproximation,
     TruncatedAnelasticLiquidApproximation,
 )
```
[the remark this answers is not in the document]
w3168: done

gadopt/approximations.py
w3168: removed minus signs

gadopt/stokes_integrators.py
```diff
@@ -692,6 +710,120 @@ def solve(self) -> None:
...
+        approximation: SmallDisplacementViscoelasticApproximation,
+        /,
+        *,
+        m_list: list,
+        dt: float,
+        **kwargs,
+    ) -> None:
+        self.m_list = m_list
+        # Effective viscosity THIS IS A HACK. need to update SIPG terms for compressibility?
+        approximation.mu = approximation.viscosity[0]/(approximation.maxwell_times[0]+dt)
```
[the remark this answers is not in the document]
w3168: I've just implemented this as `approximation.effective_viscosity(dt)` and corrected it for the burgers case where it should have been a summation including the extra invernal variable

tests/3d_weerdesteijn/generate_surface_meshes.sh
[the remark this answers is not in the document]
w3168: No I'll remove it! I added it at one stage because I was using them for higher resolution for the paper but the repo isnt the place for them at the moment.

gadopt/approximations.py
```diff
@@ -412,15 +420,93 @@ def buoyancy(self, p, T):
...
-    away from a reference state. We assume a Maxwell viscoelastic rheology, i.e.
-    stress is the same but viscous and elastic strains combine linearly. We follow
-    the approach by Zhong et al. 2003 redefining the problem in terms of incremental
+    def rho_continuity(self):
+        # StokesSolverBase in stokes_integrators.py currently requires rho_continuity
+        # from approximation. This is only strictly necessary for quasi-compressible
+        # approximations in Mantle convection e.g. TALA and ALA, but is used in setting up
+        # the continuity equation for the IncompressibleMaxwellApproximation to use similar
+        # code for the incompressible viscous mantle convection in StokesSolverBase.
+        return 1
```
Patol75: `rho_continuity` has two purposes in `StokesSolverBase`: to define an element of the application context and to define an equation attribute used in `divergence_term`. For the latter, this is an easy fix: the `Equation` instance holds the approximation, so `rho_continuity` can be directly called in `divergence_term`. For the former, one way around it is to define an empty application context in `StokesSolverBase` and then have each solver add to it based on its requirements. Then we should be able to get rid of this method.

gadopt/approximations.py
```diff
@@ -432,49 +518,288 @@ class SmallDisplacementViscoelasticApproximation:
 
     N.b. that the implentation currently assumes all terms are dimensional.
```
Patol75: `implementation`

Is it still true that all terms are dimensional only?

gadopt/approximations.py
```diff
@@ -432,49 +518,288 @@ class SmallDisplacementViscoelasticApproximation:
...
+
+    def stress(self, u, **kwargs) -> ufl.core.expr.Expr:
+        internal_variables = kwargs.get("internal_variables", None)
+        if internal_variables is None:
+            raise KeyError(
+                f"The internal_variables kwarg must be provided for stress() in {self.__class__.__name__}"
+            )
+
+        div_u = div(u) * Identity(len(u))
+        d = self.deviatoric_strain(u)
```
Patol75: These two do not need to be their own variables.

gadopt/approximations.py
```diff
@@ -432,49 +518,288 @@ class SmallDisplacementViscoelasticApproximation:
...
+
+    def buoyancy(self, displacement: Function) -> ufl.core.expr.Expr:
+        # Compressible part of buoyancy term due to the density perturbation
+        # written on rhs of equations
+        buoyancy = super().buoyancy(displacement)
+        # By default this term is included but in some cases e.g. to reproduce
+        # simplified analytical cases such as the Cathles 2024 benchmark in
+        # /tests/glacial_isostatic_adjustment/iv_ve_fs.py we need to remove
+        # this effect.
+        buoyancy += -self.B_mu * self.g * -self.density * div(displacement)
```
Patol75: Two minus signs

gadopt/approximations.py
```diff
@@ -432,49 +518,288 @@ class SmallDisplacementViscoelasticApproximation:
...
+        return buoyancy
+
+    def compressible_adv_hyd_pre(
+        self, u_r: Function | ufl.core.expr.Expr
+    ) -> ufl.core.expr.Expr:
+        # Hydrostatic prestress advection term which is applied
+        # as a `normal_stress` boundary condition when the `free_surface` tag is
+        # specified in the `bcs` dictionary of the `InternalVariableSolver`
+        # through the `set_free_surface_boundary` method.
+        return self.B_mu * self.density * self.g * u_r
```
Patol75: This method is identical to `hydrostatic_prestress_advection` in the parent class, but it has a different name. Why?

gadopt/approximations.py
```diff
@@ -432,49 +518,288 @@ class SmallDisplacementViscoelasticApproximation:
...
+    inverse modelling of post-seismic deformation. Geophysical Journal International,
+    ggw414.
+
+    Crawford, O., Al-Attar, D., Tromp, J., Mitrovica, J. X., Austermann, J., &
+    Lau, H. C. (2018). Quantifying the sensitivity of post-glacial sea level change
+    to laterally varying viscosity. Geophysical journal international, 214(2), 1324-1363.
+
+    Automated forward and adjoint modelling of viscoelastic deformation of the solid
+    Earth.  Scott, W.; Hoggard, M.; Duvernay, T.; Ghelichkhan, S.; Gibson, A.;
+    Roberts, D.; Kramer, S. C.; and Davies, D. R. EGUsphere, 2025: 1–43. 2025.
```
Patol75: `This is a helper class to simplify setting up (compressible) Maxwell rheology using the internal variable approach. For more details, refer to the documentation of the parent class.`

Patol75 (changes requested):
Some final comments from me. Looking at the demo now.

gadopt/approximations.py
w3168: Yeah sorry I never ended up incorporating your nondimensional changes to the old way of zhong doing things

gadopt/approximations.py
```diff
@@ -432,49 +518,288 @@ class SmallDisplacementViscoelasticApproximation:
...
+        ]
+        self.mu0 = ensure_constant(sum(self.shear_modulus))
+
+    def deviatoric_strain(self, u: Function) -> ufl.core.expr.Expr:
+        dim = len(u)
+        e = sym(grad(u))
+        # N.b. for 2d simulations dividing by 1/3 (instead of 1/2) may be slightly
+        # inconsistent but analytical tests in tests/viscoelastic_internal_variable/
+        # are setup assuming 3D geometry.
+        return e - 1 / 3 * tr(e) * Identity(dim)
```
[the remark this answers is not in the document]
w3168: ok let's work on that after this!

gadopt/approximations.py
w3168: It's the integrated by parts term which has the same form. I've changed it so it returns the hydrostatic prestress advection expression and updated the comment a little.

gadopt/momentum_equation.py
```diff
@@ -172,6 +191,42 @@ def momentum_source_term(
     return -F
 
 
+def advection_hydrostatic_prestress_term(
+    eq: Equation, trial: Argument | ufl.indexed.Indexed | Function
+) -> Form:
+    # Advection of background hydrostatic pressure used in linearised
+    # GIA simulations. This method implements the body integral and
```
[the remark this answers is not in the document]
w3168: ok

gadopt/stokes_integrators.py
```diff
@@ -686,12 +701,147 @@ def solve(self) -> None:
...
+            UFL expression for the updated internal variable using Backward Euler
+        """
+        m_new = (m + self.dt / maxwell_time * self.strain) / (1 + self.dt / maxwell_time)
+        return m_new
+
+    def solve(self) -> None:
+        super().solve()
+        # Update internal variable term for using as a RHS explicit forcing in the next timestep
+        for m, alpha in zip(self.internal_variables, self.approximation.maxwell_times):
+            m.interpolate(self.update_m(m, alpha))
```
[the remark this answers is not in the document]
w3168: ok!

Patol75 (approved):
🚀 🚀 🚀

PR #269, opened 2025-07-31
Implement a non-recursive Makefile structure

angus-g:
## Context

I'll start off by saying that this is more of a request for feedback than a definite feature that should go in. I'll definitely take the blame for implementing a recursive Make system for running the tests/demos. There are some advantages:
- Makefiles for a given directory are fairly standalone
- things like parallel make still work just fine

But I also see some disadvantages:
- incorporating a case into the overall build system requires a bit of "global state" in the `makefile.cases.inc` file
- we can't record dependencies between separate cases (i.e. saying that dynamic topography requires the base case)
- the rules enforced some structure on filenames (tutorial notebooks having the same name as their containing directory)

## This PR

I've attempted to restructure the system with a bit of retrospective context about what we're actually using it for. In particular, I think it's a bit more logical for adding new cases (and demos). Importantly, we can also put arbitrary dependencies between different targets and cases. However, the structure of an individual Makefile does radically change.

I've written a new documentation page giving a bit of an overview of the system, and how to approach it from the viewpoint of adding a new case: https://gist.github.com/angus-g/8ba3b2374a8abd1920f99e02e06e2b28

To get an idea of how the files themselves are structured, I'd probably recommend reading the base [`Makefile`](https://github.com/g-adopt/g-adopt/blob/angus-g/makefile-restructure/Makefile), then [`demos/Makefile`](https://github.com/g-adopt/g-adopt/blob/angus-g/makefile-restructure/demos/Makefile), [`demos/mantle_convection/Makefile`](https://github.com/g-adopt/g-adopt/blob/angus-g/makefile-restructure/demos/mantle_convection/Makefile) and finally [`demos/mantle_convection/base_case/Makefile`](https://github.com/g-adopt/g-adopt/blob/angus-g/makefile-restructure/Makefile). The "category-level" files are vaguely the same, and largely won't require interaction (except for adding to a single line for adding new cases), but the Makefile for the base case is where the largest changes are.

At the moment, I'm not settled on the level of automation (magic) to include, so I haven't got full feature parity across the cases: you can't run `make clean` in an individual directory. The base case does show a possible implementation of this, using the `$(CLEAN_$(d))` target. 

## What I'm looking for

Ultimately, it would be nice to know whether this at all desirable. We could instead just live with running some of our test cases several times. We have 3 or 4 re-runs across the test suite currently, but this is the only strategy we have at the moment as we can't set dependencies between cases. Nothing is set in stone, and if there's something you violently disagree with, I'm open to changing it!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,7 @@
 *.vtu
 *.h5
 __pycache__/
+
+# test artifacts
+.done
+params.log
--- a/.rules.mk
+++ b/.rules.mk
@@ -82,3 +82,31 @@ endef
 
 %.ipynb: %.py
 	jupytext --to ipynb --execute $< --run-path $(dir $(abspath $<))
+
+# recurse into subdirs
+define include_subdir =
+dir := $(dir)
+include $$(dir)/Makefile
+endef
+
+# for defining the targets within e.g. demos or tests
+# this relies on make_targets having been called within the
+# directory within the "usual" path
+define subdir_targets =
+.PHONY: $(1) clean-$(1)
+$(1):
+	$$(MAKE) -C .. $(2)/$(1)
+
+clean-$(1):
+	$$(MAKE) -C .. clean-$(2)/$(1)
+endef
+
+# for defining the targets of the current directory
+define make_targets =
+.PHONY: $(1) clean-$(1)
+$(1): $$(TGT_$(1))
+
+clean-$(1):
+	rm -f $$(CLEAN_$(1))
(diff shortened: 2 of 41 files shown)


Patol75:
Thank you for this nice proposal, @angus-g. I am not against changing the Makefile logic, as long as it remains reasonably comprehensible. The attached documentation page is very helpful, and so long as we can keep it findable, then I would support this PR if you believe the new logic will bring improvements and prevent long-term headaches.

PR #270, opened 2025-08-04
Update multi-material demos and tests

Patol75:
In this PR, I revisit all existing multi-material demos and tests, improving them based on my experiences with the software since the test suite was introduced. Notable changes:
- demos are now tested against expected values
- temperature field defined on a discontinuous space
- better mesh refinements throughout all demos and tests
- more meaningful simulation run times

The main goals are robustness and reliability of the testing framework and improved clarity of demos.

New tests and demos will be introduced in a follow-up PR.

### DIFF
--- a/tests/multi_material/benchmarks/trim_2023/simulation.py
+++ b/tests/multi_material/benchmarks/trim_2023/simulation.py
@@ -122,7 +122,7 @@ checkpoint_restart = 0
 # Insufficient mesh refinement can lead to unwanted motion of material interfaces.
 domain_dims = (1, 1)
 mesh_gen = "firedrake"
-mesh_elements = (256, 256)
+mesh_elements = (128, 128)
 
 # Parameters to initialise level set
 callable_args = (
--- a/tests/multi_material/benchmarks/van_keken_1997_thermochemical/mesh/mesh.geo
+++ b/tests/multi_material/benchmarks/van_keken_1997_thermochemical/mesh/mesh.geo
@@ -1,11 +1,11 @@
-horizontal_resolution = 0.02;
+horizontal_resolution = 0.03125;
 
 domain_dimension_x = 2;
 domain_dimension_y = 1;
 
 bottom_thickness = 0.12;
 top_thickness = 0.12;
-layer_resolution = 0.004;
+layer_resolution = 0.008;
 
 Point(1) = {0, 0, 0, horizontal_resolution};
 Point(2) = {domain_dimension_x, 0, 0, horizontal_resolution};
@@ -16,7 +16,7 @@ Extrude {0, bottom_thickness, 0} {
   Curve{1}; Layers {bottom_thickness / layer_resolution}; Recombine;
 }
 Extrude {0, domain_dimension_y - bottom_thickness - top_thickness, 0} {
-  Curve{2}; Layers {19}; Recombine;  // Vertical resolution: 0.04
+  Curve{2}; Layers {16}; Recombine;  // Vertical resolution: 0.0475
 }
 Extrude {0, top_thickness, 0} {
   Curve{6}; Layers {top_thickness / layer_resolution}; Recombine;
--- a/tests/multi_material/test_multi_material.py
+++ b/tests/multi_material/test_multi_material.py
@@ -65,7 +65,7 @@ diagnostics = {
         (lambda data: abs(data["max_visc"][-1] - 1.92), operator.le, 6e-2),
     ],
     "trim_2023": [
-        (lambda data: abs(data["rms_velocity"][-1] - 157.0), operator.le, 4e0),
+        (lambda data: abs(data["rms_velocity"][-1] - 157.0), operator.le, 6e0),
     ],
     "van_keken_1997_isothermal": [
         (lambda data: abs(max(data["rms_velocity"]) - 3.1e-3), operator.le, 4e-5),
@@ -82,7 +82,7 @@ diagnostics = {
                 - 487.0
             ),
             operator.le,
-            5e1,
+            4e1,
         )
     ],
 }


demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -158,101 +156,106 @@
...
 # -
 
+# Let us now verify that the material fields have been correctly initialised. We plot
+# the compositional Rayleigh number across the domain.
+
+# + tags=["active-ipynb"]
+# fig, axes = plt.subplots()
+# axes.set_aspect("equal")
+# contours = tricontourf(
+#     Function(psi).interpolate(Ra_c), levels=linspace(0.0, 1.0, 11), axes=axes
```
angus-g, line 166: Oops, `Ra_c` doesn't exist. I guess you mean `RaB`?

demos/multi_material/thermochemical_buoyancy/thermochemical_buoyancy.py
```diff
@@ -161,130 +160,150 @@
...
-    boundary.top: {"uy": 0},
-    boundary.left: {"ux": 0},
-    boundary.right: {"ux": 0},
-}
-temp_bcs = {boundary.bottom: {"T": 1}, boundary.top: {"T": 0}}
+# + tags=["active-ipynb"]
+# fig, axes = plt.subplots()
+# axes.set_aspect("equal")
+# contours = tricontourf(
+#     Function(psi).interpolate(Ra_c), levels=linspace(0.0, 4.5e5, 11), axes=axes
```
angus-g, line 170: As with the other one, I guess this should be `RaB`, not `Ra_c`?

angus-g (commented):
Couple of failures from the [`render.yml` run](https://github.com/g-adopt/g-adopt/actions/runs/16766835867/job/47473473765)

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -62,24 +64,24 @@
...
-mesh = RectangleMesh(nx, ny, lx, ly, quadrilateral=True)
-mesh.cartesian = True
-boundary = get_boundary_ids(mesh)
+mesh = RectangleMesh(*mesh_elements, *domain_dims, quadrilateral=True)
+mesh.cartesian = True  # Tag the mesh as Cartesian to inform other G-ADOPT objects
+boundary = get_boundary_ids(mesh)  # Object holding references to mesh boundary IDs
 
-V = VectorFunctionSpace(mesh, "CG", 2)  # Velocity function space (vector)
-W = FunctionSpace(mesh, "CG", 1)  # Pressure function space (scalar)
+V = VectorFunctionSpace(mesh, "Q", 2)  # Velocity function space (vector)
```
drhodrid, line 74: This convention is different from our previous demos. I'd argue we should be consistent. So perhaps reverting to CG is the least work? The fact that we are using quads is implicit from the mesh right (quadrilateral=True) and I find CG more intuitive.

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -62,24 +64,24 @@
...
+W = FunctionSpace(mesh, "Q", 1)  # Pressure function space (scalar)
 Z = MixedFunctionSpace([V, W])  # Stokes function space (mixed)
-Q = FunctionSpace(mesh, "CG", 2)  # Temperature function space (scalar)
+Q = FunctionSpace(mesh, "DQ", 2)  # Temperature function space (scalar)
```
drhodrid, line 77: "Temperature" function space? Does this example have a temperature?

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -131,25 +133,21 @@
...
+# of equations is non-dimensional and only includes compositional buoyancy under the
+# Boussinesq approximation. Moreover, physical parameters are constant through space
+# apart from density. As a result, the system is fully defined by the compositional
+# Rayleigh number. We use the `material_field` function to define its value throughout
+# the domain (including the shape of the material interface transition) and provide it
+# to our approximation.
 
 # +
-Ra = 0  # Thermal Rayleigh number
+Ra = 0.0  # Thermal Rayleigh number
```
drhodrid, line 150: Will be nice when we can drop this!

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -158,101 +156,106 @@
...
+# initial pressure and velocity.
+stokes_solver = StokesSolver(
+    stokes,
+    T,
+    approximation,
+    bcs=stokes_bcs,
+    nullspace=stokes_nullspace,
+    transpose_nullspace=stokes_nullspace,
+)
+stokes_solver.solve()
```
drhodrid, line 210: Why is this needed outside of the time loop? Perhaps a comment to help a user. It's different to our other cases where the solves are all inside the time loop.

demos/multi_material/thermochemical_buoyancy/thermochemical_buoyancy.py
```diff
@@ -1,101 +1,102 @@
...
-# domain heated from below, a denser material sits at the bottom boundary beneath a
-# lighter material. Whilst the compositional stratification is stable, heat transfer
-# from the boundary generates positive buoyancy in the denser material, allowing thin
-# tendrils to be entrained in the convective circulation. To resolve these tendrils
-# using the level-set approach, significant mesh refinement is needed, making the
-# simulation computationally expensive. This tutorial will be updated once the
+# convection presented in [van Keken et al. (1997)](https://doi.org/10.1029/97JB01353).
+# Inside a 2-D domain heated from below, a denser material sits at the bottom boundary
+# beneath a lighter material. Whilst the compositional stratification is stable, heat
+# transfer from the boundary generates positive buoyancy in the denser material,
```
drhodrid, line 19:
```suggestion
# transfer from the boundary generates positive buoyancy,
```

demos/multi_material/thermochemical_buoyancy/thermochemical_buoyancy.py
```diff
@@ -1,101 +1,102 @@
...
+# allowing thin tendrils to be entrained in the convective circulation. To resolve these
```
drhodrid, line 20:
```suggestion
# allowing thin tendrils of denser material to be entrained by convective circulation. To resolve these
```

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -158,101 +156,106 @@
...
+
+# Instantiate a solver object for level-set advection and reinitialisation. G-ADOPT
+# provides default values for most arguments; we only provide those that do not have
+# one. No boundary conditions are required, as the numerical domain is closed.
```
drhodrid, line 214:
```suggestion
# one. No boundary conditions are required for the level set, as the numerical domain is closed.
```

demos/multi_material/thermochemical_buoyancy/thermochemical_buoyancy.py
```diff
@@ -1,101 +1,102 @@
...
+# As with all examples, the first step is to import the `gadopt` package, which also
 # provides access to Firedrake and associated functionality.
 
 from gadopt import *
 
-# For this problem, it is useful to define a mesh with non-uniform refinement. To this
-# end, we use the GMSH library to generate a mesh file in a format compatible with
-# Firedrake. We specifically increase vertical resolution at the top and bottom
-# boundaries of the domain.
+# For this problem, in the absence of adaptive mesh refinement, it is useful to define a
```
drhodrid, line 31:
```suggestion
# For this problem it is useful to define a
```

demos/multi_material/thermochemical_buoyancy/thermochemical_buoyancy.py
```diff
@@ -1,101 +1,102 @@
...
+# mesh with non-uniform spatial refinement. To this end, we use the GMSH library to
```
drhodrid, line 32: I think this is the first tutorial where we use gmsh -- so perhaps a little more background to it and some context to some of the mesh generation calls below would aid a new user.

demos/multi_material/thermochemical_buoyancy/thermochemical_buoyancy.py
```diff
@@ -1,101 +1,102 @@
...
-boundary = get_boundary_ids(mesh)
-
-V = VectorFunctionSpace(mesh, "CG", 2)  # Velocity function space (vector)
-W = FunctionSpace(mesh, "CG", 1)  # Pressure function space (scalar)
+V = VectorFunctionSpace(mesh, "Q", 2)  # Velocity function space (vector)
+W = FunctionSpace(mesh, "Q", 1)  # Pressure function space (scalar)
 Z = MixedFunctionSpace([V, W])  # Stokes function space (mixed)
-Q = FunctionSpace(mesh, "CG", 2)  # Temperature function space (scalar)
+# Temperature function space (scalar, discontinuous, equispaced)
+Q = FunctionSpace(mesh, "DQ", 2, variant="equispaced")
```
drhodrid, line 92: A comment motivating equispaced and other options would be useful - this will be new to the user.

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
```diff
@@ -158,101 +156,106 @@
...
+    stokes,
+    T,
+    approximation,
+    bcs=stokes_bcs,
+    nullspace=stokes_nullspace,
+    transpose_nullspace=stokes_nullspace,
+)
+stokes_solver.solve()
+
+# Instantiate a solver object for level-set advection and reinitialisation. G-ADOPT
```
drhodrid, line 212: I think given this is the first demo they see, where mention of reinitialisation is included ,you should provide some background to what exactly each of these options are and what imapct changing these values will have (i.e. if you change epsilon, bla bla bla... ).

tests/multi_material/run_benchmark.py
```diff
@@ -207,52 +200,50 @@ def write_output(output_file):
...
-    quad_degree=None,
-    solver_parameters=simulation.stokes_solver_params,
     nullspace=stokes_nullspace,
     transpose_nullspace=stokes_nullspace,
 )
-
-# Solve initial Stokes system
-stokes_solver.solve()
+if benchmark == "schmalholz_2011":
+    stokes_solver.solver_parameters["snes_linesearch_type"] = "cp"
```
drhodrid, line 234: Why is this needed? Add a comment so that if we return to this at a later time, we understand.

drhodrid (commented):
An initial review from me. Generally very nice. Lots of small comments. Remind me to discuss viscosity averaging with you at some stage. I'd appreciate Angus or Dale looking over the test specific changes, but I'm fairly happy with the demos etc... So provided they're happy with your changes to testing environment, I think this is good to go.

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
Patol75: And we have gone around full circle! Haha

Patol75: No, it does not. But `StokesSolver` still requires a `Function` for temperature as of today...

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
Patol75: (It has only been a year since I said that for the first time.)

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
Patol75: If we specify initial conditions for all variables, then I do not see a reason not to have an initial condition for velocity and pressure as well. That is what this solver call is achieving. That also implies that all our demos produce Stokes and energy solutions that are out of sync, right?

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
drhodrid: I'm happy with either of these because I don't think we have use DG or DQ in previous tutorials. Perhaps DQ?

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
drhodrid: I think it depends on the order of things in the other demos. In theory if the first solve inside the time loop is a stokes solve (as it is with case case), it is already getting an initial guess for u and p with the prescribed RHS, so I don't see a problem there. Can we not do the same here, rather than have this separate call?

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
drhodrid: But boundary conditions are required for other fields. Specifying this is important for clarity. It confused me the first time.

drhodrid (commented):
Aside from the last few comments, I think this is now good to go, but Angus or Dale will need a quick check too please on testing side of things etc... Well done!

demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
Patol75: I think of it this way: `StokesSolver` requires a temperature, so I need to define a temperature `Function`, so I need to define a temperature `FunctionSpace`. There is no reason for temperature and level-set function spaces to be the same.

Patol75:
Provided CI passes, is this PR good to go?

Patol75:
It has to be between 25 August and 8 September, given the commit and CI history.

drhodrid (approved):
Looks good to go. Well done.

PR #271, opened 2025-08-04
Initialise Firedrake solvers during G-ADOPT solver init

dsroberts:
This change adds a `SolverConfig` class that the base solver classes (in this case `StokesSolver` and `GenericTransportBase`) can subclass in order to manage their `solver_parameters` dicts. The key change is that the `solver_parameters` dict can now be fully configured (including deleting keys) with the `solver_parameters_extra` argument to `StokesSolver` and `GenericTransportBase`. This means that the `setup_solver()` calls in each of those can now be moved to the respective `__init__` methods, which means we no longer need a metaclass or to override `__new__` or any other ordering tricks to make sure that `solver_parameters` is set but modifiable before `.solve()` is called.

The changes in the demos and tests show what's needed from an API point-of-view, the `solver_parameters_extra` argument contains what would have been modified in `solver.solver_parameters`. I've added logging features and a print function to the `SolverConfig` class, I see this being used for debugging purposes, or to just have a readable record of the PETSc parameters in a job's output. I've also added some unit tests for the new functionality.

I'm aware this will need to be re-done when #225 is merged, I just wanted to have a quick-and-dirty proof of concept ready for discussion on Wednesday.


tests/analytical_comparisons/delta_cylindrical_freeslip.py
```diff
@@ -91,6 +91,7 @@ def model(level, nn, do_write=False):
     marker.interpolate(conditional(r < rp, 1, 0))
     n = FacetNormal(mesh)
     stokes_solver.F += g * cos(nn*phi) * dot(jump(marker, n), avg(v)) * dS_h
+    stokes_solver.setup_solver()
```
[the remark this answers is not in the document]
dsroberts: Once #225 is in this addition can go.

gadopt/solver_options_manager.py
```diff
@@ -0,0 +1,194 @@
...
+        If an element of a Mapping is another Mapping, recursively calls itself to
+        process that Mapping. If an element of the mapping is `DeleteParam`, remove
+        it from the dict if it exists in the Mapping
+        """
+        outmap = dict(inmap)
+        for k, v in delta_map.items():
+            if v is DeleteParam:
+                if k in outmap:
+                    debug_print(self._top_level_class_name, f"Deleting {key_prefix}[{k}]")
+                    del outmap[k]
```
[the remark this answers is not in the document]
dsroberts: If a user deletes a key that was never set, isn't the outcome what they wanted? That is a good thing to add at the info or debug log level though.

gadopt/solver_options_manager.py
```diff
@@ -0,0 +1,194 @@
...
+        This method generates a logging prefix based on the class hierarchy, registers an `update_callback`
+        if provided and sets the solver configuration based on the `default_config` and `extra_config`
+        provided. This structure allows a subclass to determine its default solver settings, and a user
+        to override settings as necessary. The `update_callback` allows a subclass to specify a function that
+        must be called when the solver settings are changed such that the firedrake objects that depend on
+        these settings are reinitialised whenever the solver settings are changed.
+
+        Any Mapping type can be passed into this function, and the function will take care of copying
+        the mapping to a mutable dictionary. The `extra_config` argument is optional and reflects the
+        case when the user does not wish to modify the default solver settings provided by a subclass.
```
w3168, line 88: How about?

```suggestion
        case when the user wishes to modify the default solver settings provided by a subclass.
```

gadopt/solver_options_manager.py
```diff
@@ -0,0 +1,194 @@
...
+    parameters on object initialisation, allowing the firedrake
+    `[Non]LinearVariationalSolver solver object to be fully initialised during
+    construction of a G-ADOPT solver object.
+    """
+
+    def init_solver_config(
+        self,
+        default_config: DefaultConfigType,
+        extra_config: ExtraConfigType | None = None,
+        callback: Callable[[], None] | None = None,
```
w3168, line 75: should this be called ```update_callback``` to tie in with the doc string?

w3168 (commented):
Thanks @dsroberts I think this looks good! 

From a user perspective I think it is clearer and the debugging interface is probably very helpful to know what options are being set. 

It took me a little while to get my head around the different methods in the SolverOptions class but I think it all makes sense now! 

I have some very minor suggestions along with Thomas' ones from before. 

Also as you say this will be rebased off of #225

gadopt/solver_options_manager.py
```diff
@@ -0,0 +1,194 @@
...
+import weakref
+
+from .utility import (
+    DEBUG,
+    WARNING,
+    log_level,
+)
+
+
+def debug_print(class_name: str, string: str):
```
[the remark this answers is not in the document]
dsroberts: Didn't know that was there, I'll use that instead of `print`

gadopt/solver_options_manager.py
dsroberts: Hmmm... I guess that is needlessly obtuse. `callback` is the function argument which becomes the `update_callback` class variable. They're different things, `callback` being the actual function and `update_callback` is a weak reference to that function. Maybe `update_callback` should be renamed `callback_ref` or something to make it a bit clearer.

Patol75:
When a class inherits from multiple classes, objects (i.e. attributes, methods, etc) are resolved according to [an order Python defines](https://docs.python.org/3/tutorial/classes.html#multiple-inheritance). Essentially, we avoid any potential problem if the intersection of all parent classes is null, which should be the case here.

dsroberts:
My idea was that as the toolkit grows, the solver classes would start inheriting from more different classes. Its kind of modelled after Firedrake's `NonlinearVariationalSolver`, which has 2 base classes, `NonlinearVarationalSolverMixin` which provides adjoint related stuff, and `OptionsManager`, which handles turning Firedrake solver options into petsc options. This is why I didn't give `SolverOptions` an `__init__` method, because as the toolkit grows, there will be more base classes, and so `super.__init__()` becomes more ambiguous. We're pretty much using the inheritance to add a bunch of common methods to the different solver classes. I felt this fit well with how the solvers are initialised currently, both the Stokes and Transport solvers end with a bunch of `set_*` function calls, now they both get another one from the `SolverOptions` class.

That being said, its a bit unconventional, and as the toolkit expands it will become more cumbersome to figure out which bits of a solver class come from where. This could well be re-implemented as a `config` object owned and initialised by each solver class with little effort. The downside of that is that a consumer of the solver classes loses access to the convenience routines like `update_solver_config` and `print_solver_config`, and would instead have to find them through accessing the aforementioned `config` object (and therefore requiring a consumer to have knowledge of and uncontrolled interaction with an object that doesn't belong to it, like the current `solver_parameters` dict), or a reimplementation within the solver class itself that just calls `self.config.print_solver_config()`. I'm not wedded to the current structure, and can be convinced either way. Ideally the outcome should result in less effort to develop future solvers, and since pretty much all of us do that, everyone should have a say in this.

w3168:
Thanks @dsroberts  and @Patol75 for clarifying!

gadopt/solver_options_manager.py
```diff
@@ -0,0 +1,194 @@
...
+
+        The function provided to `update_callback` must take no arguments and return
+        nothing. When a subclass provides this function, a user does not need to
+        be aware of the underlying Problem/Solver objects in order to ensure that
+        a configuration update during an in-progress simulation takes effect properly.
+        When provided in the `init_solver_config` call, `update_callback` will run
+        when `solver_parameters` is ready, therefore `init_solver_config` can be
+        the last call directly in a Solver's `__init__` method.
+        """
+        self.update_callback = weakref.WeakMethod(callback)
```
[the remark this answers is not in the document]
dsroberts: Just a note on this one, I have this change ready to go, but it isn't worth running the CI again for. I'll incorporate it with the next round of review changes.

w3168 (dismissed):
Thanks @dsroberts  looks good to me! 

Only a small comment on an existing comment in the 3d cartesian demo....

I've ticked approve but happy to wait for others to also give their approval.

dsroberts:
> I don't think its harmful
Turns out it was harmful. In our Stokes solver configuration tests, we're importing `iterative_stokes_solver_parameters` from `stokes_integrators`, which is being modified during stokes solver setup. So you could put any garbage in the `fieldsplit_0` and `fieldsplit_1` dicts and the test will pass because its comparing literally the same objects.

gadopt/solver_options_manager.py
```diff
@@ -0,0 +1,194 @@
...
+
+        If the `callback_ref` attribute is set, it will be called on completion of the update.
+        """
+        # "Initialise" solver config attrs on first call
+        if not hasattr(self, "_top_level_class_name"):
+            self._top_level_class_name = self.__class__.__mro__[0].__name__
+        if not hasattr(self, "solver_parameters"):
+            self.solver_parameters = {}
+        if not hasattr(self, "callback_ref"):
+            self.callback_ref = None
```
dsroberts, line 175: I know this is awkward, a fix will go in along with any other changes from the next round of reviews

dsroberts:
Thanks @Patol75. I'll update my tooling for an 88 character line width. Is that written anywhere?

tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py
```diff
@@ -202,15 +202,17 @@ def gaussian(center, c, a):
 
 energy_solver = EnergySolver(T, u, approximation, delta_t, ImplicitMidpoint, bcs=temp_bcs)
 
+solver_params_extra = {"fieldsplit_0": {"ksp_converged_reason": None}, "fieldsplit_1": {"ksp_converged_reason": None}}
```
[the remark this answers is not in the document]
dsroberts: Good question. On the one hand, at this stage it is redundant, as it was in the original demo. For this change, I didn't give any thought to what the settings should be, I just copied what was there. On the other hand, if its a parameter we know we want in a particular simulation, should we leave it there regardless of what a user decides to do with the underlying G-ADOPT log level?

dsroberts:
@Patol75 I've updated the docstrings and comments based on your review. I'm not sure what you had in mind for the notes in #225, so I've just left it as-is except for replacing `stokes_solver_update` with `stokes_solver_extra`.

gadopt/solver_options_manager.py
```diff
@@ -0,0 +1,196 @@
...
+    ```
+    stokes_solver = StokesSolver( ...
+        solver_parameters_extra = { 'ksp_monitor' : DeleteParam }
+    )
+    ```
+    The resulting `stokes_solver.solver_parameters` will not contain the
+    `ksp_monitor` key.
+    """
+
+    pass
```
Patol75, line 48: Obviously, up to you, but in case you did not know it and find it interesting, this `pass` statement is not necessary.

gadopt/solver_options_manager.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+# Type alias
+ConfigType = Mapping[
+    str, str | float | int | None | type[DeleteParam] | Mapping[str, str | float | int | None | type[DeleteParam]]
+]
```
Patol75, line 54: There is probably no point writing it like below, right?

```python
DictType = Mapping[str, str | float | int | None | type[DeleteParam]]
ConfigType = Mapping[str, str | float | int | None | type[DeleteParam] | DictType]
```

gadopt/solver_options_manager.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class SolverOptions:
+    """Manage PETSc solver options in G-ADOPT.
+
+    This class is designed to be subclassed by the base class for any solvers
+    included in G-ADOPT. It provides methods for handling and modifying solver
+    parameters passed to Firedrake's `[Non]LinearVariationalSolver solver object
```
Patol75, line 62: `[Non]LinearVariationalSolver` (i.e. you are missing a `)

Patol75 (approved):
Great work! Let's get this in and then deal with #268.

tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py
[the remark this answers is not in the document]
Patol75: Let's leave it as-is; the logging update will take care of it in due course.

gadopt/stokes_integrators.py
```diff
@@ -444,49 +426,57 @@ def set_form(self) -> None:
...
             match solver_preset:
                 case "direct":
-                    self.solver_parameters |= direct_stokes_solver_parameters
+                    self.add_to_solver_config(direct_stokes_solver_parameters)
                 case "iterative":
-                    self.solver_parameters |= iterative_stokes_solver_parameters
+                    self.add_to_solver_config(iterative_stokes_solver_parameters)
                 case _:
                     raise ValueError("Solver type must be 'direct' or 'iterative'.")
         elif self.mesh.topological_dimension() == 2 and is_cartesian(self.mesh):
```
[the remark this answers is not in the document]
dsroberts: I think that's out of scope for this PR.

gadopt/solver_options_manager.py
dsroberts: I did not know that. I think I'll keep the `pass` there though, as omitting it only allowed because of the docstring.

gadopt/solver_options_manager.py
dsroberts: Strictly speaking, neither type alias is correct as the `process_mapping` function can handle mappings of an arbitrary depth. Turns out python supports these recursive type hints by way of forward declaration. I've taken this opportunity to crank the pedantry up to 11 and redefine the exported `ConfigType` to be recursive as well as add 2 new distinct internal types that are essentially the same as `ConfigType` except they are not allowed to contain `DeleteParam` and one of them is mutable.

dsroberts:
Had a quick chat with @sghelichkhani about the repeated docstrings and convinced him they should stay. There is no guarantee that someone looking at the docstring for `direct_stokes_solver_parameters` will also look at the docstring for `iterative_solver_parameters`, so advice for modifying them should be present for all of them.

gadopt/stokes_integrators.py
```diff
@@ -814,19 +800,19 @@ def __init__(
         displacement: fd.Function,
         **kwargs,
     ) -> None:
-        super().__init__(solution, approximation, **kwargs)
+        if "dt" not in kwargs:
+            raise TypeError(f"A timestep (the 'dt' keyword argument) must be provided "
```
w3168, line 804: I think this is definitely better than we had, but should we make dt a required argument of the Viscoelastic solver?

Patol75:
If everyone else is happy, I think this is good to go! Well done!

stephankramer (commented):
I think this is really nice. Much better to have everything initialized in __init__ indeed instead of having to worry about the state of the Solver but deals nicely with the need to fiddle with default solver options. Some minor suggestions

Patol75:
Sorry about the conflicts; they are from #291. They should be pretty simple to deal with.

Patol75:
All looking wonderful! If no one else has suggestions or comments, then let's try to merge within the next 24 hours.

stephankramer (approved):
Ship it! :rocket:

PR #272, opened 2025-08-06
Unpin VTK from <9.5.0

angus-g:
pyvista has had its 0.46.0 release, which should bring compatibility with this VTK version.


drhodrid (approved):
Outstanding!

PR #279, opened 2025-08-18
Fixing non-zero normal pygplates velocities

sghelichkhani:
This is addressing the issues outline in https://github.com/g-adopt/g-adopt/discussions/278.

### DIFF
--- a/gadopt/momentum_equation.py
+++ b/gadopt/momentum_equation.py
@@ -85,12 +85,11 @@ def viscosity_term(
             )
 
         if "u" in bc:
-            trial_tensor_jump = outer(eq.n, trial - bc["u"])
+            trial_tensor_jump = 2 * sym(outer(eq.n, trial - bc["u"]))
             if compressible_stress:
                 trial_tensor_jump -= (
                     2 / 3 * identity * (dot(eq.n, trial) - dot(eq.n, bc["u"]))
                 )
-            trial_tensor_jump += transpose(trial_tensor_jump)
             # Terms below are similar to the above terms for the DG dS integrals.
             F += (
                 2
@@ -102,10 +101,9 @@ def viscosity_term(
             F -= inner(outer(eq.n, eq.test), stress) * eq.ds(bc_id)
 
         if "un" in bc:
-            trial_tensor_jump = outer(eq.n, eq.n) * (dot(eq.n, trial) - bc["un"])
+            trial_tensor_jump = 2 * outer(eq.n, eq.n) * (dot(eq.n, trial) - bc["un"])
             if compressible_stress:
                 trial_tensor_jump -= 2 / 3 * identity * (dot(eq.n, trial) - bc["un"])
-            trial_tensor_jump += transpose(trial_tensor_jump)
             # Terms below are similar to the above terms for the DG dS integrals.
             F += (
                 2
--- a/tests/unit/test_symmetry.py
+++ b/tests/unit/test_symmetry.py
@@ -1,0 +1,98 @@
+import firedrake as fd
+import gadopt
+import pytest
+
+N = 4  # resolution in all directions
+mesh1d = fd.UnitIntervalMesh(N)
+mesh1dcircle = fd.CircleManifoldMesh(N)
+mesh2dtri = fd.UnitSquareMesh(N, N, quadrilateral=False)
+mesh2dquad = fd.UnitSquareMesh(N, N, quadrilateral=True)
+mesh2dcs = fd.UnitCubedSphereMesh()
+mesh2dico = fd.IcosahedralSphereMesh(1)
+meshes = {
+    "2D-tri": mesh2dtri,
+    "2D-quad": mesh2dquad,
+    "2D-extruded": fd.ExtrudedMesh(mesh1d, N),
+    "3D-tet": fd.UnitCubeMesh(N, N, N, hexahedral=False),
+    "3D-hex": fd.UnitCubeMesh(N, N, N, hexahedral=True),
+    "3D-extruded": fd.ExtrudedMesh(mesh2dquad, N),
+    "3D-extruded-prism": fd.ExtrudedMesh(mesh2dtri, N),
+    "2D-cylinder": fd.ExtrudedMesh(mesh1dcircle, N),
+    "3D-cubed-sphere": fd.ExtrudedMesh(mesh2dcs, N),
+    "3D-icosahedral-sphere": fd.ExtrudedMesh(mesh2dico, N)
+}
+
+
+@pytest.fixture(scope="module", params=meshes.items(), ids=meshes.keys())
+def mesh(request):
+    id, mesh = request.param
+    mesh.cartesian = not any(x in id for x in ['cylinder', 'sphere'])
+    return mesh
+
+
+@pytest.fixture(scope="module",
+                params=[gadopt.BoussinesqApproximation,
+                        gadopt.ExtendedBoussinesqApproximation,
+                        gadopt.TruncatedAnelasticLiquidApproximation,
+                        gadopt.AnelasticLiquidApproximation])
+def approximation(request):
+    Ra = 1
+    Di = 1
+    if request.param is gadopt.BoussinesqApproximation:
+        return request.param(Ra)
+    else:
+        return request.param(Ra, Di)
+
+
+@pytest.fixture(scope="module", params=["TaylorHood",])
+def solution_space(request, mesh):
+    # at the moment only P2-P1 is supported
+    # would like to test discontinuous velocity, but
+    # that requires the pressure gradient term to handle that
+    match request.param:
+        case "TaylorHood":
+            V = fd.VectorFunctionSpace(mesh, "CG", 2)
+            W = fd.FunctionSpace(mesh, "CG", 1)
+            return V * W
+        case _:
+            raise ValueError("Unknown discretisation type")
+
+
+def test_stokes_symmetry(approximation, mesh, solution_space):
+    """Test symmetry of discretised Stokes matrix where expected
+
+    In particular, tests symmetry of weak bc terms."""
+    z = fd.Function(solution_space)
+    u, p = z.subfunctions
+    # use a velocity that's not divergence free, to test symmetry of div(u) terms:
+    X = fd.SpatialCoordinate(mesh)
+    u.interpolate(X)
+
+    T = fd.Function(solution_space.sub(1))
+    boundary = gadopt.get_boundary_ids(mesh)
+    bids = list(vars(boundary).values())
+    bcs = {bids[0]: {'un': 0}, bids[1]: {'normal_stress': 0}}
+    # cylindrical/spherical meshes only have 2 boundaries
+    # if we have more, let's test some more bc types
+    if len(bids) > 2:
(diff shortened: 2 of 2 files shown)


gadopt/gplates/gplates.py
```diff
@@ -86,6 +91,16 @@ class GplatesVelocityFunction(GPlatesFunctionalityMixin, fd.Function):
         ...                                    name="GplateVelocity")
         >>> gplates_function.update_plate_reconstruction(ndtime=0.0)
     """
+    # Solver parameters for tangential projection, since this spherical mesh, use iterative solver
+    tangential_project_solver_parameters = {
+        "ksp_type": "cg",
+        "pc_type": "bjacobi",
+        "ksp_rtol": 1e-9,
+        "ksp_atol": 1e-12,
+        "ksp_max_it": 20,
```
drhodrid, line 100: 20 feels low. Perhaps make 100 to provide more of a chance to hit the tolerances?

drhodrid (commented):
Very small comment from me. Thanks for this fix. I think some of the steps could do with more verbose commenting, such that we know why they are done if we ever revisit this code. On top of that, ideally @stephankramer would review this, given his previous experience with the normals/radial components in Fluidity.

gadopt/gplates/gplates.py
sghelichkhani: I chose 20 based on a discussion on FD slack: https://firedrakeproject.slack.com/archives/C1Q0Y6H8A/p1755514261165649 

The problem will be essentially the same always (inverting for the mass matrix). Currently it converges in 5 iterations. So setting it to 20 is essentially a sanity check on that solve. Happy for @stephankramer to comment on this if he is going to review this.

gadopt/gplates/gplates.py
```diff
@@ -41,8 +42,12 @@ def update_plate_reconstruction(self, ndtime):
                 self.boundary_coords, ndtime)
         )
 
-        # At this point the values are updated.
-        # So we have to make sure it is shown correctly on tape if we are annotating
+        # Project tangential with appropriate annotation handling
+        suitable_context = stop_annotating() if annotate_tape() else nullcontext()
+        with suitable_context:
```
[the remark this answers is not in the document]
sghelichkhani: Fair point!

gadopt/gplates/gplates.py
sghelichkhani: @stephankramer  It is more or less the same concern. A bare project (without specifying iterative solve parameters) usually ends up being too expensive and very slow for the spherical high resolution problem.

sghelichkhani: Hah! I always assumed the defaults for `project` are mumps. But now I see what you meant @stephankramer. So I changed the project parameters to be default to use the same. I am now only setting verbosity level.

sghelichkhani:
@stephankramer I have now changed all solve parameters to default, only setting output level. Thanks.

stephankramer:
Sorry with "I presume the failure is just a tolerance issue?" I meant the failing test: you're not passing the CI. But yes, what you wrote make sense to me - on the nodes you end up with some kind of tangential direction that is an average over the adjacent surface facets, which indeed won't be purely radial due to asymmetries - and I like the way you do the projection. One thing I am wondering, looking at this again, is whether it should be using `FacetNormal(mesh)` instead of `upward_normal(mesh)` both in the projection and the test. It's the facet normal, i.e. the actual normal to the discrete approximation of the sphere that determines the direction of the discretised normal stress, and what the flux through the boundary is for conservation properties. If you're using a degree=2 curved mesh with sufficient resolution, the difference between the two should be small though...
Thanks for the link to the discussion, hadn't seen that yet - there I had another thought on the interpolation more generally. Happy with the PR though assuming the tests pass.

sghelichkhani:
Ah! Forgot about the other reference values for the gplates case. Will fix that shortly

gadopt/gplates/gplates.py
```diff
@@ -113,6 +147,91 @@ def __init__(self, function_space, *args, gplates_connector=None, top_boundary_m
...
+                "GplatesVelocityFunction: Velocity field is discontinuous."
+                "Removing the radial component of discontinuous velocity fields is not supported."
+            )
+        else:
+            V = fd.FunctionSpace(self.function_space().mesh(), self.ufl_element())
+
+        self.tangential_velocity = fd.Function(V, name="tangential_velocity")
+
+        # Normal vector (radial direction for spherical geometry)
+        r = fd.FacetNormal(self.function_space().mesh())
```
Patol75, line 171: `r = fd.FacetNormal(V.mesh())`

gadopt/gplates/gplates.py
```diff
@@ -113,6 +147,91 @@ def __init__(self, function_space, *args, gplates_connector=None, top_boundary_m
...
+
+    def remove_radial_component(self):
+        """
+        Project velocity to tangent plane by removing radial component using linear variational solver.
+
+        This method uses a pre-built linear variational solver for efficiency, avoiding the overhead
+        of rebuilding the projection operator each time.
+        """
+        # Use the a linear variational solver for removing the radial component
+        if not hasattr(self, '_solver_is_set_up') or not self._solver_is_set_up:
```
Patol75, line 224: Why not defining `self._solver_is_set_up = False` in `__init__`?

gadopt/gplates/gplates.py
```diff
@@ -127,6 +246,9 @@ class pyGplatesConnector(object):
     # minimum distance, bellow which we do not interpolate
     #   this is just to avoid division by zero when weighted averaging
     epsilon_distance = 1e-8
+    # minimum magnitude, bellow which we do not scale the velocities
```
Patol75, line 249: `below`

gadopt/gplates/gplates.py
```diff
@@ -127,6 +246,9 @@ class pyGplatesConnector(object):
...
+    # This is to avoid division by zero when scaling the velocities
+    eps_rel = 1e-10
```
Patol75, line 251: Is `1e-10` a velocity in m/s? If yes, it does not feel that small. But perhaps it is because we are interested in surface plate velocities.

gadopt/gplates/gplates.py
```diff
@@ -384,6 +506,26 @@ def _interpolate_seeds_u(self, target_coords):
...
+        # res_u_tangential = res_u - (res_u · e_r) * e_r
+        res_u_tangential = res_u - radial_velocity_res[:, np.newaxis] * target_coords
+
+        # Calculate new magnitudes after projection
+        tangential_magnitudes = np.linalg.norm(res_u_tangential, axis=1)
+
+        # Rescale to preserve original magnitude (avoid division by zero)
+        scale_factor = np.divide(original_magnitudes, tangential_magnitudes,
+                                 out=np.ones_like(tangential_magnitudes),
+                                 where=tangential_magnitudes > pyGplatesConnector.eps_rel)
```
Patol75, line 525: It should work as well without providing the `out` argument.

gadopt/gplates/gplates.py
```diff
@@ -113,6 +147,91 @@ def __init__(self, function_space, *args, gplates_connector=None, top_boundary_m
...
+        # Setting up boundary condition, problem and solver
+        # The field is only meaningful on the boundary, so set zero everywhere else
+        self.interior_null_bc = InteriorBC(V, 0., [self.top_boundary_marker])
+
+        self.problem = fd.LinearVariationalProblem(a, L, self.tangential_velocity,
+                                                   bcs=self.interior_null_bc,
+                                                   constant_jacobian=True)
+        self.solver = fd.LinearVariationalSolver(
+            self.problem,
+            solver_parameters=self.kwargs.get("solver_parameters", self.tangential_project_solver_parameters),
```
Patol75, line 211: If you go ahead with my suggestion about the `ds` keyword arguments, then you could also have `solver_parameters` as a keyword argument.

dsroberts:
@Patol75 yep. For consistency's sake we should integrate `SolverOptions` into this. Happy to take that on after both this and #271 have been merged.

gadopt/gplates/gplates.py
```diff
@@ -72,6 +77,8 @@ class GplatesVelocityFunction(GPlatesFunctionalityMixin, fd.Function):
         val (optional): Initial values for the function. Defaults to None.
         name (str, optional): Name for the function. Defaults to None.
         dtype (data type, optional): Data type for the function. Defaults to ScalarType.
+        kwargs (dict, optional): Additional keyword arguments, inlcuding `solver_parameters`
+            (dict, optional), and `ds_kwargs` (dict, optional) for the ds measure.
```
angus-g: Aren't you updating `ds_kwargs` directly with the kwargs? I don't think this matches the code...

gadopt/gplates/gplates.py
```diff
@@ -113,6 +147,91 @@ def __init__(self, function_space, *args, gplates_connector=None, top_boundary_m
...
+        if not hasattr(self, '_solver_is_set_up') or not self._solver_is_set_up:
+            self.setup_solver()
+
+        # Project the velocity to the tangential plane
+        self.solver.solve()
+
+        # Copy the tangential velocity back to self
+        self.dat.data_with_halos[self.dbc.nodes, :] = (
+            self.tangential_velocity.dat.data_with_halos[self.dbc.nodes, :]
+        )
```
angus-g: Does this a) need halos; b) need to be `self.dat.data_wo[...] = self.tangential_velocity.dat.data_ro[...]`?

gadopt/gplates/gplates.py
```diff
@@ -113,6 +147,91 @@ def __init__(self, function_space, *args, gplates_connector=None, top_boundary_m
...
+        v = fd.TrialFunction(V)
+
+        # Define the tangential projection: v_tangential = v - (v·r)r
+        tangential_expr = self - fd.inner(self, r) * r
+
+        # Getting ds measure: this can be set by the user through kwargs
+        # We are only looking at the surface measure at the top boundary
+        ds_kwargs = {
+            "domain": self.function_space().mesh(),
+            "degree": 2 * V.ufl_element().degree()[0] + 1
```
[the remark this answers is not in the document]
sghelichkhani: It's a good question! For the stokes problem (mantle convection where we impose surface velocities) that is a pretty safe assumption.

gadopt/gplates/gplates.py
sghelichkhani: The velocities are non-dimensionalised, they're not in m/s anymore. This is also more about numerical stability. This is just small enough to catch genuinely problematic divisions. Large enough to avoid issues with floating-point precision.

angus-g:
I think as discussed this is a bugfix on top of `release` so I'll change the PR base to that.

gadopt/gplates/gplates.py
sghelichkhani: a) Since we are working on parallel, I thought it's safest to use halos. Unless it's guaranteed for halos to exchange after this?
b) I have now changed it to reflect your point about the proper accessors. `data_wo_with_halos` for writing and `data_ro_with_halos` for reading.

sghelichkhani: Discussed with Angus and he is happy.

PR #281, opened 2025-08-20
Run longtests periodically

dsroberts:
Run long tests on all HPC systems at 1905 Sunday UTC (5:05am AEST, 6:05am AEDT every Monday). Scheduled downtimes on Gadi and Setonix are on Tuesdays, so these jobs should not be affected by scheduled downtimes.

### DIFF
--- a/.github/workflows/longtest.yml
+++ b/.github/workflows/longtest.yml
@@ -11,16 +11,20 @@ on:
           - all
           - gadi
           - setonix
+  schedule:
+    - cron: "5 19 * * 0"
 
 jobs:
   setup:
     runs-on: ubuntu-latest
     outputs:
       system_matrix: ${{ steps.build_matrix.outputs.system_matrix }}
+    env:
+      SYSTEM: ${{ inputs.system || 'all' }}
     steps:
     - id: build_matrix
       run: |
-        if [[ "${{ inputs.system }}" == "all" ]]; then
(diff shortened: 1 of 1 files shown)


angus-g:
Makes sense. Maybe we could (separately) add some more visibility to the reporting of this so it doesn't get lost, like a zulip integration or big red light

dsroberts:
Definitely! I have a spare rpi zero at home, I can bring that in, set up a little tcp listener on it and connect it to some LEDs... Zulip integration would be good though, for this and for any warnings from the memory profiling too.

.github/workflows/longtest.yml
```diff
@@ -11,16 +11,20 @@ on:
           - all
           - gadi
           - setonix
+  schedule:
+    - cron: "5 19 * * 0"
```
sghelichkhani, line 15: Did we not say every second Sunday? Or did decide to go every Sunday?
```suggestion
	- cron: "5 19 1-7,15-21  * 0"
```

angus-g: Unfortunately I don't think we can do every other Sunday very easily... We could run a timer on a local machine to trigger a workflow dispatch. As it is, I think this syntax runs it on days 1-7, 15-21 *and* on Sundays. Shame cron syntax is so limited.

PR #282, opened 2025-08-21
Load optimisation checkpoints relative to checkpoint directory

angus-g:
We should be able to move optimisation checkpoints from their original location. I've modified the optimisation checkpointing test case to exercise this too.

### DIFF
--- a/gadopt/inverse.py
+++ b/gadopt/inverse.py
@@ -159,14 +159,7 @@ class LinMoreOptimiser:
         self._add_statustest()
 
         self.rol_solver.rolvector.checkpoint_path = self.checkpoint_dir / "solution_checkpoint.h5"
-        self.rol_solver.rolvector.load(self._mesh)
-
-        # ROL algorithms run in a loop like `while (statusTest()) { ... }`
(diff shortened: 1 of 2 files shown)


gadopt/inverse.py
```diff
@@ -271,15 +271,15 @@ def save(self, checkpoint_path):
...
+    def load(self, mesh, checkpoint_path):
         """Loads the checkpointed data for this vector from disk.
 
         Called by the parent Optimiser after the ROL state has
         been deserialised. The pickling routine will register
         this vector within the registry.
         """
 
-        with CheckpointFile(str(self.checkpoint_path), "r") as f:
+        with CheckpointFile(str(checkpoint_path / self.checkpoint_path.name), "r") as f:
```
sghelichkhani, line 282: I am imagining my self future to be very confused about why we have `checkpoint_path` and `self.checkpoint_path.name` here. Does it make sense to add a bit of background to this in the docstring?

sghelichkhani (approved):
Looking good. Maybe just providing a bit of background in the docsrting of why we load the path the way we do might be useful. If not merge as it is.

PR #283, opened 2025-08-25
Increase tolerance on level 5 parallel scaling test

dsroberts:
The level 5 parallel scaling tests run on less than a full node, so can be scheduled across a single CPU, or across multiple CPUs. This leads to the level 5 tests having a much more variable runtime than the level 6 or 7 tests.

### DIFF
--- a/tests/parallel_scaling/test_scaling.py
+++ b/tests/parallel_scaling/test_scaling.py
@@ -39,6 +39,10 @@ def test_scaling_iterations(level, component):
 def test_scaling_pc_setup_time(level):
     assert not isinstance(ghpc_system, str), "Attempted to run longtest without gadopt_hpc_helper module"
 
+    # The level 5 parallel scaling tests run on less than a full node, so can
+    # be scheduled across a single CPU, or across multiple CPUs. This leads to
+    # the level 5 tests having a much more variable runtime than the level 6 or
+    # 7 tests. Therefore the level 5 tests have a higher timing tolerance.
     tol = 0.2 if level == 5 else 0.1
 
     b = Path(__file__).parent.resolve()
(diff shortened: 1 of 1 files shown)


tests/parallel_scaling/test_scaling.py
```diff
@@ -38,24 +38,29 @@ def test_scaling_iterations(level, component):
 @pytest.mark.parametrize("level", levels)
 def test_scaling_pc_setup_time(level):
     assert not isinstance(ghpc_system, str), "Attempted to run longtest without gadopt_hpc_helper module"
+
+    tol = 0.2 if level == 5 else 0.1
```
sghelichkhani, line 42: Since we might come back to this years later and wonder why, do you think it makes sense to add your description of this PR as a comment here? 

I know we can trace back the change through git, just feel much more exposed that way

tests/parallel_scaling/test_scaling.py
```diff
@@ -38,24 +38,29 @@ def test_scaling_iterations(level, component):
...
-    assert abs((expected - stokes_pc_setup) / expected) < 0.1
+    assert abs((expected - stokes_pc_setup) / expected) < tol
 
 
 @pytest.mark.longtest
 @pytest.mark.parametrize("level", levels)
 def test_scaling_total_solve_time(level):
     assert not isinstance(ghpc_system, str), "Attempted to run longtest without gadopt_hpc_helper module"
 
+    tol = 0.2 if level == 5 else 0.1
```
sghelichkhani, line 58: Same here.

sghelichkhani (approved):
Looking good! Pending the one suggestion can go in.

PR #287, opened 2025-09-02
adapt for MeshSequence

ksagiyam:
Required for https://github.com/firedrakeproject/firedrake/pull/3478, but can be merged independently.


ksagiyam:
@sghelichkhani Is the CI failure my fault, or is it a common issue?

gadopt/utility.py
```diff
@@ -92,8 +92,17 @@ def update_timestep(self):
         return float(self.dt_const)
 
 
+def is_cartesian(mesh):
+    try:  # firedrake main
+        from firedrake import MeshSequenceGeometry  # noqa: F401
+
+        return mesh.unique().cartesian
+    except ImportError:  # firedrake release
+        return mesh.cartesian
```
dsroberts, line 101: As per overall comment, could you please add the following to L25:
```
try:
    from firedrake import MeshSequenceGeometry
except ImportError:
    MeshSequenceGeometry = None
```

change `is_cartesian(mesh)` to this
```suggestion
def is_cartesian(mesh):
    if MeshSequenceGeometry is not None and isinstance(mesh, MeshSequenceGeometry):
        return mesh.unique().cartesian
    else:
        return mesh.cartesian
```
This fits more with our preferred coding style, and gives us a little extra sanity checking in case we ever want the `is_cartesian` function to work on other objects that have a `.cartesian` property.

dsroberts (changes requested):
Thanks for your contribution @ksagiyam. We've been working around this on your submesh_core branch for a while now, so its good to have an 'official' fix. That being said, what you have doesn't quite fit with our (unwritten, unofficial) style preferences, specifically, the import in a place other than the top of a file. All we need is the try/except block moved to the top with the rest of the imports and `is_cartesian` modified accordingly.

PR #289, opened 2025-09-04
Download plate reconstructions with curl

dsroberts:
This commit should allow PRs from forks to run in our CI by removing the need to access any repository secrets.


sghelichkhani (approved):
Thanks @dsroberts for this. I am not sure why we decided to go with s3config at the time (maybe not to expose our data repository??). But I think at this point in time we are assuming that data.gadopt.org is pretty much exposed to all users. So this should be fine.

angus-g (approved):
Perfect, thanks!

PR #291, opened 2025-09-17
Make direct solver parameters the default for 2-D simulations

### LINKED ISSUE #125: Inconsistency on transition from direct --> iterative solvers
For 2-D problems, the Stokes solver defaults transition from direct to iterative when the geometry is cylindrical (they are iterative by default in 3-D). The energy solver, however, does not make the same transition, and only does so in 3-D. For consistency, they should be the same IMO.

Patol75:
To have a consistent behaviour between transport and Stokes solvers, this PR enforces direct solver parameters for all 2-D simulations. Existing demos and tests that relied on an iterative solver now explicitly specify this need.

Closes #125

### DIFF
--- a/demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
+++ b/demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
@@ -447,6 +447,7 @@ stokes_solver = ViscoelasticStokesSolver(
     displacement,
     dt=dt,
     bcs=stokes_bcs,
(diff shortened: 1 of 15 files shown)


Patol75:
Is there a reason to make the 2-D cylindrical cases use an iterative solver? I tested locally `2d_cylindrical` and `2d_cylindrical_TALA_DG` with the default direct solver. The simulations ran ~20x faster than the default iterative solver (my RAM might have bottlenecked, but still), and `pytest` passed. These tests take a long time in the CI; we could use that speed-up.

PR #292, opened 2025-09-19
Fix mermaid links

dsroberts:
Mermaid diagrams have been moved one directory up in the page heirachy, links must be changed accordingly.


sghelichkhani (approved):
Approved! Thanks Dale.

Issue #293, opened 2025-09-22
Proposal: development and release branching strategy


angus-g:
Given that we have a smoke test in Firedrake `master`, we need to make sure that at least some subset of G-ADOPT is compatible with that API. However, we also (typically) tend to develop against the latest release: our test suite is run with the `firderake-vanilla-default:latest` container, which is the latest release.

I think we need a parallel branch strategy, where one branch targets, and is tested against `master` (using the `firedrake-vanilla-default:dev-main` container), and the other maintains compatibility with the latest release. The question then is which is our default development branch?

1. follow a "typical" development strategy where we target `master`, except for releases or bugfixes. Synchronisation is done when we do G-ADOPT releases (which probably need a less ad-hoc strategy in themselves)
2. continue targeting the latest Firedrake release, maintaining only API compatibility updates in a `dev-master` branch. Synchronisation happens on Firedrake releases

stephankramer:
Yes I agree with all that. I think either of the two options above has some drawbacks or benefits, but I'm leaning to option 1:

Some reasoning:
* Some development work needs to happen using firedrake main, as it depends on new functionality (or updates to it) in firedrake. Always hard to predict, but I think things like: particles, submesh, petsc changes etc. We don't want to have a situation where we have two branches that admit development PRs (as it complicates integration) so option 2 would mean that we only admit PRs in our main branch which means that development that relies on firedrake main can only be merged once that functionality lands in firedrake release. This leads to branches with several new features existing for longer, which in general is something we want to discourage.
* Despite our good efforts at test coverage - which also in option 2 would catch API breakages etc. - I think it would be good if at least some of our more experienced users/devs regularly run gadopt with firedrake main on _actual_ cases to spot any major regressions. This of course does have the obvious drawback of having to develop against a moving target, chasing bugs that turn out to be in Firedrake, and would otherwise been fixed upstream if you'd just track firedrake release - but I guess we should share some of that burden (of at least reporting it), and if it isn't fixed before the next release it might take quite a bit longer when we only find out then.

So yes I would argue for 1: we have a main and a release branch that relies on resp. Firedrake main and release, and we more or less follow Firedrake's policy: New features go into main. Release only gets necessary bug-fixes and additions to documentation. Release would be CI tested on a schedule to spot any breakages due to environment changes, or due to firedrake release fixes. Main would regularly be tested when merges happen (and maybe additional scheduled ones if there's a lull in development).

[...]

PR #294, opened 2025-09-24
Don't import interpolate from firedrake.__future__

angus-g:
This has been deprecated for a while.


dsroberts:
This is one pretty convincing argument in favour of #293.

Patol75:
This seems to break the following block in `TimestepAdaptor`:
```python
        # J^-1 u is a discontinuous expression, using op2.MAX it takes the maximum value
        # in all adjacent elements when interpolating it to a continuous function space
        # We do need to ensure we reset ref_vel to zero, as it also takes the max with any previous values
        self.ref_vel_interpolate = interpolate(abs(dot(JacobianInverse(self.mesh), self.u)), V, access=op2.MAX)
```
The `base_case` evolution becomes erroneous. If one decreases the maximum allowed time step, then the evolution is sensible again. I tried fixing the block above but without success. Perhaps @stephankramer will have an idea.

PR #298, opened 2025-10-01
separate nullspaces 

### LINKED ISSUE #297: Proposal: move nullspace helper functions to a separate module
In #268 I want to separate the u-component (i.e. velocity or displacement) out of `create_stokes_nullspace` because in the substitution method for internal variable we are only solving for the displacement, so the logic in the current stokes nullspace setup (which assumes a mixed space) does not work.

My current approach is to have a `create_u_nullspace` function which is called in `create_stokes_nullspace` when setting up the velocity part of the nullspace basis. 

Since there would now be three `create nullspace` functions I think it would make sense to separate these into a separate module `nullspaces.py`

w3168:
Setup velocity (and displacement) nullspace in a separate helper function to `create_stokes_nullspace` to deal with non mixed spaces in #268.

Also move helper nullspace functions to their own module to make things tidier. 

Closes #297

### DIFF
--- a/.github/workflows/render.yml
+++ b/.github/workflows/render.yml
@@ -3,7 +3,7 @@ name: Render demo notebooks
 on:
   push:
     branches:
-      - master
+      - main
   workflow_dispatch:
     inputs:
       trigger:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,7 @@ name: Run regression tests
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
   workflow_dispatch:
     inputs:
--- a/gadopt/stokes_integrators.py
+++ b/gadopt/stokes_integrators.py
@@ -794,10 +794,10 @@ class BoundaryNormalStressSolver(SolverConfigurationMixin):
                 "BoundaryNormalStressSolver: Pressure field is discontinuous. Using an equivalent continous lagrange element."
             )
             Q = fd.FunctionSpace(
-                self.mesh, "Lagrange", self.p.function_space().ufl_element().degree()
+                self.p.function_space().mesh(), "Lagrange", self.p.function_space().ufl_element().degree()
             )
         else:
-            Q = fd.FunctionSpace(self.mesh, self.p.ufl_element())
+            Q = fd.FunctionSpace(self.p.function_space().mesh(), self.p.ufl_element())
 
         self.force = fd.Function(Q, name=f"force_{self.subdomain_id}")
 


gadopt/nullspaces.py
```diff
@@ -0,0 +1,212 @@
...
+            )
+            y_rotV = fd.Function(V).interpolate(
+                fd.as_vector((X[2], 0, -X[0]))
+            )
+            z_rotV = fd.Function(V).interpolate(
+                fd.as_vector((-X[1], X[0], 0))
+            )
+            basis = [x_rotV, y_rotV, z_rotV]
+        else:
+            raise ValueError("Unknown dimension")
```
dsroberts, line 194:
```suggestion
            raise ValueError("Can only handle 2 or 3 dimensional spaces")
```
A more human readable error message.

gadopt/nullspaces.py
```diff
@@ -0,0 +1,212 @@
+"""Helper functions to generate nullspaces for stokes problems
+
+The module contains a function, `create_stokes_nullspace`,
+to automatically generate null spaces compatible with PETSc
+solvers for the mixed velocity-pressure Stokes system
+as well as the function `create_u_nullspace' which isolates
+the translational and rotational nullspaces associated with
+the velocity (or displacement) field.
```
dsroberts, line 8: Is it 'null space' or 'nullspace'?

dsroberts (changes requested):
A good addition overall, a few minor comments from me. If the docstrings for these functions need updating, now is the time to do it. Also, while you're here, can you reduce the docstring widths in `ala_right_nullspace` down to 88 characters where possible?

gadopt/nullspaces.py
w3168: I'll change it to 'null space' based on https://gmd.copernicus.org/articles/15/5127/2022/ and https://en.wikipedia.org/wiki/Kernel_(linear_algebra)

gadopt/nullspaces.py
```diff
@@ -0,0 +1,216 @@
+"""Helper functions to generate null spaces for stokes problems
```
angus-g, line 1:
```suggestion
"""Helper functions to generate null spaces for Stokes problems
```

gadopt/nullspaces.py
```diff
@@ -0,0 +1,216 @@
...
+
+    return fd.MixedVectorSpaceBasis(Z, null_space)
+
+
+def create_u_nullspace(
+    V: fd.functionspaceimpl.WithGeometry,
+    rotational: bool = False,
+    translations: list[int] | None = None,
+) -> fd.nullspace.VectorSpaceBasis:
+    """Create a null space for the velocity (or displacement) in a Stokes system
```
angus-g, line 166: Just on a very fiddly naming point, is `create_u_nullspace` the most descriptive we can get? I *would* suggest `create_velocity_nullspace`, but you do include the possibility of displacement as well.

gadopt/nullspaces.py
```diff
@@ -0,0 +1,216 @@
...
+      V: Firedrake function space associated with the velocity or displacement
+      rotational: Whether to include all rotational modes
+      translations: List of translations to include
+
+    Returns:
+      A Firedrake vector space basis incorporating the null space components
+
+    """
+    X = fd.SpatialCoordinate(V.mesh())
+    dim = len(X)
```
angus-g, line 178: *Maybe* prefer `V.mesh().geometric_dimension()`? At least the intent is clearer than going through the spatial coordinate, not that it should make a difference.

gadopt/nullspaces.py
```diff
@@ -0,0 +1,216 @@
...
+    ala_approximation: AnelasticLiquidApproximation | None = None,
+    top_subdomain_id: str | int | None = None,
+) -> fd.nullspace.MixedVectorSpaceBasis:
+    """Create a null space for the mixed Stokes system.
+
+    Arguments:
+      Z: Firedrake mixed function space associated with the Stokes system
+      closed: Whether to include a constant pressure null space
+      rotational: Whether to include all rotational modes
+      translations: List of translations to include
```
angus-g, line 115: While we're at it, probably good to document what the `translations` list should look like?

gadopt/nullspaces.py
w3168: I guess the point is it should be general for velocity and displacement. Maybe we could call it rigid body nullspace ?

gadopt/nullspaces.py
w3168: thanks!

w3168:
can we merge this?

drhodrid:
Looks good to me and happy for this to go in, provided a test is added in a subsequent (and soonish) merge?

angus-g (approved):

PR #299, opened 2025-10-01
Diagnostics Base Class

### LINKED ISSUE #295: Proposal: Diagnostics base class
In #268, there are a few places where the maximum and minimum displacement are used as diagnostic quantities. The method by which this is calculated looks very similar to `GeodynamicalDiagnostics.ux_max`. To save reimplementing stuff like this as G-ADOPT functionality expands, I suggest we create a base diagnostics class that contains useful routines like `Function` max/min across the whole domain/boundary, domain volume/surface calculations and figuring out measures. `GeodynamicalDiagnostics` or a hypothetical `GIADiagnostics` could inherit from that and provide the same API it does now, but drawing from these already implemented base-level routines. E.g. 
```
class BaseDiagnostics:
    def __init__(self):
        ...

    def _function_max(f):
        f_data = f.dat.data_ro
        return f.comm.allreduce(f_data.max(), MPI.MAX)

class GeodynamicalDiagnostics(BaseDiagnostics):
    def __init__(self, z, T, ...):
        self.z = z
        self.T = T
       ...

   def T_max(self):
      return self._function_max(self.T)
```

dsroberts:
A base class to handle common diagnostic routines. Treating each function separately like this may seem like overkill, however I'm anticipating this will see use with the self gravity work, so its designed from the start to support diagnostics across multiple meshes. Closes #295.

### DIFF
--- a/gadopt/diagnostics.py
+++ b/gadopt/diagnostics.py
@@ -4,17 +4,69 @@ parameters and call individual class methods to compute associated diagnostics.
 
 """
 
-from firedrake import (
-    Constant, DirichletBC, FacetNormal, Function,
-    assemble, dot, ds, dx, grad, norm, sqrt,
-)
+from firedrake import Constant, DirichletBC, FacetNormal, Function, assemble, dot, ds, dx, grad, norm, sqrt
 from firedrake.ufl_expr import extract_unique_domain
 from mpi4py import MPI
 
 from .utility import CombinedSurfaceMeasure
 
 
-class GeodynamicalDiagnostics:
+class FunctionAttributeHolder:
+    def __init__(self, quad_degree: int, func: Function):
+        self.mesh = extract_unique_domain(func)
+        self.function_space = func.function_space()
+        self.dx = dx(domain=self.mesh, degree=quad_degree)
+        self.ds = CombinedSurfaceMeasure(self.mesh, quad_degree) if self.function_space.extruded else ds(self.mesh)
+        self.normal = FacetNormal(self.mesh)
+        self.volume = assemble(Constant(1) * self.dx)
+        # Fill in as necessary
+        self.boundary_nodes: dict[int, list[int]] = {}
+
+    def get_boundary_nodes(self, boundary_id: int) -> list[int]:
+        if boundary_id not in self.boundary_nodes:
+            bc = DirichletBC(self.function_space, 0, boundary_id)
+            tmp = Function(self.function_space)
+            self.boundary_nodes[boundary_id] = [n for n in bc.nodes if n < len(tmp.dat.data_ro)]
+        return self.boundary_nodes[boundary_id]
+
+
+class BaseDiagnostics:
+    def __init__(self, quad_degree: int, **funcs: Function):
+        self._attrs: dict[Function, FunctionAttributeHolder] = {}
+
+        for name, func in funcs.items():
+            if len(func.subfunctions) == 1:
+                setattr(self, name, func)
+                self._init_single_func(quad_degree, func)
+            else:
+                for i, subfunc in enumerate(func.subfunctions):
+                    setattr(self, f"{name}_{i}", func)
+                    self._init_single_func(quad_degree, subfunc)
+
+    def _init_single_func(self, quad_degree: int, func: Function):
+        self._attrs[func] = FunctionAttributeHolder(quad_degree, func)
+
+    def _function_max(self, f: Function, boundary_id: int | None = None):
+        if boundary_id:
+            f_data = f.dat.data_ro[self._attrs[f].get_boundary_nodes(boundary_id), 0]
+        else:
+            f_data = f.dat.data_ro
+
+        return f.comm.allreduce(f_data.max(), MPI.MAX)
+
+    def _function_min(self, f: Function, boundary_id: int | None = None):
+        if boundary_id:
+            f_data = f.dat.data_ro[self._attrs[f].get_boundary_nodes(boundary_id), 0]
+        else:
+            f_data = f.dat.data_ro
+
+        return f.comm.allreduce(f_data.min(), MPI.MIN)
+
+    def _function_avg(self, f: Function):
+        return assemble(f * self._attrs[f].dx) / self._attrs[f].volume
+
+
+class GeodynamicalDiagnostics(BaseDiagnostics):
     """Typical simulation diagnostics used in geodynamical simulations.
 
     Arguments:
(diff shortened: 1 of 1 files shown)


gadopt/diagnostics.py
```diff
@@ -49,56 +240,49 @@ def __init__(
...
 
-        self.u, self.p = z.subfunctions[:2]
-        self.T = T
-
-        self.dx = dx(domain=mesh, degree=quad_degree)
-        self.domain_volume = assemble(Constant(1) * self.dx)
-
-        if self.u.function_space().extruded:
-            self.ds = CombinedSurfaceMeasure(mesh, quad_degree)
+        if isinstance(T, Function):
```
stephankramer, line 246: I realize this is actually an issue in master already, but above we have
```
   T: Function = 0.0
```
which doesn't make sense as 0.0 is not a Function :-)
So I think I would prefer:
```
   T: Function | None = None
```
and then here we'd simply
```
   if T:
      self.T = T 
      super().__init__(quad_degree, u=u, p=p, T=T)
  else:
      self.T = 0.0
      super().__init__(quad_degree, u=u, p=p)
```

dsroberts: <Pedantic hat on> `super().__init__(..., T=T)` gives you `self.T = T` anyway, and what you've proposed isn't safe either as `T` may evaluate to false. The correct way to do this is:
```
if T is None:
      self.T = 0.0
      super().__init__(quad_degree, u=u, p=p)
  else:
      super().__init__(quad_degree, u=u, p=p, T=T)
```

gadopt/diagnostics.py
```diff
@@ -1,20 +1,211 @@
...
+          Maximum value of f across the specified domain/component
+        """
+        if f in self:
+            if boundary_id:
+                f_data = f.dat.data_ro[self._attrs[f].get_boundary_nodes(boundary_id)]
+            else:
+                f_data = f.dat.data_ro
+            if dim is not None and self.dim_valid(f):
+                f_data = f_data[:, dim]
+            return f.comm.allreduce(f_data.max(), MPI.MAX)
```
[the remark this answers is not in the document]
dsroberts: This was the cause of the test failure. I wonder why its only showing up now?

w3168:
Thanks @dsroberts I think this is all good! One thing that would be nice is an option to get the max/min etc of radial and tangential components of vectors in spherical geometry. 

I hacked something together in #268 but it would be nice to make it more general. I don't think we should delay this PR because the current CI tests only check the horizontal velocity component so any addition would be untested.

Patol75:
@dsroberts Now that #268 is in, what is the way forward for this PR?

dsroberts:
I've decided I'll follow UFL and make all of the objects held by a `FunctionContext` `@property`s. Other than that it'll be a matter of integrating the new diagnostics from #268. I think its best to hold off on this until after the workshops as I'll be on leave next week and the week after. That way there are no surprises and I can handle any new diagnostics that come out of the Richardson stuff at the same time.

gadopt/diagnostics.py
```diff
@@ -38,94 +599,109 @@ class GeodynamicalDiagnostics:
...
+
+    def uv_max(self, boundary_id: int | None = None) -> float:
+        "Maximum value of vertical component of velocity/displacement"
+        return self.max(self.get_radial_component(self.u), boundary_id)
+
+    def l2_norm_surface(self) -> float:
+        return self.l2norm(self.get_radial_component(self.u), self.top_id)
+
+    def l1_norm_surface(self) -> float:
+        return self.l1norm(self.get_radial_component(self.u), self.top_id)
```
angus-g: `surface` vs. `top`? I think `top` is probably slightly clearer, given that the bottom is also a surface?

w3168:
Thanks @dsroberts it looks much cleaner from an api point of view for a gia user!

gadopt/diagnostics.py
```diff
@@ -1,22 +1,583 @@
...
+        """Initialise a BaseDiagnostics object.
+
+        Sets the `quad_degree` for measures used by this object and passes the
+        remaining keyword arguments through to `register_functions`.
+
+        Args:
+            quad_degree (int): The quadrature degree for the measures held by this object
+            **funcs (fd.Function | None): key-value pairs of Firedrake functions to associate
+            with this instance
+        """
```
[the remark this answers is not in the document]
dsroberts: Google has lied to me then: https://google.github.io/styleguide/pyguide.html#384-classes

gadopt/diagnostics.py
dsroberts: True. I just used @w3168's naming convention from the GIA tests. Will change.

tests/3d_spada/3d_spada.py
```diff
@@ -298,9 +298,9 @@
 
 plog = ParameterLog("params.log", mesh)
 plog.log_str(
-    "timestep time dt u_rms u_rms_surf ux_max uv_min"
+    "timestep time dt u_rms u_rms_surf ux_max uv_min disp_min disp_max"
```
Patol75:
```suggestion
    "timestep time dt u_rms u_rms_surf ux_max disp_min disp_max"
```

gadopt/diagnostics.py
```diff
@@ -1,22 +1,578 @@
...
+
+    This class is intended to be subclassed by domain-specific diagnostic classes
+
+    Args:
+        quad_degree: The quadrature degree for the measures held by this object
+        **funcs: key-value pairs of Firedrake functions to associate with this
+        instance
+    """
+
+    def __init__(self, quad_degree: int, **funcs: fd.Function | None):
```
Patol75:
```suggestion
    def __init__(self, quad_degree: int, **funcs: fd.Function):
```

gadopt/diagnostics.py
```diff
@@ -1,22 +1,578 @@
...
+            boundary_id (optional): Boundary ID. If not provided or set to `None`,
+            will integrate across entire domain. If provided, will integrate along
+            the specified boundary only. Defaults to None.
+
+        Returns:
+            float: L2 norm
+        """
+        self._check_present(f)
+        measure = self._get_measure(f, boundary_id)
+        return fd.sqrt(fd.assemble(fd.dot(f, f) * measure))
```
Patol75: Firedrake provides [`norm`](https://www.firedrakeproject.org/firedrake.html#firedrake.norms.norm). Worth using here?

gadopt/diagnostics.py
```diff
@@ -38,94 +594,131 @@ class GeodynamicalDiagnostics:
...
-        else:
-            self.ds = ds(mesh)
+        u, p = z.subfunctions[:2]
+        super().__init__(quad_degree, u=u, p=p, T=T)
 
         if bottom_id:
-            self.ds_b = self.ds(bottom_id)
-            self.bottom_surface = assemble(Constant(1) * self.ds_b)
+            self.ds_b = self._function_contexts[self.u].ds(bottom_id)
+            self.bottom_surface = fd.assemble(fd.Constant(1) * self.ds_b)
```
Patol75: Should we provide a function to handle such a surface integral in the same way as `get_volume`?

Patol75 (changes requested):
Awesome work, @dsroberts. Here are some minor comments and suggestions.

gadopt/utility.py
```diff
@@ -529,7 +529,7 @@ def values(self):
         return self.__dict__.values()
 
     def __contains__(self, item):
-        return item in self.__dict__
+        return item in self.values()
```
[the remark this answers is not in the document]
dsroberts: What I care about is whether boundary IDs (i.e. 1,2,'top','bottom',etc) are in the `BoundaryIDNamespace` instance, not the `left`, `right`, etc. I made a mistake with #341 originally, the above change is what I intended.

tests/3d_weerdesteijn/3d_weerdesteijn.py
```diff
@@ -421,25 +420,19 @@
...
-    bc_displacement = DirichletBC(vertical_displacement.function_space(), 0, boundary.top)
-    displacement_z_min = vertical_displacement.dat.data_ro_with_halos[bc_displacement.nodes].min(initial=0)
     # Minimum displacement at surface (should be top left corner with
     # greatest (-ve) deflection due to ice loading
-    displacement_min = vertical_displacement.comm.allreduce(displacement_z_min, MPI.MIN)
+    displacement_min = gd.uv_min(boundary.top) * D
     log("Greatest (-ve) displacement", displacement_min)
-    displacement_z_max = vertical_displacement.dat.data_ro_with_halos[bc_displacement.nodes].max(initial=0)
-    displacement_max = vertical_displacement.comm.allreduce(displacement_z_max, MPI.MAX)
+    displacement_max = gd.uv_max(boundary.top) * D
```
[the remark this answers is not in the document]
dsroberts: The `uv_min`/`uv_max` functions don't take a scale argument. I did talk with @w3168 about this, and he said that a `scale` argument was not appropriate here multiplying by `D` is a stand-in for redimensionalising. I think the idea being that if we're going to incorporate redimensionalising quantities into the diagnostics, we should do it properly.

gadopt/diagnostics.py
```diff
@@ -1,22 +1,578 @@
...
+            boundary_id (optional): The boundary ID to check. If set to `None`,
+            assume we're performing a volume integral, so no boundary ID is necessary.
+            Otherwise check the boundary ID against those derived from the mesh
+            belonging to this instance. Defaults to None.
+
+        Raises:
+            KeyError: Mesh does not have a boundary corresponding to `boundary_id`
+        """
+        if boundary_id is None:
+            return
```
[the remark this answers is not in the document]
dsroberts: I'll remove the default value, but leave that as it is as most functions that call `_check_boundary_id` are allowed to set `boundary_id=None`. `boundary_id=None` indicates an integral across the entire domain and is valid here.

gadopt/diagnostics.py
```diff
@@ -1,22 +1,578 @@
...
+            across all components. Defaults to None.
+
+        Returns:
+            The minimum and negative maximum value of the function over the
+            specified domain
+        """
+        self._check_present(func_or_op)
+        if isinstance(func_or_op, ufl.core.operator.Operator):
+            f = self._get_func_for_op(func_or_op)
+            f.interpolate(func_or_op)
```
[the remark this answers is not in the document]
dsroberts: Maybe in future. None of the existing diagnostics use `project`, and my understanding is that there is some additional complexity here as we need to set solver options in that case? If there is a compelling need to `project` a diagnostic quantity instead of `interpolate`, then we can cross that bridge when we get to it.

gadopt/diagnostics.py
dsroberts: I did look at that, but decided to go this way instead. There are a bunch of L2 norm calculations throughout G-ADOPT that are implemented this way. And I think Firedrake's `norm` function is way over-generalised for our needs, we don't need to handle complex numbers, we don't need to handle anything other than L1 and L2 at this stage. Firedrake's L1 norm is way more complicated than ours:
```
assemble((inner(f,f)**(1/2))*dx)**(1/1)
```
vs
```
assemble(abs(f) * dx)
```

gadopt/diagnostics.py
dsroberts: Makes sense, will do.

gadopt/diagnostics.py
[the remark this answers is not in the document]
Patol75: I found in some of my scripts that bare `project` could give better results than `interpolate`, especially when gradients are involved. But we can indeed revisit later.

gadopt/diagnostics.py
```diff
@@ -1,22 +1,578 @@
...
+        ```
+        If an input function is set to `None`, the attribute will still be created
+        but set to 0.0. If a mixed function is entered, each subfunction will have
+        a `FunctionContext` object associated with it, and the attribute will be named
+        with an additional number to denote the index of the subfunction i.e.:
+
+        ```
+        > diag.register_functions(self,F)
+        > type(diag.F)
+        AttributeError: 'BaseDiagnostics' object has no attribute 'F'
```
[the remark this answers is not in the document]
Patol75: Maybe we could raise a different error, for example: `Requested F, which lives on a mixed space. Instead, access subfunctions via F_0, F_1, etc...`.

PR #302, opened 2025-10-07
Request Riesz representation on derivative calls

angus-g:
For testing and visualisation, we want to be dealing with the primal derivative quantity (the gradient), so we need to explicitly apply the Riesz representation. This is the API change due to the `AbstractReducedFunctional` in pyadjoint.

### DIFF
--- a/demos/PDE_constrained_optimisation/PDE_constrained_boundary.py
+++ b/demos/PDE_constrained_optimisation/PDE_constrained_boundary.py
@@ -125,7 +125,15 @@ tape.clear_tape()
 
 T0.project(T_wrong)
 
-m = Control(T0)
+# At this stage we need to define the control variable for the adjoint-based optimisation problem
+# by wrapping T0 in a Control object.
+# This registers T0 as the optimisation variable with pyadjoint's automatic differentiation framework.
+# Furthermore, the L2 Riesz map specification is crucial for two reasons:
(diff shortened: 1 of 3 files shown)


angus-g:
Just playing around, if we see the remaining failures are from the other release issues we can merge this into #307 and batch it all together.

angus-g:
@sghelichkhani mentioned that because the specification of the Riesz representation has changed, we should be a little more explicit in our description of how to configure this. That's just a text change, then this is good to merge.

sghelichkhani (approved):
I added some text where it needs to explain why riesz_map in the demos. 

Will probably need to revisit this again. If the default maping is `None`, I would not be surprised if things need to be adjusted in the adjoint regression tests.

angus-g:
> If the default maping is None, I would not be surprised if things need to be adjusted in the adjoint regression tests.

Not sure what you mean by this? The default is L2 if none is specified.

sghelichkhani:
> > If the default maping is None, I would not be surprised if things need to be adjusted in the adjoint regression tests.
> 
> Not sure what you mean by this? The default is L2 if none is specified.

I see! Sorry I was only looking at pyadjoint side! 

I think you are correct.  In `firedrake/adjoint_utils/function.py:L222` we have
 
```python3
        return value.riesz_representation(riesz_map=riesz_map or "L2")
```

PR #307, opened 2025-10-07
Batch fixes for Firedrake 2025.10.1 release

angus-g:
This still needs the pyadjoint changes to land, but we can set the merge base for the existing release fixes into this branch.


angus-g:
Now that this has all the constituent components, I'll make sure it passes the tests then drop the pyadjoint commit -- it'll fail at that point until we get a new Firedrake container.

PR #308, opened 2025-10-08
2d cylindrical demo via internal variables

w3168:
Converting the 2d cylindrical maxwell benchmark via the zhong formulation to a compressible internal variable case.

### DIFF
--- a/demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
+++ b/demos/multi_material/compositional_buoyancy/compositional_buoyancy.py
@@ -131,9 +131,10 @@ assign_level_set_values(
 
 # fig, axes = plt.subplots()
 # axes.set_aspect("equal")
-# contours = tricontourf(psi, levels=linspace(0.0, 1.0, 11), axes=axes, cmap="PiYG")
-# tricontour(psi, axes=axes, levels=[0.5])
-# fig.colorbar(contours, label="Conservative level set")
+# axes.margins(0.0)
+# pcolor = tripcolor(psi, cmap="PiYG", axes=axes)
+# tricontour(psi, levels=[0.5], axes=axes)
+# fig.colorbar(pcolor, label="Conservative level set")
 # -
 
 # We next define the material fields and instantiate the approximation. Here, the system
@@ -160,10 +161,9 @@ approximation = BoussinesqApproximation(Ra, RaB=RaB)
 # + tags=["active-ipynb"]
 # fig, axes = plt.subplots()
 # axes.set_aspect("equal")
-# contours = tricontourf(
-#     Function(psi).interpolate(RaB), levels=linspace(0.0, 1.0, 11), axes=axes
-# )
-# fig.colorbar(contours, label="Compositional Rayleigh number")
+# axes.margins(0.0)
+# pcolor = tripcolor(Function(psi).interpolate(RaB), axes=axes)
+# fig.colorbar(pcolor, label="Compositional Rayleigh number")
 # -
 
 # As with the previous examples, we set up an instance of the `TimestepAdaptor` class
@@ -293,7 +293,7 @@ while True:
 # + tags=["active-ipynb"]
 # fig, axes = plt.subplots()
 # axes.set_aspect("equal")
-# contours = tricontourf(psi, levels=linspace(0.0, 1.0, 11), axes=axes, cmap="PiYG")
-# tricontour(psi, axes=axes, levels=[0.5])
+# contours = tricontourf(psi, levels=linspace(0.0, 1.0, 11), cmap="PiYG", axes=axes)
+# tricontour(psi, levels=[0.5], axes=axes)
 # fig.colorbar(contours, label="Conservative level-set")
 # -
--- a/demos/multi_material/thermochemical_buoyancy/thermochemical_buoyancy.py
+++ b/demos/multi_material/thermochemical_buoyancy/thermochemical_buoyancy.py
@@ -137,8 +137,8 @@ assign_level_set_values(psi, epsilon, signed_distance)
 #
 # fig, axes = plt.subplots()
 # axes.set_aspect("equal")
-# contours = tricontourf(psi, levels=linspace(0.0, 1.0, 11), axes=axes, cmap="PiYG")
-# tricontour(psi, axes=axes, levels=[0.5])
+# contours = tricontourf(psi, levels=linspace(0.0, 1.0, 11), cmap="PiYG", axes=axes)
+# tricontour(psi, levels=[0.5], axes=axes)
 # fig.colorbar(contours, label="Conservative level-set")
 # -
 
@@ -202,8 +202,9 @@ T.interpolate(max_value(min_value(Tu + Tl + Tr + Ts - 3.0 / 2.0, 1.0), 0.0))
 # + tags=["active-ipynb"]
 # fig, axes = plt.subplots()
 # axes.set_aspect("equal")
-# contours = tricontourf(T, levels=linspace(0.0, 1.0, 11), axes=axes, cmap="inferno")
-# fig.colorbar(contours, label="Temperature")
+# axes.margins(0.0)
+# pcolor = tripcolor(T, cmap="inferno", axes=axes)
+# fig.colorbar(pcolor, label="Temperature")
 # -
 
 # As with the previous examples, we set up an instance of the `TimestepAdaptor` class
@@ -341,6 +342,6 @@ while True:
 # + tags=["active-ipynb"]
 # fig, axes = plt.subplots()
 # axes.set_aspect("equal")
-# contours = tricontourf(T, levels=linspace(0.0, 1.0, 11), axes=axes, cmap="inferno")
-# tricontour(psi, axes=axes, levels=[0.5])
+# contours = tricontourf(T, levels=linspace(0.0, 1.0, 11), cmap="inferno", axes=axes)
+# tricontour(psi, levels=[0.5], axes=axes)
 # fig.colorbar(contours, label="Temperature")


demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -3,12 +3,13 @@
 #
 # In this tutorial, we examine an idealised 2-D loading problem in an annulus domain.
 #
-# This example focusses on differences between running simulations in a 2-D annulus and 2-D Cartesian domain.
-# These can be summarised as follows:
+# This example focusses on differences between running simulations in a 2-D annulus
+# and 2-D Cartesian domain. These can be summarised as follows:
```
drhodrid: Add our lovely new way to link back to a simpler tutorial -- i.e. your base case here.

w3168: added

demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -3,12 +3,13 @@
...
 # 1. The geometry of the problem - i.e. the computational mesh.
-# 2. The radial direction of gravity (as opposed to the vertical direction in a Cartesian domain).
+# 2. The radial direction of gravity (as opposed to the vertical direction in a
+# Cartesian domain).
 # 3. Solving a problem with laterally varying viscosity.
```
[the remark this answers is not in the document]
w3168: added

demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -17,61 +18,78 @@
...
 
 # +
 # Set up geometry:
-rmin = 3480e3
-rmax = 6371e3
-D = rmax-rmin
-nz = 32
-ncells = 180
-dz = D / nz
+radius_values = [6371e3, 6301e3, 5951e3, 5701e3, 3480e3]
```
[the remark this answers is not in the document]
w3168: added a comment about spada

demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -17,61 +18,78 @@
...
 # We next set up the function spaces, and specify functions to hold our solutions.
 
 # +
-# Set up function spaces - currently using the bilinear Q2Q1 element pair:
-V = VectorFunctionSpace(mesh, "CG", 2)  # (Incremental) Displacement function space (vector)
-W = FunctionSpace(mesh, "CG", 1)  # Pressure function space (scalar)
-S = TensorFunctionSpace(mesh, "DQ", 2)  # (Discontinuous) Stress tensor function space (tensor)
+V = VectorFunctionSpace(mesh, "CG", 2)  # Displacement function space
+S = TensorFunctionSpace(mesh, "DQ", 1)  # Stress tensor function space
+DG0 = FunctionSpace(mesh, "DQ", 0)  # Density/shear modulus function space
```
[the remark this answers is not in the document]
w3168: changed

demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -88,44 +106,49 @@
...
 # -
 
-# Let's start initialising some parameters. First of all let's get the (symbolic) spatial coordinates of the mesh
+# Let's start initialising some parameters. First of all let's get the (symbolic)
+# spatial coordinates of the mesh
 
 X = SpatialCoordinate(mesh)
 
 # Now we can set up the background profiles for the material properties.
 # In this case the density and shear modulus vary in the vertical direction.
```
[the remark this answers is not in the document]
w3168: changed

demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -185,136 +216,172 @@ def bivariate_gaussian(x, y, mu_x, mu_y, sigma_x, sigma_y, rho, normalised_area=
 
 def setup_heterogenous_viscosity(viscosity):
     heterogenous_viscosity_field = Function(viscosity.function_space(), name='viscosity')
-    antarctica_x, antarctica_y = -2e6, -5.5e6
+    antarctica_x, antarctica_y = -2e6/D, -5.5e6/D
```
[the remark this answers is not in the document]
w3168: I think I had this comment 
`We'll put some regions of low viscosity near the South Poleouth Pole (inspired by West Antarctica)` so maybe that is ok? This is going to move into a later demo so should be easier to see the difference then.

demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -185,136 +216,172 @@ def bivariate_gaussian(x, y, mu_x, mu_y, sigma_x, sigma_y, rho, normalised_area=
...
 
-    low_viscosity_antarctica = bivariate_gaussian(X[0], X[1], antarctica_x, antarctica_y, 1.5e6, 0.5e6, -0.4)
-    heterogenous_viscosity_field.interpolate(-3*low_viscosity_antarctica + viscosity * (1-low_viscosity_antarctica))
+    low_visc = 1e20/viscosity_scale
```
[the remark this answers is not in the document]
w3168: This is more about just setting up a lower viscosity than the average by an order of mangitude. Maybe we could make it lower but I haven't tried to push it that far yet in an inversion...

gadopt/gia_demo_utilities.py
```diff
@@ -0,0 +1,287 @@
+"""This module provides helper functions for plotting, associated with the
+glacial isostatic adjustment demos, via Pyvista. This includes plotting
+'ice rings' showing ice thickness offset from the surface of
+the domain. This module also default options for plotting viscosity
```
Patol75: `has default`?

gadopt/gia_demo_utilities.py
```diff
@@ -0,0 +1,287 @@
...
+    colatitude = atan2(X[0], X[1])
+
+    # Position opposite disc centre in radians
+    opp = disc_centre - pi if disc_centre >= 0 else disc_centre + pi
+
+    # Angular distance accounting for discontinuity at 'south pole'
+    angular_distance = conditional(abs(colatitude - disc_centre) < pi,
+                                   abs(colatitude-disc_centre),
+                                   pi - abs(colatitude-opp)
+                                   )
```
Patol75:
```python
angular_distance_raw = abs(colatitude - disc_centre)
angular_distance = conditional(
    angular_distance_raw < pi, angular_distance_raw, 2 * pi - angular_distance_raw
)
```

demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -3,12 +3,13 @@
 #
 # In this tutorial, we examine an idealised 2-D loading problem in an annulus domain.
 #
-# This example focusses on differences between running simulations in a 2-D annulus and 2-D Cartesian domain.
-# These can be summarised as follows:
+# This example focuses on differences between running simulations in a 2-D annulus
+# and a 2-D Cartesian domain in our [previous tutorial](../base_case). These can be summarised
+# as follows:
```
Patol75:
``` python
# This example focuses on differences between running simulations in a 2-D cylindrical annulus
# and a 2-D Cartesian box, which was employed in our [previous tutorial](../base_case). These can be summarised
# as follows:
```

demos/glacial_isostatic_adjustment/2d_cylindrical/2d_cylindrical.py
```diff
@@ -88,525 +109,228 @@
...
 # -
 
-# Next let's initialise the viscosity field. In this tutorial we are
-# going to make things a bit more interesting by using a laterally
-# varying viscosity field. We'll put some regions of low viscosity
-# near the South Pole (inspired by West Antarctica) as well as in the lower mantle.
-# We've also put some relatively higher viscosity patches of mantle in the
-# northern hemisphere to represent a downgoing slab.
+# Now let's setup the ice load. For this tutorial we will have two synthetic ice sheets.
+# Let's put one a larger one over the South Pole, with a total horizontal
```
Patol75: Extra `one`

demos/glacial_isostatic_adjustment/2d_cylindrical_lvv/2d_cylindrical_lvv.py
```diff
@@ -0,0 +1,420 @@
...
+
+    heterogenous_viscosity_field.interpolate(
+        conditional(vertical_component(X) > radius_values_nondim[1],
+                    viscosity,
+                    heterogenous_viscosity_field))
+
+    return heterogenous_viscosity_field
+
+
+viscosity = setup_heterogenous_viscosity(background_viscosity)
```
Patol75: Is this block the main difference from the other demo? If yes, I do not think it warrants a separate demo file. This block does not involve any G-ADOPT functionality. Could it be possible to include the isoviscous and non-isoviscous scenarios in the same file?

Patol75 (changes requested):
Some comments for now.

demos/glacial_isostatic_adjustment/2d_cylindrical_lvv/2d_cylindrical_lvv.py
dsroberts: If this were a test I'd agree, but I think its important to demonstrate the progression from one case to the next. After all `2d_compressible_ALA.py` and `2d_compressible_TALA.py` have much more in common than this file and `2d_cylindrical.py`.

w3168: Thanks @Patol75 and @dsroberts I think I agree with Thomas that it might be better to have the radial and lateral case in the same demo just running the lateral case after. I guess we don't do that in other demos but maybe it makes sense...

@dsroberts I guess the point of the demo was more that we can run laterally varying cases not exactly about how we set this specific viscosity field. It was really just about using 'blobs' to create something that looked interesting and had obvious lateral variations in viscosity. I should a few comments in the demo would help to clarify that. I was actually tempted to move `setup_heterogenous_viscosity` to `gia_demo_utilities.py` but  I am not sure now given your comment about moving the `ice_sheet_disc` function back into the demo.

Another probably more useful demo might show how to load a predefined viscosity file to a function.

w3168: actually ha I realised i changed the name from antarctica_x to southple_x at somepoint during earlier reviews which is a bit unclear! I guess I meant centre of the blob near the pretend antarctica/south pole. I'll think of a better name

w3168: I added more comments and docstrings to the lateral viscosity function so hopefully that helps with some of the ambiguity

PR #317, opened 2025-10-22
Fix multi-material plots in demos

Patol75:
I have noticed on the website ([here](https://gadopt.org/tutorials/multi_material/compositional_buoyancy/) and [here](https://gadopt.org/tutorials/multi_material/thermochemical_buoyancy/)) that some `tricontourf` plots in multi-material demos do not render properly. In these cases, I replace `tricontourf` by `tripcolor`.

Could someone remind me if there is a way to render (and how to visualise) the website based on a PR? Thank you.


dsroberts (approved):
LGTM

PR #321, opened 2025-10-22
Add multi-material 2-D cylindrical demo

Patol75:
Multi-material capabilities are not yet tested in non-Cartesian geometries. Here, I adapt the mantle-convection 2-D cylindrical demo to include a compositional basal layer, thereby proposing a test case for the level-set, energy, and Stokes solvers in an annulus.

### DIFF
--- a/demos/multi_material/2d_cylindrical/2d_cylindrical.py
+++ b/demos/multi_material/2d_cylindrical/2d_cylindrical.py
@@ -142,7 +142,9 @@ approximation = BoussinesqApproximation(Ra, RaB=RaB, mu=mu)
 # the compositional Rayleigh number across the domain.
 
 # + tags=["active-ipynb"]
-# VTKFile("RaB.pvd").write(Function(psi).interpolate(RaB))
+# VTKFile("RaB.pvd").write(
+#     Function(psi, name="Compositional Rayleigh number").interpolate(RaB)
+# )
 # RaB_data = pv.read("RaB/RaB_0.vtu")
 # plotter = pv.Plotter(notebook=True)
 # plotter.add_mesh(RaB_data, cmap="BrBG")
@@ -166,10 +168,10 @@ t_adapt = TimestepAdaptor(
 
 stokes_nullspace = create_stokes_nullspace(Z, rotational=True)
 
-# Boundary conditions are next specified. Boundary conditions for temperature are set to
-# $T = 0.0$ at the surface ($r_{\text{max}}$) and $T = 1.0$ at the base
-# ($r_{\text{min}}$). For velocity, we implicitly specify free‐slip conditions on both
-# boundaries by explicitly setting the normal velocity component to 0.
+# Boundary conditions are specified next. For temperature, they are set to $T = 0.0$ at
+# the surface ($r_{\text{max}}$) and $T = 1.0$ at the base ($r_{\text{min}}$). For
+# velocity, we implicitly specify free‐slip conditions on both boundaries by explicitly
+# setting the normal velocity component to 0.
 
 temp_bcs = {boundary.bottom: {"T": 1.0}, boundary.top: {"T": 0.0}}
 stokes_bcs = {boundary.bottom: {"un": 0.0}, boundary.top: {"un": 0.0}}


Patol75:
Animation of the proposed demo

https://github.com/user-attachments/assets/635b3ff3-a976-42c9-aa3a-77754213e908

demos/multi_material/2d_cylindrical/2d_cylindrical.py
```diff
@@ -0,0 +1,320 @@
+# Idealised 2-D multi-material mantle convection problem inside an annulus
+# =
+
+# In this tutorial, we analyse mantle flow in a 2-D annulus domain whereby a denser and
+# more viscous layer, which could represent some lower-mantle structures on Earth, is
```
drhodrid: Be a little more quantitative here -- how much denser, how much more viscous?

demos/multi_material/2d_cylindrical/2d_cylindrical.py
```diff
@@ -0,0 +1,320 @@
...
+# initially defined in the lower third of the domain.
+
+# This example highlights the effect of the denser and more viscous layer on the global
+# mantle dynamics already presented in a
```
drhodrid:
```suggestion
# dynamics already presented in a
```

drhodrid (commented):
I have added a few comments -- perhaps some duplication of the 2-D cylindrical case can be removed to reduce length and focus on those level-set aspects? I think its a nice demo otherwise. 

I've not yet run it, but I do wonder if a higher buoyancy number would be more appropriate for a demo, really showcasing how the different material properties influence the solution (i.e. less mingling between layers etc...)? Otherwise well done for putting together. 

Is there a plan for a simple demo on how to initialise different level set geometries (circle, line, square etc...) in 2-D Cartesian given all of the new functionality added?

demos/multi_material/2d_cylindrical/2d_cylindrical.py
Patol75: Is it important at this stage? Perhaps I can spell it out where I define the material properties.

Patol75:
Just a quick comment: there seems to be an issue with the viscosity. If I define `mu_dense := 5.0`, then `StokesSolver` quickly diverges. It is not clear to me why.

Patol75:
Viscosity issue fixed: it was the unwanted `constant_jacobian=True`.

I have updated a few comments and decided to use an `arithmetic` scheme for `RaB` and `geometric` scheme for viscosity (despite the actual value ranges) to be consistent with the choices made for the dimensional test cases.

Additionally, I have added a missing `(` to the temperature initial condition in the equivalent `mantle_convection` demo and fixed a lengthy line in the `thermochemical buoyancy` demo.

Updated animation:

https://github.com/user-attachments/assets/9a3e0257-39a7-45c4-8192-8c66357c827c

w3168 (approved):
Thanks @Patol75 this looks very clear and easy to follow. I have a few comments that don't really need addressing and a few small suggestions. 

Happy to merge once you've fixed them.

Do you need a corresponding PR on the website (@angus-g )?

Patol75:
Thank you, @w3168. I will address your comments tomorrow.

Here is the updated animation:

https://github.com/user-attachments/assets/f614eb18-5e34-4b3e-8ad1-fc3efa95e58e

w3168:
Thanks @Patol75 the new changes with the dense compositional layer looks good to me too. Happy for this to be merged once you fix the lint and the small changes above.

demos/multi_material/2d_cylindrical/2d_cylindrical.py
```diff
@@ -0,0 +1,311 @@
...
+# + tags=["active-ipynb"]
+# import pyvista as pv
+# -
+
+# We next set up the mesh and function spaces and define functions to hold our solutions
+# and the simulation's time step.
+
+# +
+# Annulus radii (e.g. [van Keken (2001)](https://doi.org/10.1016/S0031-9201(01)00195-9))
+rmin, rmax = 1.2083, 2.2083
```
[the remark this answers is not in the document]
Patol75: Yes, it is intended. The original values are not optimal, and you can see that they have also been changed in [2d_cylindrical_TALA_DG](https://github.com/g-adopt/g-adopt/blob/main/tests/2d_cylindrical_TALA_DG/2d_cylindrical_TALA_DG.py). I do not think it is a big deal, but happy to discuss indeed. I believe the values have remained what they are in [2d_cylindrical](https://github.com/g-adopt/g-adopt/blob/main/demos/mantle_convection/2d_cylindrical/2d_cylindrical.py) to keep the consistency with [Davies et al. (2022)](https://gmd.copernicus.org/articles/15/5127/2022/).

PR #324, opened 2025-10-28
adding Irksome integration to gadopt

sghelichkhani:
This PR replaces gadopt's Runge-Kutta time steppers with Irksome, addressing limitations with nonlinear mass terms.

## Problem
G-ADOPT time steppers fail with nonlinear mass terms like $C(h) dh/dt$ in Richards equation. SNES diverges when mass coefficients depend on current solution, requiring e.g.,  lagged terms $(C(h_old) dh/dt)$ for convergence.

## Solution
- a `IrksomeIntegrator` adapter maintaining G-ADOPT API
- Implemented tableau mapping from gadopt to Irksome schemes

My main focus here was backward compatibility


stephankramer:
Wow, this is impressive, replacing all that without needing to change any of the outside code. Of course we can discuss if we can simplify the outside code as well, but this may be a good way to transition (and ensuring nothing breaks).

Without having looked at it for too long, I can see some potential simplifications. The `gadopt_to_irksome_tableau` seems a bit complex. I think you can replace it, and the definitions of ERKGeneric and DIRKGeneric with:
```
class RKGeneric(IrksomeIntegrator):
    stage_type = "deriv"
    def __init__(
        self,
        equation: Equation,
        solution: firedrake.Function,
        dt: float,
        solution_old: Optional[firedrake.Function] = None,
        solver_parameters: Optional[dict[str, Any]] = {},
        strong_bcs: Optional[list[firedrake.DirichletBC]] = None,
    ):
        butcher = create_custom_tableau(self.a, self.b, self.c)
        super().__init__(
            equation=equation,
            solution=solution,
            dt=dt,
            butcher=butcher,
            stage_type=self.stage_type,
            strong_bcs=strong_bcs,
            solver_parameters=solver_parameters
        )

class ERKGeneric(RKGeneric)
    """Generic explicit Runge-Kutta time integrator using Irksome."""
    stage_type = "explicit"


class DIRKGeneric(RKGeneric):
    """Generic diagonally implicit Runge-Kutta time integrator using Irksome."""
    stage_type = "dirk"
```
i.e. just generate the butcher tableau from the a,b,c class attributes directly and the only difference between the two then is the stage_type which you can also specify as a class attribute.  Note that this is dropping (as in your code) the `solution_old` argument, so we need to think about that.
I know this doesn't handle the midpoint rule and backward euler, but you could do a a similar abstract class that simply picks up the butcher tableau from a class attribute and then
```
   class BackwardEulerAbstract(GenericIntegratorFromStaticButcherTableau):
       butcher = IrksomeBackwardEuler()
       stage_type = "dirk"
```
etc. This would also handle a number of the IRKSomeXXX classes that you define at the end (and involve quite a bit of boiler plate) - the ones that don't require any arguments to create the butcher tableau. For the ones that require an argument, we already need to think about how that actually works: at the moment we're passing a timestepper class into the solver and then solver creates the timestepper object, so that requires it to always have the same arguments - we could introduce something like timestepper_kwargs, but then if the user is clever enough to fidle with that we could also allow them to just provide the butcher tableau that they create themselves maybe?

gadopt/time_stepper.py
```diff
@@ -117,120 +124,189 @@ def advance(
...
+
+        # Create MeshConstant objects for time variables (what Irksome expects)
+        mesh = solution.ufl_domain()
+        mc = MeshConstant(mesh)
+        self.t = mc.Constant(0.0)
+        self.dt_mesh_const = mc.Constant(float(dt))  # Internal MeshConstant for Irksome
+
+        # Build the Irksome form: F = equation.mass(Dt(solution)) - equation.residual(solution)
+        # Note: equation.residual returns the negative of the left-hand contribution
+        F = equation.mass(Dt(solution)) - equation.residual(solution)
```
stephankramer: Presumably this doesn't yet address the issue that started this - this still won't work if your mass term is of the form $$C(u)\frac{\partial u}{\partial t}$$. I think (but this does require changes on the outside) if we indeed have successfully replaced all timesteppers with IRKSome indeed, the mass term could be just like any other term where the _equation_ code would be responsible for inserting the `Dt` operator. Then the equation would just produce the entire residual, including the mass term

gadopt/time_stepper.py
```diff
@@ -117,120 +124,209 @@ def advance(
...
+        stage_type: str = "deriv",
         solution_old: Optional[firedrake.Function] = None,
-        solver_parameters: Optional[dict[str, Any]] = {},
         strong_bcs: Optional[list[firedrake.DirichletBC]] = None,
+        bc_type: str = "DAE",
+        solver_parameters: Optional[dict[str, Any]] = None,
     ):
-        super(ERKGeneric, self).__init__(
-            equation, solution, dt, solution_old, solver_parameters, strong_bcs
+        super().__init__()
```
Patol75: This statement should not be needed because `TimeIntegratorBase` does not define `__init__`.

gadopt/time_stepper.py
```diff
@@ -117,120 +124,209 @@ def advance(
...
+        dt: Integration time step (Firedrake Constant or float)
+        butcher: Irksome Butcher tableau (e.g., GaussLegendre, RadauIIA)
+        stage_type: Type of stage formulation (e.g., "deriv")
+        solution_old: Firedrake function representing the equation's solution
                       at the previous timestep
-      solver_parameters: Dictionary of solver parameters provided to PETSc
-      strong_bcs: List of Firedrake Dirichlet boundary conditions
-
+        strong_bcs: List of Firedrake Dirichlet boundary conditions
+        bc_type: Boundary condition type for Irksome ("DAE" or "ODE")
```
Patol75: Do we support `EquationBC` as part of `strong_bcs`? If yes, best to specify that such a condition is only compatible with `bc_type="DAE"`.

gadopt/time_stepper.py
```diff
@@ -117,120 +124,209 @@ def advance(
...
-                solver = firedrake.LinearVariationalSolver(prob, options_prefix=self.name + '_k{:}'.format(i),
-                                                           solver_parameters=self.solver_parameters)
-                self.solver.append(solver)
-
-    def initialize(self, solution):
-        self.solution_old.assign(solution)
+    def initialize(self, init_solution):
+        """Initialize the time integrator."""
+        self.solution.assign(init_solution)
         self._initialized = True
```
Patol75: We should remove any initialisation happening outside class instantiation, like we did for `stokes_integrators.py`.

gadopt/time_stepper.py
```diff
@@ -117,120 +124,209 @@ def advance(
...
+    # Define scheme mappings
+    scheme_mappings = {
+        # Special cases with direct Irksome equivalents
+        BackwardEulerAbstract: lambda: (IrksomeBackwardEuler(), "dirk"),
+        ImplicitMidpointAbstract: lambda: (GaussLegendre(1), "dirk"),
+
+        # Explicit schemes
+        ForwardEulerAbstract: lambda: (create_custom_tableau(temp_scheme.a, temp_scheme.b, temp_scheme.c), "explicit"),
+        SSPRK33Abstract: lambda: (create_custom_tableau(temp_scheme.a, temp_scheme.b, temp_scheme.c), "explicit"),
+        ERKMidpointAbstract: lambda: (create_custom_tableau(temp_scheme.a, temp_scheme.b, temp_scheme.c), "explicit"),
```
Patol75: Are we missing `eSSPRKs3p3Abstract` and `eSSPRKs10p3Abstract` here for level-set purposes?

gadopt/time_stepper.py
```diff
@@ -117,120 +124,209 @@ def advance(
...
+    tableau, stage_type = gadopt_to_irksome_tableau(scheme_class)
+
+    return IrksomeIntegrator(
+        equation=equation,
+        solution=solution,
+        dt=dt,
+        butcher=tableau,
+        stage_type=stage_type,
+        **kwargs
+    )
```
Patol75: Is this function used somewhere?

gadopt/time_stepper.py
```diff
@@ -117,120 +124,189 @@ def advance(
...
+        if not self._initialized:
+            self.initialize(self.solution)
+
+        # Sync dt_mesh_const with the reference dt before advancing
+        self.dt_mesh_const.assign(float(self.dt_reference))
 
-        Tendencies must have been evaluated first.
+        # Update forcings if provided
+        if update_forcings is not None:
+            update_forcings(float(self.t))
```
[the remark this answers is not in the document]
Patol75: Actually, we do use it to update the level-set gradient in the level-set reinitialisation equation. I note that the approach in #157 removes this need.

sghelichkhani:
There are weirdly three tests that are failing 😒 . 
- [ ] `test_benchmark[multi_material/compositional_buoyancy]` in `demos/test_all.py`, with `u_rms: got 206.34235433153748, expected 205.305458067097`
- [ ] `test_scalar_advection_diffusion[integrated_q]` in `tests/scalar_advection_diffusion/test_scalar_advection_diffusion.py` Got: `array([0.18807727])` Expected: `0.18701895`
- [ ]    And the other one in `test_benchmark[multi_material/thermochemical_buoyancy]`; `[left]:  [206.34235433153748, 31.999999999999822]`,  `[right]: [205.305458067097, 31.99999999999993]`

Some of the differences are very small. But still weird. @Patol75 Any thoughts?

gadopt/time_stepper.py
```diff
@@ -377,16 +326,18 @@ def __init__(
         solver_parameters: Optional[dict[str, Any]] = {},
         strong_bcs: Optional[list[firedrake.DirichletBC]] = None,
     ):
-        # Get the Butcher tableau and stage type for this scheme
-        butcher, stage_type = gadopt_to_irksome_tableau(self.__class__.__bases__[1])
+        if self.butcher_tableau is None:
+            raise NotImplementedError(
+                f"{self.__class__.__name__} must define a butcher_tableau class attribute"
+            )
```
Patol75: Given [Python docs](https://docs.python.org/3/library/exceptions.html#NotImplementedError), you should probably raise `ValueError` here instead. Or you could make `butcher_tableau` an abstract method.

gadopt/time_stepper.py
```diff
@@ -227,16 +240,61 @@ def advance(self, update_forcings: Callable | None = None, t: float | None = Non
...
+                    g.assign(sum((ks[j] * (self.stepper.AA[i, j] * dt) for j in range(i)), u0))
+
+                    # Update BC constants for the variational problem
+                    self.stepper.update_bc_constants(i, c)
+                    a.assign(self.stepper.AA[i, i])
+
+                    # Call update_forcings before solving this stage (if provided)
+                    if update_forcings is not None:
+                        # Calculate stage time: t + c[i] * dt
+                        stage_time = float(self.t) + float(c) * float(dt)
```
Patol75: Not that it matters for the level-set issue, but why isn't this `c[i]`?

gadopt/time_stepper.py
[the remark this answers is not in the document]
sghelichkhani: See this: https://github.com/g-adopt/g-adopt/discussions/338

gadopt/time_stepper.py
```diff
@@ -117,120 +124,209 @@ def advance(
...
-    def solve_tendency(self, i_stage, update_forcings, t) -> None:
-        """Evaluates the tendency of i-th stage"""
-        if self._nontrivial:
-            if update_forcings is not None and t is not None:
-                update_forcings(t + self.c[i_stage] * self.dt)
-            elif update_forcings is not None:
-                update_forcings()
+        # Update time if provided
+        if t is not None:
+            self.t.assign(float(t))
```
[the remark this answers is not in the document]
sghelichkhani: I have done the `dt` change. About `t`: as usual, not keen to do an API change unless necessary. I don't think this is necessary.

gadopt/time_stepper.py
sghelichkhani: Dropped it now!

Patol75 (commented):
Thank you for adding a test demonstrating the Irksome adaptive timestepping! If you are happy to remove `update_forcings` entirely, I will give the whole PR another look right after.

gadopt/equations.py
```diff
@@ -128,6 +128,47 @@ def residual(
...
+        # Apply Dt to the solution for the mass term
+        mass_form = self.mass(dt_operator(solution))
+
+        # Get the spatial residual terms
+        residual_form = self.residual(solution)
+
+        # Combine: mass term minus residual (which is already negative)
+        # This gives us: Dt(u) + spatial_terms = 0 (in weak form)
+        return mass_form - residual_form
+
```
Patol75: Is this function necessary? Can we not just write `F = equation.mass(Dt(solution)) - equation.residual(solution)` in line 242 of `time_stepper.py`?

tests/scalar_advection/generate_expected.py
```diff
@@ -1,4 +1,17 @@
...
+
+# Generate expected values for adaptive test
+try:
+    final_error_adaptive = np.loadtxt("final_error_adaptive.log")
+    num_steps_adaptive = np.loadtxt("num_steps_adaptive.log")
+    dt_stats_adaptive = np.loadtxt("dt_stats_adaptive.log")
+
+    np.save("expected_error_adaptive.npy", [final_error_adaptive])
+    np.save("expected_steps_adaptive.npy", [num_steps_adaptive])
+    np.save("expected_dt_stats_adaptive.npy", dt_stats_adaptive)
```
Patol75: How about using [`savez`](https://numpy.org/doc/stable/reference/generated/numpy.savez.html) (or `savez_compressed` if relevant) to generate a single file?

tests/scalar_advection/Makefile
```diff
@@ -2,15 +2,20 @@ include ../../rules.mk
...
 	$(run-python)
 
+final_error_adaptive.log num_steps_adaptive.log dt_stats_adaptive.log : category := tests
+final_error_adaptive.log num_steps_adaptive.log dt_stats_adaptive.log: scalar_advection_adaptive.py
+	$(run-python)
+
 clean:
-	rm -f *.pvd *.vtu *.pvtu *.h5 final_error.log
+	rm -f *.pvd *.vtu *.pvtu *.h5 final_error.log final_error_adaptive.log
+	rm -f num_steps_adaptive.log dt_stats_adaptive.log
```
Patol75: Any reason not to use `*.log`?

tests/scalar_advection/scalar_advection_adaptive.py
```diff
@@ -0,0 +1,134 @@
+# Demo for pure scalar advection with adaptive timestepping - this is adapted from the Firedrake DG_advection demo,
+# but here we use G-ADOPT's GenericTransportSolver with IrksomeRadauIIA adaptive timestepping and use a CG discretisation with
```
Patol75: How about putting a full stop at the end of the first line and then starting the second line with `Here, however`?

tests/unit/test_irksome_time_steppers.py
```diff
@@ -0,0 +1,736 @@
...
+            residual_terms=[diffusion_term, source_term],
+            mass_term=mass_term,
+            eq_attrs=eq_attrs,
+            bcs={}
+        )
+
+        # Test different orders
+        for order in [1, 2, 3]:
+            integrator = IrksomeGaussLegendre(equation, u, dt=0.01, order=order)
+            assert integrator is not None
```
Patol75: Are these two tests the same? If yes, can we use `parametrize`?

gadopt/time_stepper.py
```diff
@@ -96,140 +102,290 @@ def get_final_solution(self):
...
+        solver_parameters: Optional[dict[str, Any]] = None,
+        initial_time: Number = 0.0,
+        adaptive_parameters: Optional[dict[str, Any]] = None,
+        **irksome_kwargs,
     ):
-        super(ERKGeneric, self).__init__(
-            equation, solution, dt, solution_old, solver_parameters, strong_bcs
+        self.equation = equation
+        self.solution = solution
+        self.solution_old = solution_old or firedrake.Function(solution)
```
Patol75: `self.solution_old = solution_old or firedrake.Function(solution, name=solution.name() + " (old)")`

gadopt/time_stepper.py
```diff
@@ -96,140 +102,290 @@ def get_final_solution(self):
...
-      strong_bcs: List of Firedrake Dirichlet boundary conditions
+            # Sync dt_mesh_const back to dt_reference
+            # (Irksome updated dt_mesh_const internally during advance)
+            self.dt_reference.assign(float(adapt_dt))
+
+            # Return tuple so users can track the actual dt used
+            return (adapt_error, float(adapt_dt))
 
+        # Non-adaptive: return None for consistency
+        return None
```
Patol75: `return`

gadopt/time_stepper.py
```diff
@@ -96,140 +102,290 @@ def get_final_solution(self):
...
+
+        # Keep reference to original dt constant for syncing
+        self.dt_reference = ensure_constant(dt)
+
+        # Create MeshConstant objects for time variables (what Irksome expects)
+        # These are shared with Irksome's TimeStepper (ensures synchronisation)
+        mesh = solution.ufl_domain()
+        mc = MeshConstant(mesh)
+        self.t = mc.Constant(initial_time)  # Shared time variable with Irksome
+        self.dt_mesh_const = mc.Constant(float(dt))  # MeshConstant for Irksome, synced from dt_reference
```
Patol75: What do you think of
```python
self.dt_user = ensure_constant(dt)
[...]
self.dt = mc.Constant(dt)
```
?

tests/unit/test_irksome_time_steppers.py
```diff
@@ -0,0 +1,736 @@
...
+        DIRKLPUM2Abstract: lambda: (create_custom_tableau_local(temp_scheme.a, temp_scheme.b, temp_scheme.c), "dirk"),
+    }
+
+    # Get the mapping function or use default
+    mapping_func = scheme_mappings.get(
+        scheme_class,
+        lambda: (create_custom_tableau_local(temp_scheme.a, temp_scheme.b, temp_scheme.c), "deriv")
+    )
+
+    return mapping_func()
```
Patol75: You could replace this block with

```python
import pytest
from firedrake import *
from irksome import BackwardEuler as IrksomeBackwardEuler
from irksome import GaussLegendre

from gadopt import *
from gadopt.scalar_equation import diffusion_term, mass_term, source_term
from gadopt.time_stepper import *

# Define scheme mappings
scheme_mappings = {
    # Special cases with direct Irksome equivalents
    BackwardEulerAbstract: (IrksomeBackwardEuler(), "dirk"),
    ImplicitMidpointAbstract: (GaussLegendre(1), "dirk"),
    # Explicit schemes
    ForwardEulerAbstract: "explicit",
    SSPRK33Abstract: "explicit",
    ERKMidpointAbstract: "explicit",
    ERKLSPUM2Abstract: "explicit",
    ERKLPUM2Abstract: "explicit",
    eSSPRKs3p3Abstract: "explicit",
    eSSPRKs4p3Abstract: "explicit",
    eSSPRKs5p3Abstract: "explicit",
    eSSPRKs6p3Abstract: "explicit",
    eSSPRKs7p3Abstract: "explicit",
    eSSPRKs8p3Abstract: "explicit",
    eSSPRKs9p3Abstract: "explicit",
    eSSPRKs10p3Abstract: "explicit",
    # DIRK schemes
    CrankNicolsonAbstract: "dirk",
    DIRK22Abstract: "dirk",
    DIRK23Abstract: "dirk",
    DIRK33Abstract: "dirk",
    DIRK43Abstract: "dirk",
    DIRKLSPUM2Abstract: "dirk",
    DIRKLPUM2Abstract: "dirk",
}


# Utility functions for testing (moved from time_stepper.py)
def gadopt_to_irksome_tableau(scheme_class):
    """Convert a G-ADOPT scheme class to an Irksome Butcher tableau.

    Args:
        scheme_class: A G-ADOPT AbstractRKScheme class

    Returns:
        tuple: (ButcherTableau instance, stage_type string)
    """
    # Get the mapping function or use default
    mapping = scheme_mappings[scheme_class]
    if isinstance(mapping, str):
        return create_custom_tableau(
            scheme_class.a, scheme_class.b, scheme_class.c
        ), mapping
    else:
        return mapping
```

Then, for `test_tableau_conversion`, you could use `@pytest.mark.parametrize("scheme_class", scheme_mappings.keys())`. Finally, `test_invalid_scheme` would need to expect `KeyError` instead of `(AttributeError, TypeError)`.

Patol75 (changes requested):
A few suggestions to prevent some code duplication, among other things.

tests/scalar_advection/generate_expected.py
sghelichkhani: Done! Thanks

tests/scalar_advection/scalar_advection_adaptive.py
```diff
@@ -0,0 +1,134 @@
...
+# We use IrksomeRadauIIA which has an embedded error estimator for adaptive control.
+# We only include an advection term and apply weak boundary conditions on inflow regions.
+bc_in = {"q": q_in}
+bcs = {1: bc_in, 2: bc_in, 3: bc_in, 4: bc_in}
+eq_attrs = {"u": u}
+adv_solver = GenericTransportSolver(
+    "advection", q, dt, IrksomeRadauIIA,
+    eq_attrs=eq_attrs, bcs=bcs, su_advection=True,
+    timestepper_kwargs={
+        'order': 3,  # RadauIIA order
```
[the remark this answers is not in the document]
sghelichkhani: Yes! Some of these methods have special arguments, in this case you could set the order of the method. It's a whole new world TBH. Just trying to be compatible with fully implicit RK methods.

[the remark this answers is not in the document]
sghelichkhani: Not sure if you are satisfied. If so resolve this yourself @Patol75

tests/scalar_advection/test_scalar_advection.py
```diff
@@ -11,3 +11,41 @@ def test_scalar_advection():
...
+    # Load actual results
+    final_error = np.loadtxt(base / "final_error_adaptive.log")
+    num_steps = np.loadtxt(base / "num_steps_adaptive.log")
+    dt_stats = np.loadtxt(base / "dt_stats_adaptive.log")  # [min, max, mean]
+
+    # Check final error value
+    if expected_error_file.exists():
+        expected_error = np.load(expected_error_file)
+        assert np.allclose(final_error, expected_error, rtol=1e-6, atol=1e-16), \
+            f"Final error mismatch: got {final_error}, expected {expected_error}"
```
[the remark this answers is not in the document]
sghelichkhani: Done

gadopt/time_stepper.py
sghelichkhani: Am I understanding correctly that you are discussing naming? If so, my intention here with naming was to say that `self.dt_reference` is what we need to synch with inside `dt`. `dt_user` sounds like a preference. 

If you are convinced, resolve this yourself.

gadopt/time_stepper.py
```diff
@@ -96,140 +102,290 @@ def get_final_solution(self):
...
+        """
+        return self.t
+
+    def get_dt(self) -> float:
+        """Get the current value of the time step from dt_reference.
+
+        Returns:
+            The current time step.
+        """
+        return self.dt_reference
```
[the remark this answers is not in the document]
sghelichkhani: I don't know. My initial thinking was to provide that option. While I see your point. I also don't see any harm in having those methods for the future. 

Happy to remove them if you think it's messy.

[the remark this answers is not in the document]
Patol75: @dsroberts Could we have your input here? Would you recommend keeping or removing these methods?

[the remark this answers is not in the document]
dsroberts: Yes, of course. I meant, "Do `t` and `dt_reference` ever need to be modified by a user after being initialised?". I hadn't looked at the code properly before writing that, I'm doing a proper review now.

tests/unit/test_irksome_time_steppers.py
sghelichkhani: Done! Make sure it is what you wanted and resolve this.

gadopt/time_stepper.py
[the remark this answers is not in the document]
sghelichkhani: Thanks @dsroberts @Patol75. To clarify as you said, both `t` and `dt` change. There is a bit of book keeping to make sure if one defined by the user changes Irksome internal time is also changed. 

I would say let's keep it. No harm in it.

dsroberts:
I was part way through a review, but I have to go offline now. I was thinking along the lines of unifying the functionality between `TimeIntegrator` and `IrksomeIntegrator`. Its kind of odd that adaptive timestepping is sometimes controlled by the time integrator and sometimes controlled externally. I think it'd be good if we can look at how the adaptive functionality in `utility.py` can be made to work more seamlessly with the time integrators. Maybe we can reduce it to a single function that can be passed to a time integrator and have that handle adaptivity rather than leaving it to a user? Maybe `TimestepAdapter` can be a base class, the existing `TimestepAdapter` can be an implementation of it that `TimeIntegrator` can initialise and handle updating? Perhaps we provide a function that a user can call to update `dt` safely, like a setter that does the `.assign(dt)` and/or any bookkeeping necessary to keep time steps in sync for you. I think I need to spend some time actually using this functionality to really understand how it all fits together.

Second, I think there is quite a bit of repeated code between `TimeIntegrator.__init__` and `IrksomeIntegrator.__init__`. Maybe the use of `TimeIntegratorBase` as an abstract base class should be reviewed, especially if we're doing away with `initialize`. It would be a good place to put all the shared book keeping between `TimeIntegrator.__init__` and `IrksomeIntegrator.__init__`.

As a first pass I didn't have many specific comments around the code itself, just a nit picks around the type annotation and the stuff we discussed in the dev meeting.

tests/unit/test_irksome_time_steppers.py
Patol75: Thank you! Any reason not to remove
```python
import numpy as np
from irksome.ButcherTableaux import ButcherTableau


# Helper function to create custom tableau
def create_custom_tableau(a, b, c):
    """Create a Butcher tableau for Irksome.

    Args:
        a: Butcher matrix (2D array)
        b: Weights (1D array)
        c: Nodes (1D array)

    Returns:
        ButcherTableau instance
    """
    return ButcherTableau(
        A=np.array(a),
        b=np.array(b),
        btilde=None,
        c=np.array(c),
        order=len(b),
        embedded_order=None,
        gamma0=None
    )

```
?

`create_custom_tableau` is already imported as part of `from gadopt.time_stepper import *`.

sghelichkhani:
@patol75 Thanks so much  for digging into this and for doing a thorough pass on the time-stepping side of things. The fork was helpful.
I’ve gone through your changes in detail and have pulled in the bits that fit well with the current design, including specifying that you are the co-author of this (as always, I don't want just-for-formatting changes to pollute actual changes), in particular:
- [x] the simplified RK class hierarchy (no separate *Abstract/concrete pairing),
- [x] the removal of the old TimeIntegrator* base classes,
- [x] the cleaned-up Irksome wrappers and exports in time_stepper.py / __init__.py,
- [x] and the transport solver improvements (explicit `delta_t`, solver parameter helpers, callback hook, etc.).

On the time management / time_objects restructuring, I’m less convinced for now: it would fundamentally change how we expose and control time-stepping in g-adopt (user-controlled dt constants, adaptive feedback, existing demos/tests), and I think that’s beyond the scope of this PR. Even if we decided we like that direction, it really wants a separate design discussion and a dedicated follow-up PR so we can reason properly about API breakage and adaptive behaviour.

So my proposal is:
merge this PR with the “safe” structural and API-cleanup improvements from your fork (which are now in this branch),
then, if you’re still keen,let's discuss the time-management on the dev meeting first and then discuss this.

gadopt/time_stepper.py
sghelichkhani: I see your point. Just to emphasise the issue with bookkeeping of dt I keep it as `dt_irksome`, and added a bunch of comments here and there to make that clear.

Patol75:
I understand the difference between `coef * Dt(solution)` and `Dt(coef * solution)` and why it requires attention. However, I fail to see how 2bc7c28 addresses this scenario. It seems to me the changes do not change anything. Could you provide an example of this special case you mention? Could we define an argument for `IrksomeIntegrator` that represents this time-dependent coefficient (with a default value of `1.0`) and have `F = equation.mass(Dt(coefficient * solution)) - equation.residual(solution)`?

Patol75:
Sounds good to me! I just implemented it on my [fork](https://github.com/g-adopt/g-adopt/commit/9badedc3e82d9f9441684be3b08704a714f77e7c).
So whenever the CI is green here, I am in favour of merging, and then I will open a number of follow-up PRs:
- streamlining `time_stepper.py`
- removing `mass_term`
- proposing a new time and time-step interface aligned with Irksome
- using Irksome to solve free surface

tests/scalar_advection/scalar_advection_adaptive.py
```diff
@@ -0,0 +1,133 @@
...
+nubar_outfile.write(nubar)
+
+# Here is the time stepping loop with adaptive timestepping, with an output every 20 steps.
+# The timestep dt will be automatically adjusted by the adaptive stepper based on error estimates.
+t = 0.0
+step = 0
+dt_values = []  # Store all timestep values for testing
+while t < T:
+    # Set maximum dt to prevent overshooting final time
+    adv_solver.ts.stepper.dt_max = T - t
```
Patol75:
```suggestion
    if T - t < float(adv_solver.ts.dt_reference):
        # Set maximum dt to prevent overshooting final time
        adv_solver.ts.stepper.dt_max = T - t
```

Patol75 (changes requested):
Sorry, just finally learned how to use suggestions (whoops). So I am assuming the CI is likely to fail because, now that the maximum time step is updated at every loop iteration, the solution will change. I recommend the two changes below, with an update to the test expected values, such that the final state is comparable to the non-adaptive case.

Issue #232, opened 2025-06-05
Verify level-set implementation in cylindrical and spherical domains


Patol75:
The level-set implementation has only been tested using rectangular domains. Behaviour in cylindrical and spherical domains is currently considered unknown.

sghelichkhani:
If this is the case, can we please have a sanity check for when `mesh.cartesian` is `False` in https://github.com/g-adopt/g-adopt/pull/187, before we are sure things work in non cartesian domains?

Patol75:
That was my plan, too. But watch the 3-D issue update coming up.

PR #330, opened 2025-11-03
Replace Tosi et al. (2015) by Davies et al. (2022)

Patol75:
This PR stands as a preparatory step to #321. I propose mainly two changes here, described below:
- The current `2d_cylindrical.py` demo in `mantle_convection` takes too much time to run for no good reason. On `main`, CI reports times over 15 minutes for this demo (e.g. [latest main CI](https://github.com/g-adopt/g-adopt/actions/runs/18963495139)). Given that our CI runs will lengthen as new demos and tests are incorporated, I think it is good practice to decrease, where possible, the footprint of existing demos and tests. As a result, I suggest decreasing `steady_state_tolerance` from `1e-7` to `1e-4`, which requires adding a condition on a minimum simulation time required for steady state to be accepted (I chose 100 times the initial time step). Expected demo diagnostics vary minimally, as can be seen [here](https://github.com/g-adopt/g-adopt/actions/runs/18996341652/job/54256734763#step:15:140), and the reported time is now [7 minutes](https://github.com/g-adopt/g-adopt/actions/runs/18996341652).
- Level-set capabilities need to be benchmarked in non-Cartesian geometries. One simple way to achieve this is by incorporating the setup of the 2-D cylindrical demo (Davies et al., 2022) as a multi-material test with a passive material interface using `run_benchmark.py` (as was discussed in #232). The idea is that such a test case should reproduce the steady state obtained in the demo. Level-set testing already includes such a passive test, namely a benchmark case from Tosi et al. (2015). Both cases are thermochemical benchmarks and would serve the same purpose: validating that level-set incorporation does not alter the steady state of such a benchmark. However, the case from Davies et al. (2022) adds the non-Cartesian dimension, which makes it more valuable. As a result, I suggest removing the Tosi et al. (2015) case (which also takes an awful long time on the CI) and replacing it with the 2-D cylindrical benchmark of Davies et al. (2022). I make the necessary changes in `tests/multi_material` to allow for this change. In terms of validating the results from the passive level-set approach, I calculate the same diagnostics as in the demo and compare their values to the updated ones included in this PR. Relative differences are on the order of 1e-3, which could stem from the fact that `run_benchmark.py` employs a discontinuous temperature function space. I also note that I made use of subcycling (which was not tested before) and had to incorporate the changeset from #327 to avoid the exception described in #320.


drhodrid:
Just a quick one from me -- we can't do point 1 as it is. I will explain why when I have more time later in the week.

drhodrid:
Just to clarify why we can’t loosen that steady‐state tolerance (point 1 above) to have it documented. 

We actually need this test to converge to a tight tolerance, even if that means it takes a bit longer to run. The reason is that in earlier versions of G-ADOPT, the nullspace treatment and that of boundary conditions were sometimes problematic. That often only becomes visible when we try to reach very small steady‐state residuals: looser tolerances will hide the issue entirely.

I set the current value deliberately so that any regression or re-emergence of that inconsistency would cause the test to fail. It’s therefore a diagnostic safeguard, not just a convergence criterion. If we reduce the tolerance, we’d lose the ability to detect that class of bug.

So unfortunately, even though it lengthens CI time a bit, we need to keep the current setting.

PR #331, opened 2025-11-03
Add multi-material demo with free surface based on Schmeling et al. (2008)

Patol75:
This PR adds another multi-material demo that focuses on the interaction between compositional buoyancy and a free surface under the Boussinesq approximation. Although the evolution is sensible and reproduces published results, I am concerned by the fact that, in order to reproduce these results, I had to specify a material interface at the top boundary of the lithosphere. If I do not do that, then subduction becomes sluggish. My interpretation is that imposing a material interface at the top boundary decreases the viscosity there because the mantle has a lower viscosity than the lithosphere, and this viscosity decrease is needed to enable subduction.

Animation of level-set and free-surface evolutions:

https://github.com/user-attachments/assets/44b610a5-ac5a-43d8-a819-078b6d274aac


drhodrid:
I'd like @stephankramer's thoughts on this surface interface / BC point. 

In Fluidity, we typically included a weak boundary condition at the material interface (i.e. top boundary), such that the material coming in above the slab (when the free surface moved up or down) was mantle. Is what Thomas is doing here somehow equivalent or would you recommend a different approach?

demos/multi_material/free_surface/free_surface.py
```diff
@@ -0,0 +1,315 @@
...
+# We set the Rayleigh number to 0.0 to indicate explicitly the absence of thermal
+# buoyancy and the compositional Rayleigh number to 1.0 to enable compositional
+# buoyancy.
+approximation = BoussinesqApproximation(
+    Ra=0.0, RaB=1.0, delta_rho=rho_material - rho_mantle, g=9.81, mu=mu
+)
+# -
+
+# Let us now verify that the material fields have been correctly initialised. We plot
+# the density and vicosity across the domain.
```
w3168: `viscosity`

w3168 (approved):
Hi @Patol75 thanks for the nice demo. I spotted one typo but otherwise looks good. 

I played around a bit with the top material interface. As you say, moving the extra node in the top part of the level set interface to the boundary coordinates leads to much more sluggish flow. I wondered if we needed a `u_h \cdot \nabla_h \eta` term, i.e. horizontal advection of free surface gradient because the slope near the subduction zone is order 1. But adding this in didn't seem to massively change the results. 

I imagine for now the best solution is to get this in so it is CI tested and if/when more subduction zone modelling happens in g-adopt this will be a good starting point.

PR #332, opened 2025-11-04
DO NOT SQUASH merge release into main

dsroberts:


demos/multi_material/thermochemical_buoyancy/thermochemical_buoyancy.py
Patol75: Not sure if I am misunderstanding, but the `-` between `level` and `set` was removed in https://github.com/g-adopt/g-adopt/pull/320/files. It seems it would remain here.

Patol75: Ah, that is from a35b1be. Not the end of the world.

angus-g:
I did this via a manual `git merge --ff` so we didn't get a double merge commit, or end up rebasing or anything weird through the GitHub UI.

PR #333, opened 2025-11-07
Warn user when `constant_jacobian=True`

Patol75:
It has happened a few times that simulations are run with `constant_jacobian=True` given to `StokesSolver` when the Jacobian was not actually constant (e.g. when the viscosity varies in time). This PR proposes to raise a warning when this argument is supplied, such that a user can ascertain whether or not they meant to include it.

### DIFF
--- a/tests/adjoint/taylor_test.py
+++ b/tests/adjoint/taylor_test.py
@@ -3,16 +3,14 @@ This standalone script tests the robustness of the derivatives
 using the Taylor remainder convergence test.
 """
 
-import sys
-import warnings
-from pathlib import Path
-
-import numpy as np
-from cases import cases, schedules
-from mpi4py import MPI
-
 from gadopt import *
 from gadopt.inverse import *
+from mpi4py import MPI
+import numpy as np
+import sys
+from pathlib import Path
+
+from cases import cases, schedules
 
 ds_t = ds_t(degree=6)
 dx = dx(degree=6)
@@ -102,18 +100,15 @@ def rectangle_taylor_test(case, scheduler_name):
 
     # Setup Energy and Stokes solver
     energy_solver = EnergySolver(T, u, approximation, delta_t, ImplicitMidpoint, bcs=temp_bcs)
-    with warnings.catch_warnings():  # Suppress constant Jacobian warning
-        warnings.simplefilter("ignore")
-
-        stokes_solver = StokesSolver(
-            z,
-            approximation,
-            T,
-            bcs=stokes_bcs,
-            constant_jacobian=True,
-            nullspace=Z_nullspace,
-            transpose_nullspace=Z_nullspace,
-        )
+    stokes_solver = StokesSolver(
+        z,
+        approximation,
+        T,
+        bcs=stokes_bcs,
+        constant_jacobian=True,
+        nullspace=Z_nullspace,
+        transpose_nullspace=Z_nullspace,
+    )
 
     initial_timestep = 0 if case in ["Tobs", "uobs"] else timesteps - 1
 


Patol75:
Highlighting here thoughts from yesterday's developer meeting. While a warning message can be useful, it can also be annoying if a user knows what he or she is doing. Moreover, several CI-tested demos and tests would run into this warning. There should be a better way to handle this situation

sghelichkhani:
I fully agree. I am not sure having a warning for a flag that the user needs to explicitly set is going to prevent what this PR wants to prevent.

drhodrid:
It would definitely prevent the issue we've had previously (i.e. running variable viscosity cases with a non varying viscosity) -- I'm not particularly tied to this way of doing it, but we need something along similar lines.

stephankramer:
> If I understand this issue correctly constant_jacobian=True can only be used if the stokes system has no explicit time dependence.

To be precise, if the *Jacobian* of the Stokes system has no time dependence. Note that we've already assumed the Stokes system is linear, so this comes down to there are no terms that are linear in the Stokes solution variables u, p that also depend on other variables which change in time. So _most of the times_ forcing and boundary value terms are not relevant. The typical culprit is the viscosity which may depend on temperature or material distribution. The way we set things up, this is entirely bespoke and orthogonal to choices in the approximation and so it is really only the user that knows whether it is time-dependent or not. There are also all sorts of other less frequent cases, e.g. a pressure dependent buoyancy that also has some time-dependent coefficient, free surface term with variable density.

Note that this option is off by default, precisely because we can't reason for or detect this. Even for the most basic case where we have a constant (i.e. spatial constant) as the viscosity, there's no reason why the user might not be assigning a new value to it. So the option specifically is the user saying: "I know what I'm doing, trust me these coefficients are not going to change". The only trouble with that is if you start with a set up for which this is True, but then later it evolves to something where it does change, or you copy a demo that uses it and extend it. So I can see it is a bit of footgun.

I agree that we should avoid warnings for things that are completely valid things to do - but yeah, I don't really see any other solution. If it's just a single warning at the start of the simulation I think it's fine tbh. One question is how much this option really saves? Should we be setting it at all in demos? In particular because any realistic simulation in 3D on the sphere where we really care about performance, we would also be using nonlinear Stokes and this doesn't even apply.

dsroberts:
Thanks @sghelichkhani, I was thinking about this and realised that the only way to know if your Jacobian is changing in time is to assemble it and check, and the whole point of `constant_jacobian=True` is to not do that, so it really is a P=NP problem. Happy with removing it from demos and adding a warning. Thanks @stephankramer and @sghelichkhani for your insight on this one.

tests/adjoint/taylor_test.py
```diff
@@ -100,15 +102,18 @@ def rectangle_taylor_test(case, scheduler_name):
...
-        z,
-        approximation,
-        T,
-        bcs=stokes_bcs,
-        constant_jacobian=True,
-        nullspace=Z_nullspace,
-        transpose_nullspace=Z_nullspace,
-    )
+    with warnings.catch_warnings():  # Suppress constant Jacobian warning
+        warnings.simplefilter("ignore")
```
dsroberts: Why suppress the warning here? Our tests aren't user facing, so it doesn't really matter if we see a warning here. There may be other warnings we don't want to ignore coming out of here one day.

dsroberts (commented):
Seems pretty straightforward. Just wondering why you're suppressing the warning in the adjoint tests?

tests/adjoint/taylor_test.py
Patol75: Fair, I will revert that change.

dsroberts:
Hmmm... I think this may be due to the two instances of `2d_cylindrical_lvv` running at the same time and trashing the output file? I'll try running again.

angus-g (approved):

PR #336, opened 2025-11-13
Adding dependencies in Makefile demos and adding this to pages

sghelichkhani:
As title.

### DIFF
--- a/.github/workflows/render.yml
+++ b/.github/workflows/render.yml
@@ -28,6 +28,13 @@ jobs:
           apt update
           apt install -y git-lfs
       - uses: actions/checkout@v4
+        with:
+          lfs: true
+
+      - name: Fetch LFS artifacts
+        run: |
+          git lfs checkout
+
       - name: Install uv
         uses: astral-sh/setup-uv@v5
       - name: Install dependencies


demos/Makefile
```diff
@@ -22,6 +22,12 @@ $(mc_dir)/free_surface/free_surface.ipynb: $(mc_dir)/base_case/base_case.ipynb
 # dynamic topography requires the checkpoint from the base case
 $(mc_dir)/dynamic_topography/dynamic_topography.ipynb: $(mc_dir)/base_case/base_case.ipynb
 
+# smoothing requires the checkpoint file to be generated
+smoothing/smoothing.ipynb: smoothing/smoothing-example.h5
+
+smoothing/smoothing-example.h5:
+	$(MAKE) -C smoothing smoothing-example.h5
```
angus-g: Unfortunately this won't work either, you already implicitly have this rule https://github.com/g-adopt/g-adopt/blob/fbc272ae21af59ee9c46f0597dbf75d169f7280a/demos/smoothing/Makefile#L7-L13

I think it could be that we're just not doing the explicit LFS fetch... I'll push that to this branch and trigger a render to check it's working.

angus-g (changes requested):

Issue #226, opened 2025-06-04
A reproducer for the failing convergence when using schedulers


sghelichkhani:
When running the large scale adjoint with a scheduler, the taylor test fails to converge to 2.0 for longer simulations (2.0 Myrs).

PR #343, opened 2025-12-02
MONSTER: Groundwater Module

sghelichkhani:
# Monster Branch for PRs: #508 #509 #510 #511 #512

> ##  DO NOT MERGE — pending Irksome dependency on Gadi
>
> The Richards conservative time-steppers (`stage_type="value"`) depend on the conservative-update fix in [Irksome PR #226](https://github.com/firedrakeproject/Irksome/pull/226) (upstream commit `4820795` or later). The `firedrake/main-20260401` module on Gadi `/g/data/fp50` currently ships an older Irksome WITHOUT this fix, so the longtest CI will fail with `ValueError: Can't apply nonlinear operator to time derivative` until fp50 bumps its Irksome version.
>
> **Action before merge: confirm fp50 Irksome includes PR #226, then re-run the longtest CI.** No PR-side change needed once that lands.
>
> Verified locally that the chain otherwise works end-to-end: temporary PYTHONPATH override pointing at the Irksome fork made a one-tuple smoke pass (PBS job ran cleanly, profile parsed, pytest assertion passed against the committed CSV).

## Summary

Adds a groundwater module to G-ADOPT that solves the Richards equation for variably saturated flow. 

The mass term stays symbolic as `Dt(theta(h))` and is discretised with `stage_type="value"`. Irksome turns that into the exact finite difference `(theta_new - theta_old)/dt` instead of the chain-rule form `C(h)*dh/dt`, which is what gets us mass conservation to solver tolerance on DG. With the conservative-update fix that lands upstream as Irksome PR #226 (commit `4820795`), this works for both stiffly-accurate tableaux and non-stiffly-accurate ones (`ImplicitMidpoint`, `GaussLegendre(2)`, `QinZhang`); pre-fix Irksome silently mis-conserved on the non-SA path.

Three soil-water retention models live in `soil_curves.py`: Haverkamp, van Genuchten, and an exponential form. Each provides `moisture_content`, `hydraulic_conductivity`, and `water_retention` as UFL.

For the linear solve there are two preconditioners written for high aspect ratio extruded meshes (basin-scale problems). `VerticallyLumpedPC` is a two-level MG that collapses the vertical direction to an R-element on the coarse level and uses a Galerkin `P^T A P` coarse operator. `VerticallyLumpedHMGPC` is the same idea except the coarse solve sits on the fine mesh's 2D base `MeshHierarchy` and runs geometric MG there. Four presets to pick from: `direct`, `iterative`, `vlumping`, `vlumping_hmg`.

There is also a three-tier adaptive timestep controller — advective CFL from gravity-driven Darcy flux, diffusive CFL via `D(h) = K/C`, and SNES-iteration-count modulation — with retry-on-divergence.

## What's in the PR

**Source.** `gadopt/richards_equation.py`, `gadopt/richards_solver.py`, `gadopt/soil_curves.py`. Existing files touched: `preconditioners.py` for the two new PCs, `utility.py` for the timestep adaptor, `time_stepper.py` so `RKGeneric` accepts `stage_type`/`bc_type` overrides AND a new optional `t` MeshConstant kwarg (re-exported from `gadopt`). The `t` kwarg is the niche-but-essential mechanism that lets BCs and source expressions reference Irksome's stage time directly so that multi-stage methods see `t + c_i*dt` substituted at every stage; without it, time-dependent BC values freeze at `t_n` and the formal order of any multi-stage method collapses to one. Used in the temporal-convergence test and in the Vauclin demo.

**Demos.** `demos/groundwater/2d_vauclin/` (Vauclin 1979 2D water-table recharge with adaptive DIRK22) and `demos/groundwater/3d_cockett/` (Cockett et al. 2018 3D heterogeneous infiltration).

**Tests under `tests/richards/`.**
- `test_richards.py` — Tracy 2D spatial convergence on DQ0/DQ1/DQ2 and a no-flux DG1 case. DG2 was previously longtest-only because it diverged with `DIVERGED_LINE_SEARCH`; this PR fixes the root cause (under-quadrature with non-polynomial soil curves at degree 2 and an SIPG penalty below the coercivity floor) and promotes DG2 to normal CI. Observed rates: DG0 ~1.1, DG1 ~2.0, DG2 ~3.0.
- `test_mass_balance.py` — function-space sensitivity (CG1/CG2 vs DQ0/DQ1/DQ2) at a fixed stiffly-accurate tableau, plus tableau sensitivity (BackwardEuler / ImplicitMidpoint / GaussLegendre(2)) at fixed DQ1. The non-SA cases pass at solver tolerance only against an Irksome that ships PR #226 — they exhibit an O(1) mass defect on pre-fix Irksome.
- `test_temporal_convergence.py` — analytic MMS on the unit square verifying observed orders match the formal order of BackwardEuler, ImplicitMidpoint and GaussLegendre(2) (1.4, 2.0, 4.0 measured). Pins the value of the conservative-update path through `RichardsSolver`.
- `test_vlumping_variants.py` — checks both VLumping presets give the right answer on a tiny extruded mesh; also pins the warning paths for incompatible mesh hierarchies.

**Soil-curve unit tests** in `tests/unit/test_soil_curves.py`.

**Parallel scaling longtest** at `tests/parallel_scaling_richards/`. Runs the three iterative presets against Cockett 3D and two weak-scaling Murrumbidgee variants (vertical and horizontal) at 1, 2, 4 and 8 nodes on Gadi `normalsr`. 28 tuples total. Iteration counts (`expected.csv`) and Sapphire-Rapids timings (`gadi_expected.csv`) regenerated against current code; iteration counts are bit-identical to the pre-existing baseline for every tuple, timings drift ±1-5% (largest single shift 9.3% on the heaviest job, well inside the 20% test tolerance).

## Drive-by fixes

- `dodo.py` loads each `meta.py` under its own package name so two test directories with a `meta.py` each don't step on each other's relative imports.
- `demos/test_all.py` and `demos/generate_expected.py` take a `primary_checks` override so the groundwater demos can compare `min_h`/`max_h` rather than the hard-coded `u_rms`. Composes cleanly with the per-extra-check tuple kwargs added by #489.
- `IrksomeIntegrator` accepts an optional pre-existing `t` MeshConstant. Default behaviour (None) creates one internally — backwards-compatible. Re-exported from `gadopt`.

## Status against tests

- `tests/unit/test_soil_curves.py`: 16 / 16 pass locally.
- `tests/richards/test_temporal_convergence.py`: 3 / 3 pass locally (~18 s).
- `tests/richards/test_mass_balance.py`: 8 / 8 pass locally (~6 min, includes the non-SA tableau cases).
- `tests/richards/test_richards.py`: 4 / 4 pass locally (DG0/DG1/DG2/DG1-no-flux).
- `tests/richards/test_vlumping_variants.py`: passes locally.
- `tests/parallel_scaling_richards/test_parallel_scaling_richards.py`: 86 / 86 longtest assertions pass against the regenerated CSVs (verified by symlinking the production-run outputs).

## Test plan

- [x] `pytest tests/unit/test_soil_curves.py`
- [x] `pytest tests/richards/`
- [x] Demo regression for `2d_vauclin` (`expected.pkl` regenerated for stage-accurate BC)
- [ ] `pytest demos/` for the 3d_cockett demo (heavy, runs serially in CI)
- [ ] Parallel scaling longtest on Gadi normalsr — **gated on fp50 module Irksome bump (see banner)**


sghelichkhani:
Closing this monolithic PR in favour of the reviewable stack that replaces it: #522 (soil curves) → #523 (equation + core solver) → #524 (vertically-lumped preconditioners) → #525 (adaptive timestep) → #509 (SA test suite) → #510 (Vauclin demo) → #511 (Cockett demo + parallel scaling). Each PR is a single layer against the one below, which is far easier to review than this one big diff. The whole module as a single diff still lives at the top of the stack (#511) against current main, so nothing here is lost. The pre-split snapshot is preserved under the tag richardson-pre-ghstack-20260514-155753.

PR #346, opened 2025-12-15
Symlink forward model output for adjoint_ice demo

dsroberts:
Hopefully a workaround for the `adjoint_ice` failures we've been seeing in the CI until the non-recursive makefile structure is in place.


angus-g (approved):
Good to go if it passes (consistently?)

PR #352, opened 2025-12-19
Raise RuntimeError if using limiter in adjoint mode

w3168:


dsroberts (approved):
Happy for this to go in if the tests pass, though that may take a couple of attempts.

Issue #316, opened 2025-10-22
Add a dg limiter test for a flow case


w3168:
Adapt scalar_advection.py to use limiters to increase test coverage

PR #353, opened 2025-12-19
Test scalar advection with DG and G-adopt's VertexBasedP1DGLimiter

w3168:
As highlighted in issue #316 we aren't actually testing G-ADOPT's `VertexBasedP1DGLimiter` that is built on top of Firedrake's `VertexBasedLimiter`. It is kind of being tested in the `thwaites` package https://github.com/thwaitesproject/thwaites but I think it makes more sense to add a test to gadopt. 

This copies the existing scalar advection test which uses CG and Streamline Upwind advection stabilisation, but instead uses a DG discretisation with G-adopt's limiter. The test checks the error norm after rotating the field by 360 degrees as well as the min/max values. 

Applying the limiter prevent some over/undershooting at the expense of some additional smoothing which affects the accuracy i.e. 

DG without limiters
```
initial min q:  1.0
initial max q:  2.0
error norm: 0.0858080603307695
min q:  0.9394761626664626
max q:  2.1183949499490473
undershoot:  0.06052383733353739
overshoot:  0.11839494994904731
```
DG with limiters:
```
initial min q:  1.0
initial max q:  2.0
error norm: 0.09269404521407619
min q:  0.9664247119139915
max q:  1.9870538074053106
undershoot:  0.033575288086008515
overshoot:  0.012946192594689432
```

For background here is a comparison showing the difference between CG (without stabilisation), CG (plus SU stabilisation), DG, (DG +limiters) for one time step
dg_vs_cg_su_vs_lim_step1

and for a full 360 degree rotation

dg_vs_cg_su_vs_lim

Clearly, DG + limiters is much less noisy than CG and significantly better at preserving the original field than CG + SU stabilisation for this case.


Patol75:
Thanks, Will. Any idea why the minimum field value does not remain above 1? Also, not sure if it is meaningful to have a positive overshoot value when the field is not actually overshooting! :D

w3168:
and here's the summary from the zulip chat for longevity... 

_Thomas reminded me I forgot to change the element variant type to 'equispaced'. The default DG variants are 'spectral' ([see here in the manual](https://www.firedrakeproject.org/variational-problems.html#element-variants)), so the vertex based limiter wasnt working properly because the degrees of freedom weren't in the corners of the cell._

_Just to summarise the numbers:_

_Here is the min/max before, when I used the Firedrake default spectral variant_

```
min q:  0.9661267727016505
max q:  1.9870895745290473
undershoot:  0.03387322729834952
No overshoot detected
```

_Here is the min / maximum scalar value after a full rotation with the equispaced element_

```
min q:  0.9994301601160926
max q:  1.8400733290547673
undershoot:  0.0005698398839073526
No overshoot detected
```

_It's interesting because now the limiter acts more aggressively to damp the peak values._

_Here is a run with the timestep halved:_

```
min q:  0.9999959764862678
max q:  1.8211721915827521
undershoot:  4.023513732187567e-06
No overshoot detected
```
_Making the timestep 2x smaller improves the accuracy of the lower bound, so it seems like things are working_

_Here is a run with double the spatial resolution and half the timestep to keep the CFL the same_

```
min q:  0.9993348799658088
max q:  1.9951619718408207
undershoot:  0.0006651200341911689
No overshoot detected
```
_This reduces the amount of smoothing the limiter applies to the peak values so everything seems to be working well._

PR #368, opened 2026-01-29
Fix race condition in adjoint_ice demo

### LINKED ISSUE #367: Intermittent CI failure in adjoint_ice demo due to race condition
## Summary

The `adjoint_ice` demo intermittently fails in CI when running with parallel make (`make -j`). This has been occurring sporadically, causing wasted CI runs and developer time.

Example failed run: https://github.com/g-adopt/g-adopt/actions/runs/21466181885?pr=366

## Root Cause

The `adjoint_ice` demo depends on a checkpoint file (`forward-2d-cylindrical-disp-vel.h5`) produced by `2d_cylindrical_lvv`. The Makefile uses `params.log` as the build target to track completion, but this file is created at the **start** of `2d_cylindrical_lvv.py`, while the HDF5 checkpoint is only valid at the **end**.

When make runs demos in parallel, `adjoint_ice` can see `params.log` exists (thinking the build is complete) and proceed to read the HDF5 file while `2d_cylindrical_lvv` is still writing it.

## Why It Is Intermittent

The failure depends on timing:
- If `2d_cylindrical_lvv` completes before `adjoint_ice` checks: **passes**
- If `adjoint_ice` checks while `2d_cylindrical_lvv` is still running: **fails**

With parallel execution, this timing varies between runs, making the failure appear random.

sghelichkhani:
Use stamp file pattern to track `2d_cylindrical_lvv` completion accurately, and add order-only dependency to prevent parallel scheduling issues.

Closes #367

### DIFF
--- a/demos/glacial_isostatic_adjustment/Makefile
+++ b/demos/glacial_isostatic_adjustment/Makefile
@@ -6,6 +6,9 @@ cases := $(notdir $(gia_cases))
 
 all: $(cases)
 
+# Ensure 2d_cylindrical_lvv completes before adjoint_ice starts
+adjoint_ice: | 2d_cylindrical_lvv
+
 $(cases):
 	$(MAKE) -C $@ $(MAKECMDGOALS)
 
--- a/demos/glacial_isostatic_adjustment/2d_cylindrical_lvv/Makefile
+++ b/demos/glacial_isostatic_adjustment/2d_cylindrical_lvv/Makefile
@@ -2,15 +2,16 @@ include ../../../rules.mk
 
 .PHONY: all clean check
 
-all: params.log
+all: 2d_cylindrical_lvv.done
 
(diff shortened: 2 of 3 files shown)


demos/glacial_isostatic_adjustment/Makefile
```diff
@@ -6,6 +6,9 @@ cases := $(notdir $(gia_cases))
 
 all: $(cases)
 
+# Ensure 2d_cylindrical_lvv completes before adjoint_ice starts
+adjoint_ice: | 2d_cylindrical_lvv
+
```
angus-g: Having said that, maybe this is the only thing that's needed for the CI race condition? The sentinel at least avoids a re-run, so that's a win... I think we could probably put this in for now and get the non-recursive thing back up.

PR #370, opened 2026-02-03
Symbolic Jacobian Symmetrisation to deal with nonlinear rheology in SIPG

### LINKED ISSUE #369: GAMG asymmetry for non-linear rheology.
I would like to re-open this thread for advice. If I take a case like we were using previously (i.e. compressible) and make the viscosity non-linear, I once again get the issue in GAMG re asymmetry. 

Specifically, changing this:
`mu_unclipped = mu_rad * exp(-ln(delta_mu_T) * T_dev)`

To this:
```
mu_lin = mu_rad * exp(-ln(delta_mu_T) * T_dev)
mu_star, sigma_y = Constant(0.5), 2e7 + 1e6*(rmax-r)
epsilon = sym(grad(u))  # strain-rate
epsii = sqrt(inner(epsilon, epsilon) + 1e-10)  # 2nd invariant (with a tolerance to ensure stability)
mu_plast = mu_star + (sigma_y / epsii)
mu_min = conditional(le(mu_plast,mu_lin), 1.0, 0.0)
mu_unclipped = (2. * mu_lin * mu_plast) / (mu_lin + mu_plast)
```

Is this to be expected, or does it hint that there is still something up?

_Originally posted by @drhodrid in https://github.com/g-adopt/g-adopt/issues/310#issuecomment-3741746325_

sghelichkhani:
Fixes https://github.com/g-adopt/g-adopt/issues/369

### DIFF
--- a/gadopt/approximations.py
+++ b/gadopt/approximations.py
@@ -11,10 +11,11 @@ from numbers import Number
 from typing import Optional
 from warnings import warn
 
-from firedrake import Function, Identity, div, grad, inner, sqrt, sym, tr
+from firedrake import Function, Identity, derivative, div, grad, inner, sqrt, sym, tr
 import ufl
+from ufl.algorithms import expand_derivatives, extract_coefficients
 
-from .utility import ensure_constant, vertical_component
+from .utility import depends_on, ensure_constant, vertical_component
 
 __all__ = [
     "BoussinesqApproximation",
@@ -71,6 +72,24 @@ class BaseApproximation(abc.ABC):
 
         """
         pass
+
+    def tangent_stress(self, u, v, **kwargs):
+        r"""Directional derivative $D\sigma(u)[v]$ of the deviatoric stress.
+
+        This is the derivative of `stress(u)` in the direction of `v`. For a
+        viscosity that does not depend on the velocity it is equal to
+        `stress(v)`. The weak (SIPG) boundary terms in the momentum equation
+        use it so that the Jacobian of the weak boundary conditions is
+        symmetric for a strain-rate-dependent viscosity. A subclass can
+        override this with an analytic tangent stress.
+
+        Returns:
+          A UFL expression for the tangent deviatoric stress.
+
+        """
+        if not any(depends_on(self.mu, c) for c in extract_coefficients(u)):
+            return self.stress(v, **kwargs)
+        return expand_derivatives(derivative(self.stress(u, **kwargs), u, v))
 
     @abc.abstractmethod
     def buoyancy(self, p: Function, T: Function) -> ufl.core.expr.Expr:
--- a/gadopt/momentum_equation.py
+++ b/gadopt/momentum_equation.py
@@ -10,11 +10,13 @@ $$
 """
 
 from firedrake import *
+from ufl.algorithms import expand_derivatives, extract_coefficients
 from ufl.indexed import Indexed
 
 from .approximations import QuasiCompressibleInternalVariableApproximation
 from .equations import Equation, interior_penalty_factor
 from .utility import (
+    depends_on,
     is_continuous,
     normal_is_continuous,
     tensor_jump,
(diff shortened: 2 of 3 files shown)


gadopt/stokes_integrators.py
```diff
@@ -568,6 +569,51 @@ def set_form(self) -> None:
...
+        #
+        # Fix: symmetrise the boundary contribution to the Jacobian. Split
+        # F into interior and boundary parts, keep the full Newton Jacobian
+        # for interior terms, and use J_bdy_sym = 0.5*(J_bdy + J_bdy^T)
+        # for boundary terms. This is a quasi-Newton method whose
+        # approximation error is confined to boundary integrals and vanishes
+        # as the boundary condition is satisfied.
+        #
+        # The residual F is unchanged, so the converged solution is exact
+        # and the adjoint tape records the correct forward model.
```
stephankramer: Nice

gadopt/stokes_integrators.py
```diff
@@ -568,6 +569,51 @@ def set_form(self) -> None:
...
+        if self.J is None and self._needs_symmetric_boundary_jacobian():
+            self.J = self._build_symmetric_jacobian()
+
+    def _needs_symmetric_boundary_jacobian(self) -> bool:
+        """Check if Jacobian symmetrisation is needed."""
+        has_weak_sipg_bcs = any(
+            "un" in bc or "u" in bc for bc in self.weak_bcs.values()
```
stephankramer: Bastard, now you have me think about the weak "u" case as well :-) - but I think you're right that becomes asymmetric as well. Not that we're actually using weak "u"... but this makes sense

stephankramer (changes requested):
Very nice! Much simpler then I expected, no need to fiddle with solver options, which is great! Solid test as well. Just a one-line tiny request

sghelichkhani (commented):
Thanks Steph! Added the check for DG

stephankramer (approved):
Yes that's exactly right. Note that these boundary terms are consistent and proportional with the interior volume integrals for pressure gradient and divergence, which are also fundamentally not symmetric (due to density) and so $$C\neq B^T$$ more generally.

There's also the case of pressure dependent rheology (Drucker Prager) which gives a contribution in the (0,1) block of the Jacobian (the "pressure gradient" block) where again it probably doesn't make sense to mix it into the (1,0) "divergence" block.

As you mention on #369 , I think this is now a sensible way to deal with asymmetry in the velocity block. Well possible that there's more we need to do in this area, but it certainly is an improvement over just handing the asymmetric block over to our solver with solver options that don't work with that - which is what we're currently doing.

gadopt/stokes_integrators.py
```diff
@@ -568,6 +569,71 @@ def set_form(self) -> None:
...
+        )
+        nonlinear_mu = depends_on(self.approximation.mu, self.solution)
```
Patol75: So this already shows up on line 354. Can we move it to the `__init__` method, perhaps call it `self.non_linear_mu`, and then use it there and here?

gadopt/stokes_integrators.py
```diff
@@ -568,6 +569,71 @@ def set_form(self) -> None:
...
+        needs_symmetrisation = nonlinear_mu and has_weak_sipg_bcs
+        if needs_symmetrisation and not is_continuous(self.equations[0].trial_space):
```
Patol75: `is_continuous(self.solution_space[0])`

Patol75 (commented):
Just a few minor comments on my side; I trust that the two of you have the maths figured out. :)

gadopt/stokes_integrators.py
sghelichkhani: Done

Patol75 (approved):

sghelichkhani:
Check https://github.com/g-adopt/g-adopt/pull/540 for a follow up.

PR #372, opened 2026-02-05
Per-timestep diagnostics caching

dsroberts:
Adds a decorator for diagnostic functions that caches & reuses results as long as none of the underlying Firedrake functions have changed since the last call. PETSc maintains a counter of every time a vector/matrix is updated, so querying this before recalculating an expensive quantity has the potential to save a little time with e.g. field min/max calculations, especially on larger HPC-scale runs.


angus-g (approved):
Seems sensible. Curious to see a quantification of the performance differences to non-cached diags

PR #375, opened 2026-02-11
Replace recursive Makefiles

angus-g:
This is a different approach to #351 to change the structure of the task runner to be a bit more sane and dependency-capable. I've used [doit](https://pydoit.org/index.html), which is a Python task runner, and is perhaps a bit more suitable than a build system for what we're using it for.

## Structure
There's a top-level `dodo.py` file that defines a) how to find the task files; and b) the tasks themselves. There are three kinds of tasks: `run_case` is the equivalent to `make test`, `convert` is responsible for turning a Python file into a notebook, and `check` knows how to run pytest for a given case.

### Tasks
doit effectively constructs a DAG between "targets" and "file dependencies". Tasks are the edges in the graph. As mentioned above, there are three main tasks. Within each of these, there are several subtasks, representing cases (a directory like `demos/mantle_convection/base_case`) or steps within cases (the forward part of the adjoint test). When we refer to a task like `run_case`, this involves all of the subtasks contained within. Alternatively, we can refer to a single subtask, e.g. `run_case:demos/mantle_convection/base_case:run`. `doit list --all` will give a list of all tasks and subtasks. When actually running, we can also use globs to restrict to some subset of subtasks, e.g. `run_case:tests/*`.

At a technical level, a task (or subtask) is a dictionary either `yield`ed or `return`ed from a method with the name `task_*` in the `dodo.py` file. The important keys here are `actions` (the list of actions the task performs), `file_deps` (which will be considered as dependencies) and `targets` (provided to dependents). In a similar way to Make, doit won't re-run a task it considers up-to-date by looking at its deps and targets.

### "Meta" files
The next obvious question is: how do we define tasks for our cases? The theory is that `dodo.py` should remain relatively untouched in the normal course of things. It searches for files named `meta.py` within the repository that are loaded as modules with a specific structure. I think that this way should simplify changing/adding cases: we don't need to touch `Makefiles` all up the directory tree. These files are probably best explained with a few examples, starting at the easiest, and going up to fairly custom.

#### Simplest: `demos/mantle_convection/base_case`

```python3
entrypoint = "base_case.py"
notebook = "base_case.ipynb"
outputs = ["params.log", "Final_State.h5"]
pytest = "auto"
```

This is fairly typical for a demo:
1. the `entrypoint` is the script for the demo
2. the `notebook` is the name of the notebook that should be generated (this no longer has to have the same basename as the script!)
3. there's a set of `outputs` that are the directory-relative `targets` for the case: `params.log` for the `check` rule, and `Final_State.h5` for the dynamic topography case that consumes this file
4. an "auto" `pytest` uses the `test_all.py` harness

#### More complicated: `tests/adjoint`

```python3
cases = ["damping", "smoothing", "Tobs", "uobs", "uimposed"]
schedulers = ["noscheduler", "fullmemory", "fullstorage"]

checkpoint_filename = "adjoint-demo-checkpoint-state.h5"

steps = {}
for c in cases:
    for s in schedulers:
        case_name = f"{c}_{s}"

        steps[case_name] = {
            "entrypoint": "taylor_test.py",
            "cores": 2,
            "args": case_name,
            "outputs": [f"{case_name}.conv"],
            "deps": [
                {
                    "case": "tests/adjoint",
                    "step": "forward",
                    "artifact": checkpoint_filename,
                    "link": False,
                },
            ],
        }

steps["forward"] = {
    "entrypoint": None,  # dependency-only
    "deps": [
        {
            "case": "demos/mantle_convection/adjoint",
            "step": "forward",
            "artifact": checkpoint_filename,
            "teardown": False,
        },
    ],
    "outputs": [checkpoint_filename],
}

pytest = "local"
```

There's a few more things going on here. Remember that the `meta.py` is still just a Python script, so we can write arbitrary code, as long as the module ends up with the right attributes defined. In this case, instead of a single toplevel task (which is implicitly called "run"), we define a "step" for each of the configurations in the matrix.

This also introduces dependencies: all the adjoint cases depend on the forward checkpoint. The usual workflow is to link in the dependency, then unlink it after the task is done. However, because several cases depend on the one file, we introduce a meta-dependency that's responsible for linking the checkpoint in to the test directory. I think it'd be possible to add a teardown task that depends on all the `.conv` files to unlink it...

#### Most complicated: `tests/analytical_comparisons`

```python3
from .analytical import cases, get_case
from .test_analytical import configs, idfn

case_names = [
    "smooth_cylindrical_freeslip",
    "smooth_cylindrical_zeroslip",
    "delta_cylindrical_freeslip",
    "delta_cylindrical_zeroslip",
    "delta_cylindrical_freeslip_dpc",
    "delta_cylindrical_zeroslip_dpc",
]

steps = {}
for c in case_names:
    outputs = []
    levels = get_case(cases, c)["levels"]

    for config in [x for x in configs if x[0] == c]:
        for level in levels:
            outputs.append(f"errors-{c}-levels{level}-{idfn(config[2])}.dat")

    steps[c] = {
        "entrypoint": "analytical.py",
        "cores": 1,
        "args": """submit -t "tsp -N {cores} -f mpiexec -np {cores}" """ + c,
        "use_tsp": False,
        "outputs": outputs,
    }

pytest = "python3 -m tests/analytical_comparisons -m 'not longtest'"
```

This is one of the tests with dynamically determined cases. It also overrides the run task to not use `tsp` and provide custom arguments, etc.

## Using
- Run all the cases: `doit run_case`
- Run all steps of a case: `doit run_case:tests/adjoint:*`
- Check a case: `doit check:demos/base_case`

The run and convert tasks are integrated into the CI with this PR. I've removed Makefiles from the cases (except for those with HPC tests), which means there's no `make clean`. I think we can just use `git clean -fXd` to remove only ignored files, which I think I've made cover all the generated outputs.

## What next?
I think this is pretty much ready to go. Should the [wiki page](https://github.com/g-adopt/g-adopt/wiki/Creating-new-test-cases-out-of-demos) be the authoritative source for documentation? I could also document some illustrative `meta.py` files or add a page to the [developer guide](https://gadopt.org/documentation/dev/).

### DIFF
--- a/demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
+++ b/demos/glacial_isostatic_adjustment/adjoint_ice/adjoint_ice.py
@@ -513,6 +513,7 @@ updated_out_file = VTKFile("updated_out.pvd")
 
 final_target_displacement = forward_checkpoint.load_function(mesh, name="displacement", idx=max_timesteps)
 final_target_velocity = forward_checkpoint.load_function(mesh, name="velocity", idx=max_timesteps)
+forward_checkpoint.close()
 
 functional_values = []
 
--- a/demos/glacial_isostatic_adjustment/adjoint_ice/meta.py
+++ b/demos/glacial_isostatic_adjustment/adjoint_ice/meta.py
@@ -8,5 +8,5 @@ deps = [
         "notebook": "2d_cylindrical_lvv.ipynb",
     },
 ]
-outputs = ["functional.txt"]
+outputs = ["functional.txt", "taylor_test_minconv.txt"]
 pytest = "local"


demos/mantle_convection/free_surface/meta.py
```diff
@@ -0,0 +1,12 @@
+entrypoint = "free_surface.py"
+notebook = "free_surface.ipynb"
+notebook_ouptuts = ["temperature_warp.gif"]
```
sghelichkhani: Typo.
```suggestion
notebook_outputs = ["temperature_warp.gif"]
```

demos/smoothing/meta.py
```diff
@@ -0,0 +1,13 @@
+entrypoint = "smoothing.py"
+notebook = "smoothing.ipynb"
+cores = 1
+deps = [
+    {
+        "case": "tests/adjoint_2d_cylindrical",
```
sghelichkhani: `tests/adjoint_2d_cylindrical` has no `meta.py` is this intentionally kept as a longtest-only Make case?

.github/workflows/test.yml
```diff
@@ -162,7 +162,7 @@ jobs:
           tsp --timeout -T 7200
           tsp --memprof
           . .venv/bin/activate
-          make -j test
+          uv run doit run -n $(nproc --all) run_case
         env:
           GADOPT_LOGLEVEL: INFO
           BATCH_MODE: 1
```
sghelichkhani: Dead env var: `dodo.py` uses `get_var("batch_mode", "YES")` instead of reading the environment.
```suggestion
```

tests/analytical_comparisons/meta.py
```diff
@@ -0,0 +1,30 @@
...
+    levels = get_case(cases, c)["levels"]
+
+    for config in [x for x in configs if x[0] == c]:
+        for level in levels:
+            outputs.append(f"errors-{c}-levels{level}-{idfn(config[2])}.dat")
+
+    steps[c] = {
+        "entrypoint": "analytical.py",
+        "cores": 1,
+        "args": """submit -t "tsp -N {cores} -f mpiexec -np {cores}" """ + c,
```
sghelichkhani: `{cores}` is a literal string here, not an f-string. Intentional (i.e. `analytical.py` substitutes it at runtime)?
```suggestion
        "args": f"""submit -t "tsp -N {cores} -f mpiexec -np {cores}" """ + c,
```

demos/smoothing/meta.py
angus-g: Yes, this dep just stitches together the paths to the LFS-stored checkpoint file.

dsroberts:
I'm happy to look into properly integrating the HPC tests with `doit`. It doesn't make sense to hang on to bits of the vestigial makefile system where we've had to hand-roll task management code when we're switching to an actual task manager.

sghelichkhani (approved):
It looks good to me! But I think might be worth for someone else, probably @dsroberts to have a look too. My only suggestions is about single-step path seems to be forgiving for `outputs`, but the multi-step path is not. 

The way I understand it there are two code paths through `normalise_meta()` depending on whether a meta file defines a single case or multiple steps:

So for **Single-step** e.g., demos/mantle_convection/base_case/meta.py):

```python3
entrypoint = "base_case.py"
outputs = ["params.log"]
```
normalise_meta() builds a dict with safe defaults so if someone forgot outputs, it defaults to `[]`

```python3
("outputs", []),  # default is [], so getattr(meta, "outputs", []) always works
```
**Multi-step** path like (e.g. tests/adjoint/meta.py):

```
  steps = {
      "damping_noscheduler": {
          "entrypoint": "taylor_test.py",
          "outputs": ["damping_noscheduler.conv"],
          ...
      },
  }
```
so `normalise_meta()` just returns meta.steps directly (line 116). There is no validation or defaults So if someone wrote a step like:

```
  steps["sth"] = {
      "entrypoint": "run.py",
      # forgot "outputs"
  }
```
Then make_run_task() at line 103 does cfg["outputs"] -> KeyError. Same for task_check() at line 236.

Is the asymmetry is a feature? Someone authoring a new meta.py with steps might reasonably expect the same defaults apply. Not a bug since all existing meta files are correct, just thinking if we need to be consistent.

.github/workflows/test.yml
```diff
@@ -162,7 +162,7 @@ jobs:
           tsp --timeout -T 7200
           tsp --memprof
           . .venv/bin/activate
-          make -j test
+          uv run doit run -n $(nproc --all) run_case
```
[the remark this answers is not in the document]
angus-g: Yep, good idea. It does cancel dependencies where the requirements fail, and it does seem to continue with unrelated tasks. Will add.

angus-g (commented):
> Should .doit.db.db be added to .gitignore as well?

Yep.

> Also, sorry to be a pain, but can we please have docstrings and type hints in dodo.py?

Yep.

tests/analytical_comparisons/test_analytical.py
```diff
@@ -1,9 +1,16 @@
...
 import pandas as pd
 import pytest
+import itertools
+import sys
+
+
+# Need to do this in order to have the imports from
+# analytical work in the top-level directory of the repo
+sys.path.insert(0, str(Path(__file__).parent))
+from analytical import cases, get_case  # noqa: E402
```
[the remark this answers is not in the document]
dsroberts: `python3 -m pytest tests/analytical_comparison`, `python3 -m pytest` and `doit run -s check:tests/analytical_comparisons` from the top-level directory of the repo and `python3 -m pytest` in the `tests/analytical_comparisons` directory. Perhaps the difference is I'm not running G-ADOPT from within the firedrake virtualenv?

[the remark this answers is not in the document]
angus-g: I think it'll work with a `__init__.py` in the `tests/analytical_comparisons` directory for all those cases. Good old Python package discovery :roll_eyes:

w3168 (commented):
Thanks @angus-g seems to simplify setting up demos a lot from a user/demo creator point of view. 

Just flagging the two comments we talked about in the office

w3168:
@angus-g two hopefully easy fixes:

1) from zulip chat, should we run pytest even if the run step fails? 

2) Is there a way for the pytest summary to print out the subdirectory again? 

Now we are back to having three `base_case.py` e.g. https://github.com/g-adopt/g-adopt/actions/runs/22171052483

w3168:
Thanks @angus-g ! Maybe changing the subtask name can be done in another pr?

angus-g:
> Well, the issue about invalid characters in artefact paths is patched.

Great, thanks!

> However, I'm not sure its the right approach to have tsp make the judgement call about which bits of label to use for output purposes.

I agree with this, I was probably thinking more along the lines of adding another column for the summary, so it has
- subtask (i.e. the `-L` argument)
- command (currently "case" is `{subtask}: {command}`
- time
- maxrss
- success

angus-g:
> Done in `tsp`. Next run will have a nicer looking summary table.

Oops, I guess we'll need a new tsp release and to update the `TSP_VERSION` variable in actions for this to take effect.

Looks like the memprof step is working again though :tada:

angus-g:
Okay, summary works now. Are there any things I've overlooked?

PR #469, opened 2026-03-03
Add coupled formulation, precursor to introducing power law rheology for GIA

w3168:
Motivation: provide framework for implementing power law rheology, as documented in our GMD preprint. I found it easier to set nonlinear solver options by writing the momentum equation and internal variable equation(s) separately as a mixed system. 

Implementation
- Add a `mass_term_without_irksome` method to allow mass terms in the internal variables (This is a fix to revert the original `mass_term` code that was in `equations.py`, removed in #355 
- Change `dot` to `inner` for the source/sink term in `scalar_equation.py` because internal variable is a tensor.
- Adds `CoupledInternalVariableSolver` built off `StokesSolverBase`
- N.b We can use the existing rheology parameterisations (Maxwell, Burgers etc) with this solver without modifying any code in `approximations.py`

Testing: 
- Adds modified version of analytical cases from the substitute method, testing a (compressible) elastic, viscoelastic Maxwell, and viscoelastic Burgers rheology case.  
- There is a fair amount of boiler plate code that is copied from `tests/viscoelastic_internal_variable`, so potentially this could get streamlined.

Caveats:
- It follows a different implementation for non-Irksome timestepping than introduced in #355 to fix the free surface implementation. Personally I think the approach here is clearer because there is only one equation associated with each internal variable, as opposed to initialising a separate equation just for the mass term in #355. 

Irksome... : 
- I haven't tested this with Irksome but it seems like a good idea down the line. 
- I don't think  we should prioritise getting Irksome in now though, because...
- 1) Potentially, we will run into similar issues discussed on zulip re oscillating $$u^{n+1}$$ to $$u^{n}$$ solutions for the stokes + free surface system when using implicit midpoint as implemented in Irksome, since the momentum equation contains no time derivative for this internal variable formulation...  
- 2) We didn't use Irksome for the paper so it would probably require a lot more testing.

### DIFF
--- a/dodo.py
+++ b/dodo.py
@@ -463,13 +463,19 @@ def task_convert() -> Iterator[DoitTask]:
 
             outputs = [
                 case_dir / output
-                for output in cfg.get("notebook_outputs", []) + [cfg["notebook"]]
+                for output in [cfg["notebook"]] + cfg.get("notebook_outputs", [])
             ]
             notebook_deps.extend(outputs)
-            paths = " ".join(str(o.relative_to(demo_path)) for o in outputs)
+            paths = [o.relative_to(demo_path) for o in outputs]
+            notebook_path, *output_paths = paths
+            output_paths = " ".join(str(o) for o in output_paths)
             actions.append(
-                f"tar --transform='s|/.*/|/|' --append --file artifact.tar --directory demos {paths}"
+                f"tar --transform='s|/.*/|/|' --append --file artifact.tar --directory demos {notebook_path}"
             )
+            if output_paths:
+                actions.append(
+                    f"tar --append --file artifact.tar --directory demos {output_paths}"
+                )
 
             yield make_convert_task(case_dir, step, cfg)
 


w3168:
Blocks: #470

gadopt/stokes_integrators.py
```diff
@@ -859,6 +860,116 @@ def solve(self) -> None:
...
+        # Loop over number of internal variables
+        for i in range(len(maxwell_times)):
+            residual_terms.append(internal_variable_terms)
+            scaling_factors.append(-self.theta*self.scaling_factor)
+            eqs_attrs.append(
+                {"source": strain / maxwell_times[i],
+                 "sink_coeff": 1 / maxwell_times[i],
+                 "trial_old": self.solution_old_split[i+1],
+                 "dt": self.dt}
+            )
```
angus-g:
```suggestion
        for i, maxwell_time in enumerate(maxwell_times):
            residual_terms.append(internal_variable_terms)
            scaling_factors.append(-self.theta*self.scaling_factor)
            eqs_attrs.append({
                "source": strain / maxwell_time,
                 "sink_coeff": 1 / maxwell_time,
                 "trial_old": self.solution_old_split[i+1],
                 "dt": self.dt,
            })
```

gadopt/stokes_integrators.py
```diff
@@ -859,6 +860,116 @@ def solve(self) -> None:
...
+        A `MixedVectorSpaceBasis` for the operator's smallest eigenmodes (e.g. rigid
+        body modes)
+    """
+
+    name = "InternalVariable"
+
+    def __init__(
+        self,
+        solution: fd.Function,
+        approximation: BaseGIAApproximation,
```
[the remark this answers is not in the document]
w3168: makes sense!

tests/viscoelastic_internalvariable_coupled/internalvariable_viscoelastic_coupled_freesurface.py
```diff
@@ -0,0 +1,408 @@
...
+            # time)
+            error_nondim += local_error_nondim
+
+        # Write output:
+        if timestep % dump_period == 0:
+            log("timestep", timestep)
+            log("time", float(time))
+            if MPI.COMM_WORLD.rank == 0:
+                displacement_df.to_csv(
+                    f"surface_displacement_dt{float(dt)}_nx{nx}arrays_bulktoshear{float(bulk_shear_ratio)}.csv"
```
angus-g: Needs `*.csv` in the local `.gitignore` since this is a generated output.

w3168 (commented):
Thanks @angus-g

tests/viscoelastic_internalvariable_coupled/internalvariable_viscoelastic_coupled_freesurface.py
w3168: added

gadopt/stokes_integrators.py
```diff
@@ -859,6 +860,120 @@ def solve(self) -> None:
...
+      nullspace:
+        A `MixedVectorSpaceBasis` for the operator's kernel
+      transpose_nullspace:
+        A `MixedVectorSpaceBasis` for the kernel of the operator's transpose
+      near_nullspace:
+        A `MixedVectorSpaceBasis` for the operator's smallest eigenmodes (e.g. rigid
+        body modes)
+    """
+
+    name = "InternalVariable"
```
Patol75:
```suggestion
    name = "CoupledInternalVariable"
```

gadopt/stokes_integrators.py
```diff
@@ -859,6 +860,120 @@ def solve(self) -> None:
...
+        source = self.approximation.buoyancy(u) * self.k
+        strain = self.approximation.deviatoric_strain(u)
+        maxwell_times = self.approximation.maxwell_times
+
+        residual_terms = [
+            compressible_viscoelastic_terms,
+        ]
+        eqs_attrs = [
+            {"stress": stress, "source": source},
+        ]
```
Patol75:
```suggestion
        residual_terms = [compressible_viscoelastic_terms]
        eqs_attrs = [{"stress": stress, "source": source}]
```

gadopt/scalar_equation.py
```diff
@@ -140,6 +140,23 @@ def mass_term(eq: Equation, trial: Argument | Indexed | Function) -> Form:
...
+          G-ADOPT Equation.
+        trial:
+          Firedrake trial function.
+
+    Returns:
+        The UFL form associated with the mass term of the equation.
+
+    """
+    mass_scaling = getattr(eq, "mass_scaling", 1.0)
+    return mass_scaling * inner(eq.test, (trial - eq.trial_old)/eq.dt) * eq.dx
```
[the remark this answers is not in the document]
w3168: Thanks Thomas - I've had a go at implementing point 1. Point 2 we discussed with Sia on zulip and decided to change the free surface implementation to be more like this.

sghelichkhani (changes requested):
Rather than cluttering this with inline suggestions, I put together a small PR targeting your branch (g-adopt/g-adopt#477) so you can see the diff cleanly and accept or reject the whole thing in one go.

Here is a summary of what is in there and why. The mass_term in free_surface_equation.py was inconsistent with the irksome/non-irksome dispatching pattern introduced in scalar_equation.py. I unified the two so that free_surface_equation.mass_term also checks a use_irksome attribute and routes accordingly, with StokesSolver explicitly passing use_irksome=False for the existing free surface path. While doing that, I added an explicit ValueError in both mass_term functions when use_irksome=False but the required dt and trial_old attributes are absent. Without it, the failure mode is a confusing UFL assembly error that can take a while to trace back to a missing equation attribute.

On CoupledInternalVariableSolver: the theta constructor parameter was problematic because only backward Euler is actually implemented. Advertising a general theta-scheme interface without testing or using it properly for both the strain rate and sink term would give silently wrong results if someone passed theta != 1. I removed it from the constructor and promoted it to a private class constant _theta = 1.0 with a short note about what a proper theta scheme would require. I also added a docstring explaining the internal variable discretisation, the sign convention in the residual assembly loop, and what a future Irksome integration path might look like.

Happy for you to drop anything by pushing to that branch before you merge.

sghelichkhani:
None of them deal breaking, so let's focus on getting this in.

sghelichkhani (approved):
LGTM!

PR #470, opened 2026-03-03
Add power law rheology for GIA

w3168:
Second part of #469 introducing power law rheology. 

Motivation: we have been using power law rheology for our GMD preprint. 

Implementation: 
- Modifies implementation of stress terms in `approximation.py` to allow for power law rheology as outlined in our preprint. 

Testing: 
- Adds a 3d test based on the (default) substitute formulation of the 3d weerdesteijn test case but now using the coupled formulation. Two layers in the upper mantle use power law rheology, whereas the lithosphere and lower mantle still have linear rheology.
- The test is quite slow so I have limited the number of timesteps to two in the CI test.

EDIT: Misc: 
- I found a bug in tests/3d_weerdesteijn that wasn't being tested as part of the ci.... Maybe this shouldnt be in this PR?

### DIFF
--- a/gadopt/approximations.py
+++ b/gadopt/approximations.py
@@ -600,13 +600,19 @@ class InternalVariableApproximation(BaseGIAApproximation):
       shear_modulus:     shear modulus
       viscosity:         viscosity
       bulk_shear_ratio:  Ratio of bulk to shear modulus
-      power_law:         Flag for using composite power-law rheology formulation
-      exponent:          Power-law exponent for composite power-law formulation
-      transition_stress: Stress level in MPa where nonlinear, dislocation creep begins
-                         to dominate over linear, diffusion creep for composite power
-                         law rheology
-      background_stress: Background stress field to apply with composite power-law
-                         rheology , e.g. from mantle convection simulation
+      exponent:          Power-law stress exponent n. Defaults to 1, which recovers
+                         Newtonian (linear) viscosity with no special casing. Set n > 1
+                         for composite (diffusion + dislocation) creep. The power-law
+                         factor is always evaluated; for n=1 it is identically 1.
+      transition_stress: Stress level (in sqrt(J2) units, consistent with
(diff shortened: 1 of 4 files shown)


sghelichkhani (changes requested):
While reviewing this PR, a few issues were identified in the power-law rheology implementation. These are being addressed in #478, which targets this branch rather than main, so they can be reviewed and merged here first.

The main points:

1. second_stress_invariant computes sqrt(s_ij * s_ij) rather than sqrt(0.5 * s_ij * s_ij). The correct second invariant is sqrt(J2) = sqrt(0.5 * s_ij * s_ij), which is the standard quantity in mantle rheology power-law formulations. The current implementation is off by a factor of sqrt(2) relative to convention, meaning any transition_stress value supplied by the user is interpreted in non-standard units.

2. The power_law boolean flag creates diverging code paths and is mathematically redundant. With n=1 the power-law factor is identically 1, so Newtonian behaviour falls out of the same formula without a separate flag. Removing the flag also surfaces a UFL evaluation bug: 0^0 is computed as 0 rather than 1 by UFL, causing the factor to evaluate to 0.5 instead of 1 for Newtonian layers when background_stress=0. This was previously masked by doubling the input viscosity for n=1 layers in the test.

3. set_equations mutates self.approximation.maxwell_times in place. This is replaced with a local list so the approximation object is not modified by the solver.

4. A dedicated set of default solver parameters (coupled_gia_solver_parameters) is introduced for CoupledInternalVariableSolver with Newton SNES always active, removing the need for the user to conditionally merge solver parameter dicts depending on whether power-law rheology is enabled.

sghelichkhani (approved):
If https://github.com/g-adopt/g-adopt/actions/runs/22750043523/job/65982727046 passes, then merge https://github.com/g-adopt/g-adopt/pull/478, then this should be good to go. So pending on APPROVED.

PR #478, opened 2026-03-06
Refactor power-law rheology: unify Newtonian and non-Newtonian code paths

sghelichkhani:
This PR refactors the composite power-law rheology implementation in InternalVariableApproximation, resolving several mathematical and architectural issues identified during review.

The core physics -- composite diffusion and dislocation creep expressed as a viscosity correction factor -- is correct and unchanged. What changes here is how it is structured and how edge cases are handled.

The power_law boolean flag is removed. Instead, the stress exponent n (defaulting to 1) is the sole control. With n=1 the power-law factor is identically 1, recovering Newtonian viscosity without any special casing or diverging code paths. This also fixes a silent UFL bug where 0^0 evaluated to 0 rather than 1, causing Newtonian layers to have their viscosity silently halved when background_stress=0 -- previously worked around by doubling the input viscosity in the test, which is now removed.

second_stress_invariant is corrected to compute sqrt(J2) = sqrt(0.5 * s_ij * s_ij), the standard second invariant used in mantle rheology. The previous implementation computed sqrt(2*J2), which was inconsistent with standard conventions. The transition_stress value in the test is adjusted by 1/sqrt(2) accordingly.

The in-place mutation of approximation.maxwell_times inside set_equations is replaced with a local list, preserving the approximation as a passive parameter container whose state is not altered by the solver.

coupled_gia_solver_parameters is introduced as a module-level default for CoupledInternalVariableSolver, combining the fieldsplit iterative setup with Newton SNES always active. For a Newtonian problem SNES converges in one iteration at negligible cost; for power-law rheology full Newton iteration proceeds as needed. The explicit solver parameter block and conditional SNES setup in the test are removed.

Three unit tests are added covering the corrected invariant formula, the Newtonian factor=1 behaviour (the 0^0 fix), and the analytic power-law factor value for n=3.

### DIFF
--- a/gadopt/approximations.py
+++ b/gadopt/approximations.py
@@ -744,9 +744,13 @@ class InternalVariableApproximation(BaseGIAApproximation):
         n = self.exponent
         a = self.background_stress / self.transition_stress
         b = (sigma + self.background_stress) / self.transition_stress
-        # Guard against 0^0: when the base is zero, the correct value of x^0 is 1.
-        a_pow = ufl.conditional(ufl.eq(self.background_stress, 0), ufl.as_ufl(1), a ** (n - 1))
-        b_pow = ufl.conditional(ufl.eq(sigma + self.background_stress, 0), ufl.as_ufl(1), b ** (n - 1))
+        # Guard against 0^0: UFL evaluates 0^0 as 0 rather than 1. This
+        # only arises when n=1 (exponent n-1 = 0); for n>1, a zero base
+        # raised to a positive power correctly evaluates to 0. Conditioning
+        # on the exponent rather than the base avoids incorrectly setting
+        # a^(n-1) = 1 for n>1 when background_stress = 0.
+        a_pow = ufl.conditional(ufl.eq(n - 1, 0), ufl.as_ufl(1), a ** (n - 1))
+        b_pow = ufl.conditional(ufl.eq(n - 1, 0), ufl.as_ufl(1), b ** (n - 1))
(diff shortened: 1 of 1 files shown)


gadopt/stokes_integrators.py
```diff
@@ -137,6 +137,43 @@
...
+    "fieldsplit_1_assembled_pc_type": "sor",
+    "fieldsplit_1_ksp_rtol": 1e-5,
+}
+"""Default iterative solver parameters for CoupledInternalVariableSolver (GIA problems).
+
+Uses a Newton SNES outer loop with a GMRES/fieldsplit preconditioned inner solve. SNES
+is always active: for a Newtonian rheology (exponent=1) the power-law factor is
+identically 1 and the system is linear, so SNES converges in a single iteration at
+negligible extra cost. For power-law rheology (exponent > 1) the full Newton iteration
+is required. The snes_rtol is set looser than ksp_rtol so that the single-iteration
```
w3168: So I don't think is actually true based on the above. Maybe the comment is correct but so far I've had the outer tolerance tighter...

sghelichkhani: Up to you.

gadopt/stokes_integrators.py
```diff
@@ -975,6 +1011,29 @@ def set_equations(self) -> None:
                 )
             )
 
+    def set_solver_options(
```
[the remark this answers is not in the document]
sghelichkhani: Ok hopefully this is fixed now. I will go for an action run now.

PR #481, opened 2026-03-09
Exclude Zenodo tags as version candidates

angus-g:
setuptools_scm is pretty eager in the tags it picks up (anything containing a number!). We never want a "Zenodo_*" tag as a software version, so we can ignore these when getting the list of possible tags as versions.


Patol75 (approved):
Solves the `pip install` issue. Thank you!

PR #484, opened 2026-03-19
Include z translation in near nullspace for 3D

### LINKED ISSUE #483: Incomplete near nullspace in 3D analytical cases
Not too important, but I noticed that the near null space defined in the spherical analytical tests cases is a bit funny as it currently says:
```
Z_near_nullspace = create_stokes_nullspace(Z, closed=False, rotational=True, translations=[0, 1])
```
i.e. for some reason it leaves out translation in the z direction, which is bit weird. I presume this has happened because we just copied the code from the 2D case.

I'll start a branch to see if this changes things dramatically. I suspect not at all, but it's probably better to change to it to the full near nullspace just so we have the setup as a correct template.

stephankramer:
Fixes #483


dsroberts:
Easiest way to change which module is loaded in `gadopt_hpc_helper` is via the `--template-file` option. Take `run.template` from `tests/parallel_scaling` and replace `{prescript}` with:
```
module use /g/data/fp50/modules
module load firedrake/main-20260312
```
That being said, given you have to load a Firedrake module to use `gadopt_hpc_helper`, I should really make it so that it picks up whatever firedrake is loaded and uses that in the job script.

dsroberts:
> So it seems we're no longer the longtests regularly? Why is this? Was it because something broke and noone had time to look into it - or due to changes in the setup where we're running this?

This was due to the Firedrake release/main switch. G-ADOPT tracks `main`, but we don't install `main` on the HPC systems unless someone specifically needs it. To reinstate the long tests, we'd first need to have periodic installations of Firedrake `main`, which would involve periodic installations of PETSc `main`. Which would mean after a while (if everything worked perfectly) we'd have dozens of PETSc/Firedrake installs that have only ever been used once to run a long test and we run out of storage in our build project. If everything doesn't work perfectly, debugging new failures in weekly or fortnightly Firedrake installations is going to take a lot of my time - releases + main on request is about all the time I have to spend on it. Whilst the process is largely automatic, things can and do go wrong, and that takes time. For example, `main` currently doesn't install on Setonix and I haven't had time to look into that. Before we go blindly into 'just fix whatever is wrong and run long tests again' we need to develop something of a policy around how we actually manage this that takes into account the resources we have available - machine and human.

angus-g:
Well, for whatever reason the errors I was hitting the other day aren't showing up on a (local) current Firedrake `main` (with assess 1.4). It'd be good to verify correctness with an equivalent HPC build.

dsroberts:
The longtests ran successfully on this branch: https://github.com/g-adopt/g-adopt/actions/runs/24117713578.

dsroberts:
Is this ready to be merged? We have a successful longtest run on this branch. It'd be good to get it out of the way and have one less outstanding PR.

drhodrid:
This is ready to go IMO. Could you quickly comment on whether or not it changes the number of iterations @dsroberts? Sorry if I missed this somewhere!

drhodrid (approved):

dsroberts:
Iteration counts for the spherical analytical comparisons are consistent between this branch and the most recent longtest run from `main`.

PR #486, opened 2026-03-26
Fix smoke test: add safe.directory for containerised git

### LINKED ISSUE #485: Smoke test fails: dubious ownership in containerised git
The smoke_test CI job fails during `uv pip install .` because vcs-versioning calls git to determine the package version, and git refuses with a "dubious ownership" error. This happens because the container user UID differs from the UID that checked out the repo.

The build job already handles this with `git config --global --add safe.directory`, but the smoke_test job is missing it.

Failing run: https://github.com/g-adopt/g-adopt/actions/runs/23579325879

sghelichkhani:
The smoke_test job runs inside a Firedrake container where the checkout UID differs from the container user. When vcs-versioning calls git during `uv pip install .`, git refuses with a "dubious ownership" error and the build fails. The build job already had `git config --global --add safe.directory` but the smoke_test job was missing it — this adds the same line there.

Closes #485

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -57,6 +57,7 @@ jobs:
         run: |
           apt update
           apt install -y git-lfs
+          git config --global --add safe.directory $GITHUB_WORKSPACE
 
       - uses: actions/checkout@v4
         with:


dsroberts:
What I'd like to understand before approving this is why it stopped working between the 20th of March and today. The checkout action already does  `/usr/bin/git config --global --add safe.directory /__w/g-adopt/g-adopt`. It makes sense that the build and render jobs need it as we're also grabbing the `animate` repo and installing that. But why would the smoke test need it?

PR #489, opened 2026-03-30
Add horizontal diagnostic velocity to sphere 3D demo

stephankramer:
At the request of @drhodrid 

@dsroberts : this proposes to extend GeodynamicalDiagnostics , exposing the **funcs argument of the base class, so you can compute stats for additional diagnostic fields (in addition to u,p and T). Let me know what you think


drhodrid (commented):
Thanks for this @stephankramer. Two minor things from me. 1. Can we use the term tangential rather than horizontal in the demo? Is this generic enough to work for both Cartesian and spherical domains? 2. For the project, does this default to a direct solver? If so, we may need to consider a set of solver params because on the large spherical runs these direct solvers will not be possible.

dsroberts:
Hi @stephankramer. This looks OK. What I was hoping for diagnostics is that stuff like this would be incorporated into `BaseDiagnostics` or `GeodynamicalDiagnostics` directly, as I'm sure tangential diagnostics has utility beyond this demo. A lot of the new code in `3d_spherical.py` already exists in some form in `BaseDiagnostics`. We already have the cached `get_upward_component` function, we can pick an appropriate function space to interpolate an expression on to. I think this would be more useful overall if `u_hor_max`, `div_u_hor_min_base`, and `div_u_hor_max_base` were implemented as functions in `GeodynamicalDiagnostics` like `Nu_top` and `Nu_bottom`.

As an aside, the `register_functions` function in `BaseDiagnostics` is the intended path for things like tracers. In the case where a diagnostic function really has no utility outside of a specific demo, a user would first initialise a `GeodynamicsDiagnostics` object, then add new functions via `gd.register_functions(f=f)`. This specific use case isn't documented, and was a last minute addition during testing as I found it really useful for interactively constructing new diagnostics.

dsroberts:
@stephankramer this looks a lot better, thanks. In response to the third and fourth point you've raised: 3) You're right, a subclass could definitely do `self.u_vertical = vertical_component(u)`, there would be no issue with that. The advantage of choosing `gd.get_upward_component(u)` are the sanity checks it performs on the input, the lazy initialisation and not needing to manage another class variable. As for endless combinations of UFL expressions in diagnostics, I agree that that's for the geodynamicists to decide. My thinking was that the `BaseDiagnostics` contains all the lego bricks that the domain-specialists can put together however they like to construct diagnostics. As for 4), I use the standard "leading underscore indicates a private function" convention, so no underscore on `register_functions` means intended for both internal and external use. If there was some problem with calling it after initialisation, I'd have it throw an error at that point. Since `quad_degree` has a default value, its possible to call without it. `gd.register_functions(u_horiz=u_horiz)` is totally fine. In any case, I'll update the docs to reflect that use case.

dsroberts:
Looks good to me. I'll let @drhodrid do the official review.

demos/mantle_convection/3d_spherical/3d_spherical.py
```diff
@@ -41,9 +41,9 @@
 Z = MixedFunctionSpace([V, W])  # Mixed function space.
 
 z = Function(Z)  # A field over the mixed function space Z.
-u, p = split(z)  # Returns symbolic UFL expression for u and p
-z.subfunctions[0].rename("Velocity")
-z.subfunctions[1].rename("Pressure")
+u, p = z.subfunctions
```
Patol75: I imagine there is a world where @drhodrid could raise that this change does not preserve consistency across our demos, but maybe not.

Patol75 (commented):
Looking good! Thank you for introducing the functionality and demonstrating how to build upon @dsroberts's diagnostics framework. Perhaps a small comment about naming: horizontal components may not be the most accurate way to describe what is happening. For example, `vertical_component` returns a scalar, whereas `horizontal_components` would return a Firedrake Function or Expression that represents a vector orthogonal to gravity. Maybe we could rename it `horizontal_vector`, where we would make clear that horizontal means orthogonal to gravity?

gadopt/diagnostics.py
```diff
@@ -571,6 +571,24 @@ def get_upward_component(self, f: fd.Function) -> Operator:
         self._check_dim_valid(f)  # Can't take upward component of a scalar function
         return vertical_component(f)
 
+    @cache
+    def get_horizontal_components(self, f: fd.Function) -> Operator:
+        """Get the horizontal components of a function.
+
+        Returns a UFL expression for the horizontal components of a function. Uses the
```
drhodrid, line 578: Add here that this is orthogonal to gravity direction, for clarity?

demos/mantle_convection/3d_spherical/3d_spherical.py
drhodrid: Yes I agree. I think for consistency we should maintain the subfunctions as they were previously. Then lower down -- u_horizontal = gd.get_horizontal_components(u) -- use that subducntion call once and everything else can remain as you have it.

gadopt/utility.py
```diff
@@ -124,6 +124,12 @@ def vertical_component(u):
         return sum([n_i * u_i for n_i, u_i in zip(n, u)])
 
 
+def horizontal_components(u):
+    """Returns expression for vector u with vertical component removed."""
```
drhodrid: Again here can it be clearer that this isn't always vertical - i.e. in direction of gravity or similar?

drhodrid (approved):
I think aside from the very minor things I flagged, this is now good to go. I visualised the fields and it's working nicely and easy to follow and set up. Thanks @stephankramer for your efforts here. I know I asked it as a throw away request, without appreciating how much effort would be involved!

gadopt/diagnostics.py
stephankramer: sure

demos/mantle_convection/3d_spherical/3d_spherical.py
stephankramer: Fine. But tbh I think the current version is confusing: it first defines u,p as `split(z)` but then immediately follows by accessing the components in a different way. Also I think `output_file.write(*z.subfunctions, T, T_dev)` isn't very easy to understand for a novice user.

gadopt/utility.py
stephankramer: I'll add orthogonal to gravity here as well. I'm not entirely sure what you mean though: vertical is synonymous with gravity direction, no?

stephankramer:
Ugh, so this requires some more feedback then. I can't actually do:
```
u, p = split(z)
...
gd = GeodynamicalDiagnostics(z, T, boundary.bottom, boundary.top, quad_degree=6)
u_horizontal = gd.get_horizontal_components(u)
```
because `gd` does not recognize u (which is really the expression `as_vector((z[0], z[1], z[2]))`) as a Function in its function context dict. Instead it has to be either
```
u_horizontal = gd.get_horizontal_components(z.subfunctions[0])
```
which is not pretty, or
```
u_horizontal = gd.get_horizontal_components(gd.u)
```
which is bit obscure as well.

Of course I could just do
```
u_horizontal = horizontal_components(u)
```
I just wanted to be constistent with the existing `get_upward_component()` method - although to be perfectly honest I'm not sure I really see the added benefit of caching ufl expressions on the diagnostics object.

Opinions?

dsroberts:
> Ugh, so this requires some more feedback then. I can't actually do:
> 
> ```
> u, p = split(z)
> ...
> gd = GeodynamicalDiagnostics(z, T, boundary.bottom, boundary.top, quad_degree=6)
> u_horizontal = gd.get_horizontal_components(u)
> ```
> 
> because `gd` does not recognize u (which is really the expression `as_vector((z[0], z[1], z[2]))`) as a Function in its function context dict.

OK, this is a problem with the diagnostics. What you've done should definitely work, I think perhaps I assumed `split(z)` and `z.subfunctions` were the same thing. I guess once you evaluate the ufl they come out numerically the same, but have quite different forms? In any case, `split(z)` is just as valid `z.subfunctions` and the diagnostics should be able to handle that.

There are a couple of ways I can think to handle this, the leading candidate for me right now is to relax the restriction on not storing mixed functions and store `z` as well as `u` and `p`.

> although to be perfectly honest I'm not sure I really see the added benefit of caching ufl expressions on the diagnostics object.

The idea was that constructing a ufl expression is work that only needs to be done once, so even if it isn't much work, it should still only happen once. On the other hand, the `vertical_component(f)` and `horizontal_components(f)` expressions for a given `f` won't change whether they're called from inside the diagnostics or not, so perhaps the functions in `utility.py` should be cached. Given the results of `_check_present` and `_check_dim_valid` are cached, that'd remove the need to cache on the diagnostics object.

dsroberts:
> the leading candidate for me right now is to relax the restriction on not storing mixed functions

I remember now why this restriction exists. Mixed functions can now live on multiple domains because of submesh, so this is a method of restricting diagnostics to a single domain. So I guess the question here is, when we pass a UFL expression to a diagnostic function, how can we ensure that its defined only on a single mesh. Right now, the diagnostics machinery extracts all Firedrake functions from a UFL expression and checks to see if it has a `FunctionContext` for them - the existence of this object treated as the 'is initialised' test. When a function from `split(z)` is passed in, it only sees the mixed function. So the question becomes how can the diagnostics object tell the difference between `as_vector((z[0],z[1],z[2]))`, which is just `u` and `as_vector((z[4],z[-1]))`, which is `[p,phi]` in a self-gravity simulation and makes no sense at all? 

I think the answer to this might be to use `split()` within the diagnostics object. If exactly one of the outputs of `split(z)` matches the UFL expression that was passed in, then we can know that what was passed in corresponds to a function on a not-mixed function space. I guess from that point we can check the domain and element to see if it corresponds to a function space we've seen before, and then treat it as if it were a Firedrake function from that point on.

dsroberts:
OK, I'm extremely overthinking this. Easiest thing to do is have `GeodynamicalDiagnostics` itself give you `u_horizontal`:
```
    def u_horizontal(self) -> Expr:
        return self.get_horizontal_components(self.u)
```
then have
```
u_horizontal = gd.u_horizontal()
```
and then everything will work as it should. Ideally it'd be the final values coming from `GeodynamicalDiagnostics`, but if we don't know exactly what we want yet, this will do for now. Also, is it worth adding one of the new diagnostics [here](https://github.com/g-adopt/g-adopt/blob/af1674e08288f1dd5dc12a744ae38b3b3ebdeb59/demos/test_all.py#L20)? Right now the actual values aren't being tested.

stephankramer:
Thanks yeah I'll do that. And thanks for pointing me at that test_all.py code, it should absolutely be testing for it indeed!

Probably going off-topic, but important to discuss it somewhere:

> The idea was that constructing a ufl expression is work that only needs to be done once, so even if it isn't much work, it should still only happen once. On the other hand, the vertical_component(f) and horizontal_components(f) expressions for a given f won't change whether they're called from inside the diagnostics or not, so perhaps the functions in utility.py should be cached. Given the results of _check_present and _check_dim_valid are cached, that'd remove the need to cache on the diagnostics object.

I honestly think this is overkill, constructing a form or expression is relatively trivial, it's just constructing a tree, there's no computation or analysis done. You're right that you probably shouldn't be doing it more than once for the same expression, but that's easily achieved by just storing the result: the user would typically do this on top of the time-loop, and inside some class you just do it once in the `__init__()`. It's just a pattern you have to follow in Firedrake - and even if you didn't follow it, it wouldn't make a tiny bit of difference. Constructing a cache for this just adds complexity that I don't we think we need. If you do it as a hidden module cache object, you run the risk of the user blowing that cache pretty easily - e.g. say you're doing adaptive meshing, every time you adapt you have to redefine all your functions on the new mesh and all ufl expressions have to be rebuilt from scratch (again not huge cost). So you would have to make an interface to clear that cache, and remember to use it.

drhodrid:
@dsroberts - are you able to look at the failures here and make any necessary tweaks to get this over the line?

dsroberts:
Yep, no problem. Looks like a bit of a mismatch in `div_u_hor_max_base`? I'll regenerate `expected.pkl` with the latest firedrake dev container.

drhodrid:
@stephankramer -- if you're happy with this, I think it's now good to go.

stephankramer:
thanks @dsroberts

PR #490, opened 2026-04-01
Swap order of solution assignment and solver call in StokesSolverBase.solve() to ensure the solution is updated correctly.

Patol75:
As the title explains. This showed up when using `animate`. The solver's solution can change, but if one is not careful, the old solution will not be updated. This is a problem when using the viscous free-surface formulation. Ultimately, the proposed fix here copies the order in which these calls are done in IrksomeIntegrator.advance().


sghelichkhani (changes requested):
Thanks @Patol75 for the change. Though the change is reasonable, I suspect it is one of those inconsistencies because the nature of the stokes. So for this to be complete we require a test for this, otherwise there is no problem this is actually solving.

sghelichkhani:
The failures are completely unrelated to this. There has been changes to petsc in the last two days which I think mostly affect mesh generation for Periodic meshes (I think!!). That causes numerical failures. 

Needs to be dealt with in a seperate PR.

Patol75:
@w3168 Yes, after a few time-loop iterations, the mesh changes, new function spaces and functions need to be defined, interpolation occurs, and then solvers are called again. If the equation is time-dependent, the old solution needs to be provided using current `main` because assignment occurs after the solver call. This is only the case for Stokes integrators.

@sghelichkhani Thank you for looking into it; great suggestions. I see you have triggered a CI run on `main`; we are expecting failures, right? I will add a test with `animate`. It might take me a few days, though.

w3168 (dismissed):
Looks good to me now the tests pass without changing the test values

PR #493, opened 2026-04-07
Update longtest github action

dsroberts:
Updates longtest github action to use `doit` system. Allows re-enabling of periodic longtest.


angus-g:
> I'm kind of hoping there is a nicer fix than `int(float(l))`.

`Constants` don't support `int()` directly. We use this approach in the cylindrical modules

PR #505, opened 2026-05-07
Build level set initialisation demo

angus-g:
For some reason this has maybe never been on the website? Even though it doesn't provide any testable output, it still has an `entrypoint` in its `meta.py` so we can ensure it runs before the render job.

- [x] Render job: https://github.com/g-adopt/g-adopt/actions/runs/25470191245

### DIFF
--- a/demos/.pages
+++ b/demos/.pages
@@ -29,6 +29,8 @@
     - [adjoint_ice](glacial_isostatic_adjustment/adjoint_ice.ipynb)
     - [adjoint_viscosity](glacial_isostatic_adjustment/adjoint_viscosity.ipynb)
 - Multi-material
+    - Initialisation
+        - [level_set_initialisation](multi_material/level_set_initialisation.ipynb)
     - Sources of buoyancy
         - [compositional_buoyancy](multi_material/compositional_buoyancy.ipynb)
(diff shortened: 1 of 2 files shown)


angus-g:
Unrelated failure...

```
=================================== FAILURES ===================================
___________________________ test_adjoint_taylortest ____________________________
    def test_adjoint_taylortest():
        with open(Path(__file__).parent.resolve() / "taylor_test_minconv.txt", "r") as f:
            tt_minconv = [float(x) for x in f.readlines()]
    
>       np.testing.assert_allclose(tt_minconv[-1], 2, rtol=3e-2)
E       AssertionError: 
E       Not equal to tolerance rtol=0.03, atol=0
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference among violations: 0.06233557
E       Max relative difference among violations: 0.03116778
E        ACTUAL: array(1.937664)
E        DESIRED: array(2)
demos/glacial_isostatic_adjustment/adjoint_ice/test_adjoint_ice.py:9: AssertionError
```

Patol75 (approved):
Thanks, Angus. I believe I indeed forgot to add it to the website. Sorry about that.

PR #506, opened 2026-05-11
Adjust tolerance for new horizontal diagnostics test.

stephankramer:
Increased tolerance to rtol=5e-4 (currently failing on main with rtol~=1.5e-4 see https://github.com/g-adopt/g-adopt/actions/runs/25663320129/job/75329858275)

Numerical derivatives are always a bit noisy, so some variation in min/max is to be expected. If we think this is too loose we can drop or change what we test for - although for a user I think min/max of divergence at a boundary is what makes most sense?

### DIFF
--- a/demos/test_all.py
+++ b/demos/test_all.py
@@ -18,7 +18,12 @@ cases = {
     f"{mc_path}/viscoplastic_case": {"extra_checks": ["nu_top"]},
     f"{mc_path}/2d_cylindrical": {"extra_checks": ["nu_top", "T_min", "T_max"]},
     f"{mc_path}/3d_spherical": {"extra_checks": [
-        "nu_top", "t_dev_avg", "u_hor_max", "div_u_hor_min_base", "div_u_hor_max_base"]},
+        "nu_top",
+        "t_dev_avg",
+        ("u_hor_max", {"rtol": 5e-4}),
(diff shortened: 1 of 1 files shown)


demos/test_all.py
```diff
@@ -17,8 +17,10 @@
     f"{mc_path}/2d_compressible_ALA": {"extra_checks": ["nu_top"]},
     f"{mc_path}/viscoplastic_case": {"extra_checks": ["nu_top"]},
     f"{mc_path}/2d_cylindrical": {"extra_checks": ["nu_top", "T_min", "T_max"]},
-    f"{mc_path}/3d_spherical": {"extra_checks": [
-        "nu_top", "t_dev_avg", "u_hor_max", "div_u_hor_min_base", "div_u_hor_max_base"]},
+    f"{mc_path}/3d_spherical": {
+        "extra_checks": ["nu_top", "t_dev_avg", "u_hor_max", "div_u_hor_min_base", "div_u_hor_max_base"],
```
drhodrid: @angus-g is there a way to set this up such that only the final 3 diagnostics have rtol=5e-4, with the others sticking at the default?

angus-g: Yes, pushed. Is that what you had in mind?

drhodrid: That's the one! Thank you.

drhodrid (approved):

PR #507, opened 2026-05-14
Infra: time_stepper API for stage-accurate time, dodo meta loader, primary_checks, CI label fix

sghelichkhani:
Three cross-cutting infrastructure changes the Richards equation work needs but that are useful on their own, plus a one-line CI workflow fix.

`gadopt/time_stepper.py` picks up an optional `t` kwarg on `IrksomeIntegrator` so callers can pre-build a `MeshConstant`, pass it in, and reference it from time-dependent BC or source expressions. Without this, multi-stage methods see forcing frozen at $t_n$ and their formal order collapses to one. `MeshConstant` is re-exported from `gadopt` so users don't have to import it from Irksome. `RKGeneric` learns to accept `stage_type` and `bc_type` overrides through kwargs, which is what a mass-conservative Richards solver needs to inject `stage_type="value"` regardless of the tableau default. `create_custom_tableau` coerces its arrays to float64 so newer Irksome `ButcherTableau` objects accept them.

`dodo.py` loads each `meta.py` under a uuid-suffixed package name so relative imports resolve to the meta's own directory without sibling tests' equivalently named submodules colliding in `sys.modules`. A finally clause evicts all keys with that prefix.

`demos/generate_expected.py` learns a `primary_checks` key in case configs, defaulting to the existing `u_rms` behaviour. Demos with their own scalar diagnostics (groundwater needs `min_h`, `max_h`) can declare their primary columns explicitly.

`.github/workflows/test.yml`: line 84 now mirrors line 50's existing `case(...)` expression so the benchmark regression test honors the `container:main` and `container:release` labels, just like the smoke test already does. The label-aware logic was added to the smoke test in #479 but the benchmark test was overlooked at the time. Without this the upper layers of any stacked PR series can't run benchmark CI against a real container.

This is the bottom of a six-PR stack adding the Richards equation. The full stack lands the new module, its tests, two demos, parallel-scaling baselines, and non-stiffly-accurate verification at the tip.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -81,7 +81,7 @@ jobs:
     runs-on: self-hosted
     container:
       # Prefer specific branch, then PR base, then actual branch
-      image: firedrakeproject/firedrake-vanilla-default:dev-${{ inputs.target_container || github.base_ref || github.ref_name }}
+      image: firedrakeproject/firedrake-vanilla-default:dev-${{ inputs.target_container || (github.event_name == 'pull_request' && case(contains(github.event.pull_request.labels.*.name, 'container:main'), 'main', contains(github.event.pull_request.labels.*.name, 'container:release'), 'release', github.base_ref)) || github.ref_name }}
       options: --cap-add SYS_NICE
 
     concurrency:
--- a/demos/generate_expected.py
+++ b/demos/generate_expected.py
@@ -17,12 +17,15 @@ if __name__ == "__main__":
 
     for case in requested_cases:
         try:
-            extra_checks = cases[case].get("extra_checks", [])
+            case_config = cases[case]
         except KeyError:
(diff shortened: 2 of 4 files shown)


sghelichkhani:
@angus-g I know this has the one-liner workflow change that you might want to change. But obviously everything is up for change.

angus-g (approved):
I think whatever you said in Zulip doesn't really correspond to the actual content of the PR: looks like you just ported the container logic into the smoke test job, which definitely makes sense! I think the change to the kwargs handling in `RKGeneric` could be maybe a little more Pythony (using a dictionary update instead of popping), but it's fine:

```python
kwargs = {
  "stage_type": self.stage_type,
  "bc_type": self.bc_type",
} | kwargs
```

sghelichkhani:
Thanks @angus-g ! The comment re style is great! Thanks.

PR #508, opened 2026-05-14
MONSTER: Add Richards equation module for variably saturated flow

sghelichkhani:
# A monster branch that includes #521 #522 #523 #524 #525 

Introduces the Richards equation for variably saturated flow in porous media. In mixed form $(S_s S + C) \partial_t h + \nabla \cdot (K \nabla h) + \nabla \cdot (K \nabla z) = 0$, with strong Dirichlet on CG spaces and weak SIPG Dirichlet on DG spaces, following the same `Equation`/term pattern used by the existing G-ADOPT scalar transport solvers.

The mass term is written as `Dt(theta(h))` (composite time derivative) and the solver defaults to `stage_type="value"`, so Irksome discretises it as the exact finite difference $\theta(h_{n+1}) - \theta(h_n)$ per stage. Stiffly-accurate tableaux conserve mass via the trivial $u_{new} = U_s$ update; non-stiffly-accurate tableaux conserve via Irksome's conservative variational update path (which needs Irksome at or after the StageValueStepper conservative-update fix; that surface is gated to the top of the stack). Adaptive time stepping falls back to `stage_type="deriv"` because Irksome doesn't yet support adaptive with stage_value.

`soil_curves.py` packages three water-retention and conductivity models commonly used in groundwater hydrology: Haverkamp, van Genuchten, and an exponential form. Each provides `moisture_content(h)`, `relative_permeability(h)`, and `water_retention(h)` as UFL expressions.

`preconditioners.py` adds `VerticallyLumpedPC` and `VerticallyLumpedHMGPC`, two-level multigrid preconditioners that exploit the strong vertical anisotropy of typical infiltration problems. The vlumping presets are exposed alongside the existing direct and iterative parameter dictionaries.

`utility.py` adds `RichardsTimestepAdaptor`, a controller combining advective and diffusive CFL ceilings with SNES-iteration-count growth modulation. It only proposes timesteps; the caller still owns the solve loop.

`tests/unit/test_soil_curves.py` covers monotonicity, saturation limits, and analytical sanity checks for the three soil curves.

Sits on top of the time-stepper API additions in the layer below.

### DIFF
--- a/gadopt/__init__.py
+++ b/gadopt/__init__.py
@@ -22,7 +22,23 @@ from .level_set_tools import (
 )
 from .limiter import VertexBasedP1DGLimiter
 from .nullspaces import create_stokes_nullspace, rigid_body_modes
-from .preconditioners import FreeSurfaceMassInvPC, SPDAssembledPC
+from .preconditioners import (
+    FreeSurfaceMassInvPC, SPDAssembledPC,
+    VerticallyLumpedPC, VerticallyLumpedHMGPC,
+)
+from .richards_solver import (
+    RichardsSolver,
+    direct_richards_solver_parameters,
+    iterative_richards_solver_parameters,
+    vlumping_richards_solver_parameters,
+    vlumping_hmg_richards_solver_parameters,
+)
+from .soil_curves import (
+    SoilCurve,
+    HaverkampCurve,
+    VanGenuchtenCurve,
+    ExponentialCurve,
+)
 from .solver_options_manager import DeleteParam
 from .stokes_integrators import (
     BoundaryNormalStressSolver,
@@ -37,6 +53,7 @@ from .utility import (
     InteriorBC,
     LayerAveraging,
     ParameterLog,
+    RichardsTimestepAdaptor,
     TimestepAdaptor,
     interpolate_1d_profile,
     log,
--- a/gadopt/equations.py
+++ b/gadopt/equations.py
@@ -13,15 +13,10 @@ from typing import Any, Callable
 from warnings import warn
 
 import firedrake as fd
-import ufl
-from ufl.algorithms.analysis import extract_coefficients
 from ufl.indexed import Indexed
 
 from .approximations import BaseApproximation, BaseGIAApproximation
 from .utility import CombinedSurfaceMeasure
-
-# Equation-level eq_attrs consumed by Equation itself, not by any term.
-_EQUATION_RESERVED_ATTRS = frozenset({"solution", "nonlinear_coefficients"})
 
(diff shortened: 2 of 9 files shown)


w3168 (commented):
Hi @sghelichkhani and @MorrowL thanks for the PR. I have had a first look through. On the whole it looks really good - well done! 

 I know you have done a lot of work already to split up the PRs to make reviewing easier (thanks!), so I hope you do not find this too annoying, but...  Could you make this PR a bit simpler?

I would suggest moving adaptive time stepping, the vertically lumped preconditioners and possibly also the majority of the soil curves to their own respective PRs. Particularly the vertically lumped preconditioners should be reviewed by @stephankramer.

In this PR the main focus should be on the equations, solver and a test of those equations. At the moment your tests are just of the soil curves so we have to wait for a different PR to check that the actual timestepping is correct... 

I also wonder if it would be better to inherit the diffusion and source term from `scalar_equation.py`. Occasionally there are changes to the interior penalty code, so it would be nice to avoid changing this in multiple places...

On zulip do you mind providing a link to your preprint as well? It would be good to compare how your code aligns with your maths.

sghelichkhani:
Thanks Will. I've pushed the diffusion/source reuse. `richards_diffusion_term` and `richards_source_term` are gone; `RichardsSolver` now goes through `scalar_equation.diffusion_term` and `scalar_equation.source_term`. The one wrinkle was that K(h) is solution-dependent, so the scalar diffusion term picks up an optional `diffusivity_fn` attribute (a callable of the trial evaluated against each Irksome stage); existing callers that pass a fixed `diffusivity` are untouched, the term XOR-validates the two. The penalty in scalar uses shift=-1 vs the shift=0 the bespoke term had, so I've bumped Richards' default `interior_penalty` from 2.0 to 4.0 to keep the coercivity margin. Mass and gravity terms stay Richards-specific since they have no clean analogue in scalar_equation. 

Next I'll split the remaining bits: adaptive time stepping, the vertically lumped preconditioners (so Stephan can review them in isolation), and the soil curves, into their own PRs so this one comes back down to just the equation, the solver, and a real end-to-end test (Tracy 2D rather than only the soil-curve unit tests).

sghelichkhani:
Superseded by the decomposed stack (#522 through #511). Everything this branch carried, the equation module, soil curves, core solver, vertically-lumped preconditioners, the adaptive timestep controller and the soil-curve unit tests, is reproduced byte-for-byte across richards-core-02 through -05, which then add the benchmark and demo tests on top. Nothing depends on this branch any more, so closing it and deleting the branch to keep the stack clean.

PR #510, opened 2026-05-14
Vauclin 2D groundwater demo

sghelichkhani:
Adds the Vauclin (1979) water-table recharge benchmark as a g-adopt demo and wires it into the demo regression harness. The 3x2 m soil column starts with a water table at $z=0.65$ m, top infiltration of 14.8 cm/hr drives a wetting front downward, and the analytical reference from `gwassess.VauclinRichardsSolution2D` is overlaid for visual comparison. Haverkamp soil, DIRK22 with stage-accurate top-flux forcing through a shared `MeshConstant`, and `RichardsTimestepAdaptor` managing dt growth so the demo runs to $t=8$ h in a sensible wall-clock budget.

`demos/test_all.py` learns about the groundwater path and registers the case with `primary_checks=["min_h", "max_h"]` instead of `u_rms`, because head, not RMS velocity, is the meaningful scalar diagnostic for an infiltration problem. `mass_balance` is an extra check that catches drift in the global moisture budget. The `check_series` and `test_benchmark` plumbing forwards a `primary_checks` kwarg from the case config so the existing `u_rms` default stays the right answer for the Stokes/GIA benchmarks.

`pyproject.toml` adds `gwassess` to the `demos` extras (the demo imports it for the analytical overlay). `expected.pkl` is generated with `demos/generate_expected.py` against the current solver and pins `min_h`, `max_h`, and `mass_balance` at the resolution the demo runs at.

Sits on top of the Richards SA test suite.


MorrowL:
So on my end, it seems to run fine without the small perturbation to the IC. That is, it seems to convergence using:

water_table_height = vauclin_solution.WATER_TABLE_HEIGHT
h_ic = Function(V, name="InitialCondition").interpolate(
    water_table_height - X[1]
)

MorrowL:
I'm noticing that the dt the adaptive timestepper chooses seems to oscillate up and down a bit. I don't know if this is intentional but I would imagine it should start small and gradually increase over the simulation

demos/groundwater/2d_vauclin/2d_vauclin.py
```diff
@@ -0,0 +1,600 @@
...
+# The initial pressure head is the hydrostatic profile
+# $h(x, z, 0) = 0.65 - 1.001\, z$. The factor $1.001$, rather than exactly unity,
+# ensures that the upper portion of the domain starts marginally dry
+# rather than at the saturation boundary; this avoids an artificial
+# singularity in $C(h)$ at $t = 0$. Two Firedrake ``Function`` objects
+# are allocated: $h$ for the current solution, and $h_{\mathrm{old}}$
+# as a buffer for the pre-step state that is used by the retry
+# protocol.
+
+water_table_height = vauclin_solution.WATER_TABLE_HEIGHT
```
MorrowL, line 357: So here I think it's ok to not need the 'perturbed' IC anymore:
water_table_height - X[1]

demos/groundwater/2d_vauclin/2d_vauclin.py
```diff
@@ -0,0 +1,600 @@
...
+)
+
+# The adaptive time-step controller is configured with a lower bound of
+# $0.1$ s, an upper bound of $300$ s, and a large value of
+# ``minimum_diffusive_dt`` ($10^6$ s) that effectively disables the
+# diffusive tier, as discussed in the preceding section. A two-step
+# cool-down following each retry suppresses premature regrowth of
+# $\Delta t$.
+
+adaptor = RichardsTimestepAdaptor(
```
MorrowL, line 427: I'm finding that the timestep seems to be bouncing dt = 15 and 60 seconds

MorrowL (commented):
Thoughts on Vauclin demo

PR #511, opened 2026-05-14
Cockett 3D demo and Richards parallel-scaling test suite

sghelichkhani:
3D Cockett (2018) heterogeneous infiltration as both a g-adopt demo and the basis of a longtest-only weak-scaling preconditioner comparison. The demo itself is a single-node configuration that runs as part of the regular demo regression; the heavyweight multi-node sweep lives in `tests/parallel_scaling_richards`.

`demos/groundwater/3d_cockett` wraps the benchmark in the same `primary_checks=["min_h", "max_h"]` convention as the Vauclin demo, with its own `expected.pkl`. `demos/test_all.py` picks up the case registration.

`tests/parallel_scaling_richards` is the longtest-only HPC harness. `cockett_3d.py` and `murrumbidgee_3d.py` are CLI-driven (--nx, --nz, --solver, --steps) so a single script can be relaunched at multiple mesh sizes and preconditioner presets without recompilation. `richards_scaling.py` parses PETSc `-log_view` output to extract per-stage timings, KSP/SNES iteration counts, and assembly cost. The `solvers/` presets cover GMG, GAMG, block-Jacobi, and SOR for direct comparison at the same wall-clock budget. `expected.csv` pins the non-Gadi-specific solver iteration counts and `gadi_expected.csv` pins the Sapphire-Rapids timing baseline; tests allow 20% drift on timings (well above observed Gadi-to-Gadi variance) and exact match on iteration counts. All assertion-side tests carry `pytest.mark.longtest` so only the Gadi longtest workflow runs them; the doit `run_case_hpc` pipeline runs the cases via `gadopt_hpcrun` on the matching node counts before pytest gets there.

Sits on top of the Vauclin demo.


demos/groundwater/3d_cockett/3d_cockett.py
```diff
@@ -0,0 +1,154 @@
...
+Lx, Ly, Lz = 2.0, 2.0, 2.6
+nx, ny, nz = 20, 20, 26
+
+mesh2d = RectangleMesh(nx, ny, Lx, Ly, quadrilateral=True)
+mesh = ExtrudedMesh(mesh2d, nz, layer_height=Lz / nz)
+mesh.cartesian = True
+X = SpatialCoordinate(mesh)
+
+# We use a discontinuous Galerkin space of polynomial degree 0 (piecewise
+# constant) on hexahedral elements, giving 10,400 degrees of freedom.
```
MorrowL, line 44: Maybe also add that DG0 and a coarse mesh s chosen for this demo for computational efficiency

demos/groundwater/3d_cockett/3d_cockett.py
```diff
@@ -0,0 +1,154 @@
...
+dt = Constant(600.0)
+t_final = 24 * 3600  # 24 hours in seconds
+
+richards_solver = RichardsSolver(
+    h,
+    soil_curve,
+    delta_t=dt,
+    timestepper=DIRK22,
+    bcs=richards_bcs,
+    solver_parameters="direct",
```
MorrowL, line 116: Why not iterative?

MorrowL (commented):
Thoughts on Cockett demo

PR #509, opened 2026-05-14
Richards tests: mass balance, temporal convergence and Vauclin reference

sghelichkhani:
Adds the Richards solver test surface on top of the core solver and its Tracy benchmark (#523), covering the time-stepping schemes the solver supports (BackwardEuler, ImplicitMidpoint, GaussLegendre(2)).

`test_mass_balance.py` checks discrete mass conservation along two axes: function space at fixed BackwardEuler (DQ0/DQ1/DQ2 sit at solver tolerance, CG1/CG2 pick up an O(1e-4) consistency error from lacking element-wise conservation), and time-stepping scheme at fixed DQ1 (all three schemes conserve mass to solver tolerance).

`test_temporal_convergence.py` verifies each scheme delivers its formal order (1, 2 and 4) via a spatially-uniform manufactured solution that makes diffusion and gravity vanish and isolates the temporal truncation.

`vauclin_2d.py` is a reference module for the Vauclin (1979) water-table recharge benchmark with Haverkamp soil, DIRK22 and stage-accurate top-flux forcing. The 2D-Vauclin demo (#510) imports its model function.

Sits on top of #525, which carries the core solver, the Tracy convergence benchmark, the vertically-lumped preconditioner tests and the adaptive timestep controller.

PR #512, opened 2026-05-14
Non-SA Richards verification: temporal convergence MMS, tableau mass balance, Irksome pin

sghelichkhani:
Two pytest entries that exercise the non-stiffly-accurate `stage_value` path of the Richards solver, plus the dependency pin they need to run cleanly.

`test_temporal_convergence.py` drives a manufactured solution that is spatially uniform on the unit square with Dirichlet $h_0(t)$ on all sides and a matching source term, so diffusion and gravity contribute identically zero and the PDE reduces pointwise to $\partial_t \theta = S$. The error at $t_{\text{final}}$ isolates temporal truncation alone. Parametrize covers BackwardEuler (order 1), ImplicitMidpoint (order 2), and GaussLegendre(2) (order 4); the coarsest-pair log2 rate is asserted against 85% of the formal order.

`test_mass_balance.py` picks up its second test, `test_mass_balance_tableaux`, which fixes function space at DQ1 and varies the tableau (BE, IM, GL2). On an Irksome that ships the conservative variational update for non-SA `stage_value`, all three sit at solver tolerance; without that update path the IM and GL2 cases pick up an O(1) mass defect.

`pyproject.toml` pins `irksome` at upstream commit `4820795` (Conservative StageValueStepper, PR #204) so the firedrake-vanilla-default container CI has the conservative-update fix at the version this commit needs.

**DO NOT MERGE this top layer until the pin is no longer required.** The pin is short-term: once both `firedrake-vanilla-default` and the fp50 Gadi module ship Irksome at or after `4820795`, drop the pin and the comment block in `pyproject.toml`. The five layers below this one in the stack do not depend on the pin and can be merged independently.

Top of the six-PR Richards stack.


sghelichkhani:
Folding this into #509. Now that the conservative-update path is in upstream Irksome and the pin is gone, there is no reason to keep the non-SA verification on a separate branch, so the tableau mass-balance check and the temporal-convergence test now live in #509 alongside the rest of the Richards tests, exercised across BackwardEuler, ImplicitMidpoint and GaussLegendre(2) without distinguishing stiffly-accurate from non-stiffly-accurate. Closing this one.

PR #513, opened 2026-05-15
Add radial velocity diagnostic

drhodrid:
**Summary**
This PR adds a new convenience diagnostic, GeodynamicalDiagnostics.u_radial(), for accessing the radial/upward component of the velocity field in a way that is consistent with the existing GeodynamicalDiagnostics.u_horizontal() helper.

The motivation is to make it easier and less error-prone to output and analyse radial velocity diagnostics in spherical models, particularly when comparing the full velocity magnitude with its radial and tangential components. The new helper mirrors the existing horizontal-velocity diagnostic and uses the same underlying upward-normal machinery already present in G-ADOPT (important for consistency between radial and tangential components).

**Changes made**

1. Added u_radial() to GeodynamicalDiagnostics
This PR adds a new u_radial() method to gadopt/diagnostics.py. It returns the component of velocity parallel to the upward normal by calling the existing upward-component helper.

This is analogous to the existing u_horizontal() helper, which returns velocity components orthogonal to the upward normal.

In spherical or annular geometries, u_radial() corresponds to the radial velocity. In Cartesian geometries, it corresponds to the vertical/upward velocity component.

2. Updated the 3D spherical mantle convection demo
The 3D spherical demo now uses the new diagnostic and logs radial velocity extrema alongside the existing horizontal-velocity diagnostics.

Specifically, the demo now records u_radial_min and u_radial_max, and the parameter-log header has been updated so that these new columns are clearly labelled.

@dsroberts or @angus-g could you update the expected values for these params in the test log -- I ran out of time for that. 

3. Added unit tests
This PR adds tests in tests/unit/test_diags.py to check that:

1. u_radial() is consistent with the existing upward-component diagnostic machinery.
2. u_radial() and u_horizontal() decompose the velocity magnitude consistently.

The main consistency check is that the full velocity magnitude decomposes into radial and horizontal components as expected: the squared velocity magnitude equals the squared radial component plus the squared horizontal magnitude. This ensures that the radial and horizontal velocity components are defined relative to the same upward normal.

**Testing**

I tested the changes locally in a Firedrake environment.

The focused diagnostics test suite passes:

109 passed in 25.96s

I also ran flake8 on the files changed in this PR and fixed the whitespace/style issues it flagged.

**Note on AI assistance**

This is my first PR prepared with AI support, using GPT-5.5. I used it largely as a test of whether AI assistance could help with a small, well-scoped contribution: identifying the right place for the convenience function, checking consistency with the existing u_horizontal() implementation, drafting tests, and checking the workflow before committing.

It was very useful for navigating the code structure and keeping the PR focused. I reviewed and tested the changes locally before opening the PR.


demos/mantle_convection/3d_spherical/3d_spherical.py
```diff
@@ -126,10 +126,15 @@
 output_frequency = 1
 
 plog = ParameterLog('params.log', mesh)
-plog.log_str("timestep time dt maxchange u_rms nu_top nu_base energy avg_t t_dev_avg "
-             "u_hor_max div_u_hor_min_base div_u_hor_max_base")
+plog.log_str(
+    "timestep time dt maxchange u_rms "
+    "Nu_top Nu_bottom energy_conservation T_avg T_dev_avg "
```
drhodrid, line 131: @angus-g -- the names I added here are a bit more explicit. Should we change back or update the test fields (this is why it's currently failing I think!).

dsroberts: I'm pretty sure the failure is due to the expected values not being updated yet.

angus-g: Updating the names is fine, but the corresponding names need to go in `test_all.py`.

drhodrid:
Thanks @angus-g !

tests/unit/test_diags.py
```diff
@@ -234,6 +234,68 @@ def test_function_extraction():
...
+
+    assert error < 1.0e-12
+
+
+def test_u_radial_and_u_horizontal_decompose_velocity():
+    """
+    Test that the radial and horizontal velocity components decompose the full
+    velocity magnitude consistently.
+    """
+    mesh = get_mesh("annulus")
```
dsroberts:
```suggestion
@pytest.mark.parametrize(
    "mesh_type,top_bdy,bottom_bdy", [("square", 2, 1), ("annulus", "top", "bottom")]
)
def test_u_radial_and_u_horizontal_decompose_velocity(
    mesh_type: Literal["square", "annulus"],
    top_bdy: int | str,
    bottom_bdy: int | str,
):
    """
    Test that the radial and horizontal velocity components decompose the full
    velocity magnitude consistently.
    """
    mesh = get_mesh(mesh_type)
```
Given we're testing for the correctness of a diagnostic (as opposed functionality of the diagnostic module itself), we should make sure it works on cartesian meshes too.

tests/unit/test_diags.py
```diff
@@ -234,6 +234,68 @@ def test_function_extraction():
...
+    Z = fd.MixedFunctionSpace([V, Q])
+
+    z = fd.Function(Z)
+    u, _ = z.subfunctions
+
+    X, Y = fd.SpatialCoordinate(mesh)
+    u.interpolate(fd.as_vector([X + Y, 2.0 * Y - X]))
+
+    T = fd.Function(Q)
+    diags = gadopt.GeodynamicalDiagnostics(z, T, "bottom", "top")
```
dsroberts:
```suggestion
    diags = gadopt.GeodynamicalDiagnostics(z, None, bottom_bdy, top_bdy)
```
Update boundary values to match change in function args above. Also, `None` is a valid value for `T` here, no need to initialise a function that's never used.

demos/mantle_convection/3d_spherical/3d_spherical.py
```diff
@@ -192,6 +217,7 @@
     plog.log_str(f"{timestep} {time} {float(delta_t)} {maxchange} {gd.u_rms()} "
                  f"{nusselt_number_top} {nusselt_number_base} "
                  f"{energy_conservation} {gd.T_avg()} {T_dev_avg} "
+                 f"{u_radial_min} {u_radial_max} "
```
dsroberts, line 220: Should these new fields be added to `extra_checks` in `test_all.py`?

tests/unit/test_diags.py
```diff
@@ -234,6 +234,68 @@ def test_function_extraction():
...
+
+    radial = diags.u_radial()
+    expected = diags.get_upward_component(u)
```
dsroberts:
```suggestion
    u_test, _ = fd.split(z)

    diags = gadopt.GeodynamicalDiagnostics(z, None, bottom_bdy, top_bdy)

    radial = diags.u_radial()
    expected = gadopt.utility.vertical_component(u_test)
```

tests/unit/test_diags.py
```diff
@@ -234,6 +234,68 @@ def test_function_extraction():
     assert gadopt.diagnostics.extract_functions(vc) == {f}
 
 
+def test_u_radial_matches_upward_component():
+    """
+    Test that GeodynamicalDiagnostics.u_radial returns the upward/radial
+    component of the velocity field.
+    """
+    mesh = get_mesh("annulus")
```
dsroberts:
```suggestion
@pytest.mark.parametrize(
    "mesh_type,top_bdy,bottom_bdy", [("square", 2, 1), ("annulus", "top", "bottom")]
)
def test_u_radial_matches_upward_component(
    mesh_type: Literal["square", "annulus"],
    top_bdy: int | str,
    bottom_bdy: int | str,
):
    """
    Test that GeodynamicalDiagnostics.u_radial returns the upward/radial
    component of the velocity field.
    """
    mesh = get_mesh(mesh_type)
```
I feel this test as it stands is redundant. `u_radial` literally returns `upward_component(u)`. We can turn this into a more useful test by comparing `u_radial()` and the result of `gadopt.utility.vertical_component(fd.split(z)[0])`. `split` returns a list tensor expression created solely from `z` whose components should be identical to `z.subfunctions`.. This is one of the things we were having issues with in the original horizontal_component PR. Its a vastly different code path that should have the same numerical result. This updated test would also make sense to run on the cartesian mesh as well.

dsroberts (commented):
The robot has done OK. I think it's been a little shallow when it comes to the tests. To be fair, I'm sure you didn't set it loose on the Firedrake code base, so it wouldn't have known about `fd.split()`. I do feel like it could have picked up the other stuff though.

stephankramer:
Just to comment on the name, is `u_radial` really the best name for that? You're now describing it as "radial component of velocity, but oh if you're in Cartesian domain it's the z-component". I just think it might be confusing if in a Cartesian setup you see `gd.u_radial()` being used. Names should be immediately obvious, so I'd much prefer it using upward or vertical in the name, Of course doesn't stop you in a spherical demo to say `u_radial = gd.u_vertical()`.
That is, unless geodynamicsts are much more likely to always call the vertical component radial even when discussing a small scale Cartesian model, but I don't think that's the case - in fact I'd say the opposite, you frequently see $$\hat k$$ being used even in spherical models, even though $$\hat k$$ is really the z unit vector.

drhodrid:
Thanks both. I have renamed the general diagnostic to u_vertical(), while keeping the local/output name u_radial in the spherical demo where that interpretation is appropriate. I also updated the diagnostics tests so they cover both Cartesian square and annulus meshes, and added u_radial_min/u_radial_max to the 3d_spherical extra_checks. The targeted 3d_spherical demo test now passes locally.

dsroberts (approved):
LGTM. Thanks for sorting the additional testing

PR #515, opened 2026-05-16
Infrastructure plumbing for the gplates indicator/geotherm stack

sghelichkhani:
This is the base of the gtrack stack: the small plumbing that lets the later PRs import and configure cleanly. Nothing here is behavioural on its own.

The file-path helper in `gplatesfiles.py` now accepts either a single string or a list for each reconstruction entry. A bare string used to be treated as a sequence of characters, which is the bug Daniel flagged; it now normalises a lone string to a one-element list and carries the continental and static polygon defaults. `log()` gains an optional `level=` kwarg so callers can gate verbose output on `GADOPT_LOGLEVEL`.

On Angus's point about the dependency: rather than adding `gtrack` to the hard dependencies (which would force pygplates onto every install), it goes in as an optional `gplates` extra. A plain `pip install g-adopt` stays lean, and `pip install g-adopt[gplates]` activates the plate stack. The regression CI job installs that extra alongside pygplates, and there's the small `gadopt/gplates/__init__.py` flake8 ignore.

Note for reviewers: this stack was rebuilt on top of the Source/Output redesign, so the diff you see now is the current shape rather than the original one. The five PRs still compose to the same gtrack feature as before.

### DIFF
--- a/gadopt/gplates/gplatesfiles.py
+++ b/gadopt/gplates/gplatesfiles.py
@@ -9,7 +9,9 @@ _default_muller2022_plate_files = {
         "1000-410-Divergence.gpml",
         "1000-410-Topologies.gpml",
         "1000-410-Transforms.gpml"
-    ]
+    ],
+    "continental_polygons": "shapes_continents.gpml",
+    "static_polygons": "shapes_static_polygons_Merdith_et_al.gpml",
 }
 
 reconstructions = {
@@ -53,17 +55,29 @@ reconstructions = {
 
 
 def check_and_get_absolute_paths(base_path: Path, filenames: dict):
+    # Normalize all values to lists
+    def to_list(value):
+        return value if isinstance(value, list) else [value]
+
     # Check if all files are present
-    all_files_present = all((base_path / filename).exists() for files in filenames.values() for filename in files)
+    all_files_present = all(
+        (base_path / filename).exists()
+        for files in filenames.values()
+        for filename in to_list(files)
+    )
 
     if not all_files_present:
         raise FileNotFoundError("Some files are missing. Cannot proceed without downloading the required files.")
 
     # Return absolute paths of the files
-    return {
-        key: [str(base_path / filename) for filename in files]
-        for key, files in filenames.items()
-    }
+    # Keep single values as single strings, not lists
+    result = {}
+    for key, files in filenames.items():
+        if isinstance(files, list):
+            result[key] = [str(base_path / filename) for filename in files]
+        else:
+            result[key] = str(base_path / files)
+    return result
 
 
 def ensure_reconstruction(reconstruction: str, base_path: str | Path):
--- a/gadopt/gplates/gplatesfiles.py
+++ b/gadopt/gplates/gplatesfiles.py
@@ -75,15 +75,11 @@ def check_and_get_absolute_paths(base_path: Path, filenames: dict):
     if not all_files_present:
         raise FileNotFoundError("Some files are missing. Cannot proceed without downloading the required files.")
 
-    # Return absolute paths of the files
-    # Keep single values as single strings, not lists
-    result = {}
-    for key, files in filenames.items():
-        if isinstance(files, list):
-            result[key] = [str(base_path / filename) for filename in files]
-        else:
-            result[key] = str(base_path / files)
-    return result
+    # Return absolute paths of the files, normalized to lists
+    return {
+        key: [str(base_path / filename) for filename in to_list(files)]
+        for key, files in filenames.items()
+    }
 
 
 def ensure_reconstruction(reconstruction: str, base_path: str | Path):
--- a/gadopt/gplates/gplatesfiles.py
+++ b/gadopt/gplates/gplatesfiles.py
@@ -1,3 +1,4 @@
+from collections.abc import Sequence
 from pathlib import Path
 
 _default_muller2022_plate_files = {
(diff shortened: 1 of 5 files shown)


angus-g:
Two things here. Not a huge deal, but potential future footgun is that files must be lists and not generic iterables. I think more important is that this gives an explicit dependency on `pygplates`. I *think* I'd prefer it as an optional dependency still, but we can have a dependency set in the pyproject rather than requiring the package to be explicitly named (as we're currently doing for `pygplates`).

```diff
diff --git a/gadopt/gplates/gplatesfiles.py b/gadopt/gplates/gplatesfiles.py
index dc59c22772..34a5c0fdd1 100644
--- a/gadopt/gplates/gplatesfiles.py
+++ b/gadopt/gplates/gplatesfiles.py
@@ -56,6 +56,8 @@
 
 def check_and_get_absolute_paths(base_path: Path, filenames: dict):
     # Normalize all values to lists
+    # XXX note that this will mean that tuples will error (not a huge
+    # deal, but potentially API-narrowing)
     def to_list(value):
         return value if isinstance(value, list) else [value]
 
diff --git a/pyproject.toml b/pyproject.toml
index b1410fe2a5..b8f5b4a341 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@
   "shapely",
   "vtk",
   "irksome @ git+https://github.com/firedrakeproject/Irksome",
-  "gtrack",
+  "gtrack", # XXX we do now expliictly depend on gplates for *all* installations with this...
 ]
 
 [project.optional-dependencies]
```

gadopt/gplates/gplatesfiles.py
```diff
@@ -53,17 +55,29 @@
 
 
 def check_and_get_absolute_paths(base_path: Path, filenames: dict):
+    # Normalize all values to lists
+    def to_list(value):
+        return value if isinstance(value, list) else [value]
```
dsroberts: Yeah, I fall over this one all the time. You need to do something like
```
if isinstance(value, str):
  return [value]
elif isinstance(value, Sequence):
  return list(value)
else:
  return [ value ]
```
Strings are sequences, so have to be dealt with first, then you can test for all other sequences, then finally other scalar types.

sghelichkhani:
I've rebuilt this branch on top of the Source/Output redesign and current main, so the diff is the current shape now.

Angus, on the dependency: It's now an optional `gplates` extra instead, so a plain install stays lean and `pip install g-adopt[gplates]` pulls in the plate stack.

dsroberts, the `to_list` string handling is fixed the way you suggested

sghelichkhani:
> Two things here. Not a huge deal, but potential future footgun is that files must be lists and not generic iterables. I think more important is that this gives an explicit dependency on `pygplates`. I _think_ I'd prefer it as an optional dependency still, but we can have a dependency set in the pyproject rather than requiring the package to be explicitly named (as we're currently doing for `pygplates`).
> 
> ```diff
> diff --git a/gadopt/gplates/gplatesfiles.py b/gadopt/gplates/gplatesfiles.py
> index dc59c22772..34a5c0fdd1 100644
> --- a/gadopt/gplates/gplatesfiles.py
> +++ b/gadopt/gplates/gplatesfiles.py
> @@ -56,6 +56,8 @@
>  
>  def check_and_get_absolute_paths(base_path: Path, filenames: dict):
>      # Normalize all values to lists
> +    # XXX note that this will mean that tuples will error (not a huge
> +    # deal, but potentially API-narrowing)
>      def to_list(value):
>          return value if isinstance(value, list) else [value]
>  
> diff --git a/pyproject.toml b/pyproject.toml
> index b1410fe2a5..b8f5b4a341 100644
> --- a/pyproject.toml
> +++ b/pyproject.toml
> @@ -21,7 +21,7 @@
>    "shapely",
>    "vtk",
>    "irksome @ git+https://github.com/firedrakeproject/Irksome",
> -  "gtrack",
> +  "gtrack", # XXX we do now expliictly depend on gplates for *all* installations with this...
>  ]
>  
>  [project.optional-dependencies]
> ```

Thanks @angus-g . I think I have addressed your issues in the new commit.

gadopt/gplates/gplatesfiles.py
```diff
@@ -53,17 +56,34 @@
...
 
 def check_and_get_absolute_paths(base_path: Path, filenames: dict):
+    # Normalize all values to lists
+    def to_list(value):
+        if isinstance(value, str):
+            return [value]
+        elif isinstance(value, Sequence):
+            return list(value)
+        else:
+            return [value]
```
sghelichkhani, line 66: @dsroberts @angus-g This is done now.

gadopt/gplates/gplatesfiles.py
```diff
@@ -53,15 +56,28 @@
...
+    )
 
     if not all_files_present:
         raise FileNotFoundError("Some files are missing. Cannot proceed without downloading the required files.")
 
-    # Return absolute paths of the files
+    # Return absolute paths of the files, normalized to lists
     return {
-        key: [str(base_path / filename) for filename in files]
+        key: [str(base_path / filename) for filename in to_list(files)]
```
sghelichkhani, line 80: I used the same conversion before apply here too!

dsroberts:
If we're going to have a `gplates` set of optional dependencies, should `pygplates` be in there?

sghelichkhani:
> If we're going to have a `gplates` set of optional dependencies, should `pygplates` be in there?

Good question! My initial thinking was no, because `pygplates` is a dependency of `gtrack` anyways, and hopefully down the line, we remove any direct dependency of pygplates out of gadopt and put it directly in gtrack. 

However ... adding pygplates does not add any cost ... and it just makes everything transparent! So VERY OPEN to it!

dsroberts (approved):

angus-g (approved):
> we remove any direct dependency of pygplates out of gadopt and put it directly in gtrack

Yeah, I think this would be the point at which we remove the direct dependency.

So agreed with @dsroberts that this is good to go now.

PR #366, opened 2026-01-29
MONSTER: Add LithosphereConnector for time-dependent lithosphere indicator fields

sghelichkhani:
# A Monster branch for PRs #514 #515 #516 #517 #518 #519

## Summary

Adds lithosphere indicator field capabilities to G-ADOPT by integrating with the [gtrack](https://pypi.org/project/gtrack/) package. The indicator field is ~1 inside the lithosphere and ~0 in the mantle, with a smooth tanh transition suitable for viscosity modification.

## What's included

- **LithosphereConnector**: Combines oceanic lithosphere (ages tracked via gtrack) with continental lithosphere (back-rotated present-day thickness)
- **GplatesScalarFunction**: Firedrake Function that auto-updates from plate reconstructions
- **LithosphereConfig**: Configuration dataclass following G-ADOPT's config pattern
- **MPI support**: Rank 0 handles I/O and gtrack computations, broadcasts to other ranks
- **Extended pyGplatesConnector**: Added `continental_polygons` and `static_polygons` parameters

## Demo & Tests

- Demo in `demos/mantle_convection/gplates_lithosphere/` runs with 4 MPI ranks
- Unit tests in `tests/unit/test_lithosphere_connector.py`

## Dependencies

- Adds `gtrack` to `pyproject.toml`

closes https://github.com/g-adopt/g-adopt/issues/383

PR #516, opened 2026-05-16
Add the Source / OutputStrategy / ScalarFieldConnector core and GplatesScalarFunction

sghelichkhani:
This replaces the original four-class connector hierarchy with a cleaner composition, which is the heart of the redesign Daniel's review prompted. What used to be `LithosphereConnector`, `PolygonConnector`, and their two hand-rolled geotherm subclasses is now the cartesian product of two ideas: a `Source` owns the stateful gtrack machinery and answers "where do source points live and what do they carry at age X", and an `OutputStrategy` is pure post-processing that turns interpolated arrays into a scalar field. A `ScalarFieldConnector` glues the two together with kNN interpolation and an `(age, coords)` cache.

The piece I'd point a reviewer at first is the `provides`/`requires` contract: a Source advertises the property keys it returns, an Output advertises what it needs, and the connector checks `requires <= provides` when you build it. So pairing a polygon source (no age column) with an erf geotherm (needs age) fails the moment you wire them, not on the first update.

Only the abstractions land here: the `Source` ABC with its per-age cache and lazy-load hook, the `OutputStrategy` ABC, `MeshConfig` and `InterpolationConfig`, and `PlateModelFiles` for the user-owned polygon paths (these no longer hang off `pyGplatesConnector`, which was only ever a velocity object). `GplatesScalarFunction` wraps any connector as a Firedrake `Function` and round-trips through the adjoint tape; it rejects vector/tensor spaces up front instead of failing cryptically deep inside Firedrake.

This closes most of the inline comments from the original review: the magic minimum-points number is now a named constant, the redundant numpy import is gone, the broadcast uses `comm.bcast` directly, the `_ad_*` helpers name their result clearly, and the scalar-space check Daniel asked for is in. The shallow-copy and silent-drop config concerns are moot because the config-merging machinery they referred to was deleted in the redesign. The result cache keys on the identity of the mesh-coordinate buffer rather than hashing ~24 MB on every call, and sibling connectors sharing one source reuse a single kNN geometry build.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -129,7 +129,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           uv venv --system-site-packages
-          uv pip install .[demos,optimisation,gplates] pygplates
+          uv pip install .[demos,optimisation,gplates]
 
       - name: Get Muller et al 2022 plate reconstructions
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -129,7 +129,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           uv venv --system-site-packages
-          uv pip install .[demos,optimisation] pygplates
+          uv pip install .[demos,optimisation,gplates] pygplates
 
       - name: Get Muller et al 2022 plate reconstructions
         run: |
--- a/gadopt/gplates/connectors.py
+++ b/gadopt/gplates/connectors.py
@@ -34,7 +34,7 @@ from mpi4py import MPI
 from scipy.spatial import cKDTree
 
 from ..utility import log, DEBUG
-from .outputs import MeshConfig, OutputStrategy, _DEFAULT_MESH_PARAMETERS
+from .outputs import MeshConfig, OutputStrategy
 from .sources import Source
 
 
@@ -148,8 +148,8 @@ class ScalarFieldConnector:
 
         self.source = source
         self.output = output
-        self.mesh = mesh or _DEFAULT_MESH_PARAMETERS
-        self.interpolation = interpolation or _DEFAULT_INTERPOLATION_PARAMETERS
+        self.mesh = mesh or MeshConfig()
+        self.interpolation = interpolation or InterpolationConfig()
         self.gc_collect_frequency = gc_collect_frequency
 
         # Result cache: keyed on (age, identity of the target_coords buffer).
--- a/gadopt/gplates/connectors.py
+++ b/gadopt/gplates/connectors.py
@@ -179,7 +179,6 @@ class ScalarFieldConnector:
         return self.source.comm
 
     # Main entry point
-
     def get_indicator(
         self,
         target_coords: np.ndarray,
@@ -197,6 +196,7 @@ class ScalarFieldConnector:
         if use_cache:
             return self._cached_result
 
+        # If the cache is not suitable, prepare the source and compute the result
         sources_dict = self.source.prepare(age)
         result = self._compute(sources_dict, target_coords)
         self._update_cache(age, target_coords, result)
@@ -209,12 +209,15 @@ class ScalarFieldConnector:
         return result
 
     # Cache
-
     def _check_cache(self, age: float, target_coords: np.ndarray) -> bool:
+        # Case when everything is fresh
         if self.reconstruction_age is None:
             return False
+        # Case where we have gone over the delta_t
         if abs(age - self.reconstruction_age) >= self.delta_t:
             return False
+        # Case where we do not even have a cached result (Not sure how this happens)
+        # But just for safety!
         if self._cached_result is None or self._cached_coords_ref is None:
             return False
         # A dead referent dereferences to None and is never ``is`` the live
@@ -226,15 +229,16 @@ class ScalarFieldConnector:
             level=DEBUG)
         return True
 
+    # Update the cache
     def _update_cache(
         self, age: float, target_coords: np.ndarray, result: np.ndarray
     ) -> None:
+        # Here we are just weak referencing the target_coords array
         self.reconstruction_age = age
         self._cached_result = result
         self._cached_coords_ref = weakref.ref(target_coords)
 
     # Computation
-
     def _compute(
         self,
         sources_dict: dict[str, np.ndarray],
--- a/gadopt/gplates/connectors.py
+++ b/gadopt/gplates/connectors.py
@@ -28,7 +28,6 @@ from __future__ import annotations
 import gc
 import weakref
 from dataclasses import dataclass
-from typing import TYPE_CHECKING
 
 import numpy as np
 from mpi4py import MPI
@@ -37,9 +36,6 @@ from scipy.spatial import cKDTree
 from ..utility import log, DEBUG
 from .outputs import MeshConfig, OutputStrategy, _DEFAULT_MESH_PARAMETERS
 from .sources import Source
-
-if TYPE_CHECKING:
-    pass
 
 
 # ---------------------------------------------------------------------------
--- a/gadopt/gplates/connectors.py
+++ b/gadopt/gplates/connectors.py
@@ -57,7 +57,7 @@ class InterpolationConfig:
     ``too_far`` and handled by the OutputStrategy (which decides whether
     to fill with a default thickness, switch to mantle temperature, etc).
 
-    Defaults: 0.1 rad \approx 640 km — generous; tighten to e.g. 0.02 rad
+    Defaults: 0.1 rad ≈ 640 km — generous; tighten to e.g. 0.02 rad
     (~127 km) for sharper boundaries. For polygon sources where the
     boundary is encoded by zero-thickness halo seeds, the threshold
     degenerates into a pathological-query guard and the actual roll-off
--- a/gadopt/gplates/connectors.py
+++ b/gadopt/gplates/connectors.py
@@ -87,9 +87,6 @@ class InterpolationConfig:
             )
 
 
-_DEFAULT_INTERPOLATION_PARAMETERS = InterpolationConfig()
-
-
 # ---------------------------------------------------------------------------
 # ScalarFieldConnector
 # ---------------------------------------------------------------------------
--- a/gadopt/gplates/connectors.py
+++ b/gadopt/gplates/connectors.py
@@ -1,0 +1,335 @@
+"""ScalarFieldConnector — composes a Source with an OutputStrategy.
+
+The connector orchestrates one timestep:
+
+  age = source.ndtime2age(ndtime)
+  source.validate_age(age)
+  if cached (age, identity of the target_coords buffer) match: return cached
+  sources_dict = source.prepare(age)             # collective; per-age cached
+  interp = self._interpolate(sources_dict, target_coords, output.requires)
+  result = output.compute(interp, r_target, too_far, mesh)
+  cache + return
+
+The two cache layers (source.prepare's per-age cache and the connector's
+(age, identity of the target_coords buffer) cache) are independent. The
+source cache decision is deterministic on ``age`` alone, so it is identical
+on every rank — no collective is needed there. The connector cache decision
+differs per rank (the target buffer differs with partitioning) and is
+allreduced before any collective work runs.
+
+Source/Output pairing is validated at construction: ``output.requires`` must
+be a subset of ``source.provides``. Mismatches (e.g. PolygonSource paired
+with GeothermERFOutput, which needs ``"age"``) fail immediately with a clear
+error rather than silently dropping the missing key.
+"""
+
+from __future__ import annotations
+
+import gc
+import weakref
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+import numpy as np
+from mpi4py import MPI
+from scipy.spatial import cKDTree
+
+from ..utility import log, DEBUG
+from .outputs import MeshConfig, OutputStrategy, _DEFAULT_MESH_PARAMETERS
+from .sources import Source
+
+if TYPE_CHECKING:
+    pass
+
+
+# ---------------------------------------------------------------------------
+# Interpolation config
+# ---------------------------------------------------------------------------
+
+@dataclass
+class InterpolationConfig:
+    """Spherical kNN-interpolation knobs.
+
+    The interpolation step normalises both source and target points to the
+    unit sphere, builds a cKDTree over sources, queries ``k_neighbors``
+    nearest neighbours per target node, and computes a weighted average
+    using either inverse-distance (``"idw"``) or Gaussian (``"gaussian"``)
+    weights.
+
+    Target nodes whose nearest source seed is farther than
+    ``distance_threshold`` (in radians on the unit sphere) are flagged as
+    ``too_far`` and handled by the OutputStrategy (which decides whether
+    to fill with a default thickness, switch to mantle temperature, etc).
+
+    Defaults: 0.1 rad ≈ 640 km — generous; tighten to e.g. 0.02 rad
+    (~127 km) for sharper boundaries. For polygon sources where the
+    boundary is encoded by zero-thickness halo seeds, the threshold
+    degenerates into a pathological-query guard and the actual roll-off
+    length is controlled by ``gaussian_sigma`` instead.
+    """
+
+    kernel: str = "idw"
+    k_neighbors: int = 50
+    distance_threshold: float = 0.1
+    gaussian_sigma: float = 0.04
+
+    def __post_init__(self):
+        valid_kernels = ("idw", "gaussian")
+        if self.kernel not in valid_kernels:
+            raise ValueError(
+                f"kernel must be one of {valid_kernels}, got '{self.kernel}'"
+            )
+        if self.k_neighbors < 1:
+            raise ValueError(f"k_neighbors must be at least 1, got {self.k_neighbors}")
+        if self.distance_threshold <= 0:
+            raise ValueError(
+                f"distance_threshold must be positive, got {self.distance_threshold}"
+            )
+        if self.gaussian_sigma <= 0:
+            raise ValueError(
+                f"gaussian_sigma must be positive, got {self.gaussian_sigma}"
+            )
+
+
+_DEFAULT_INTERPOLATION_PARAMETERS = InterpolationConfig()
+
+
+# ---------------------------------------------------------------------------
+# ScalarFieldConnector
+# ---------------------------------------------------------------------------
+
+class ScalarFieldConnector:
+    """Composition of a Source and an OutputStrategy into a single time-
+    varying scalar field on a mesh.
+
+    Construction validates ``output.requires <= source.provides``; this
+    catches the obvious mis-pairings (e.g. PolygonSource + GeothermERFOutput,
+    which needs ``"age"``) at the point of wiring rather than at the first
+    ``get_indicator`` call.
+
+    Two consumers (e.g. an indicator and a geotherm) that share the same
+    Source instance see a single, coherent advance of the source's
+    underlying state per geological age — the source's per-age cache
+    enforces that, so the order of ``get_indicator`` calls between
+    consumers is immaterial.
+
+    ``gc_collect_frequency`` controls how often a full ``gc.collect()`` runs
+    in the update loop (every Nth ``get_indicator``). The default of ``10``
+    matches gtrack's own internal GC cadence and periodically breaks the
+    pygplates C++ reference cycles without paying a collection on every call.
+    Set it to ``None`` to disable the connector-level collect entirely (relying
+    on gtrack's internal collect plus Python's automatic generational GC) when
+    you have profiled GC as hot and confirmed memory stays bounded; set it to
+    ``1`` for a lithosphere spin-up or very-long adjoint run where the
+    connector is driven for thousands of ages and the tightest bound on C++
+    cycle accumulation is wanted (the per-call cost is negligible there against
+    the per-age ``step_to``).
+    """
+
+    def __init__(
+        self,
+        source: Source,
+        output: OutputStrategy,
+        *,
+        mesh: MeshConfig | None = None,
+        interpolation: InterpolationConfig | None = None,
+        gc_collect_frequency: int | None = 10,
+    ):
+        if not output.requires <= source.provides:
+            missing = output.requires - source.provides
+            raise ValueError(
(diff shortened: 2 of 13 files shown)


gadopt/gplates/connectors.py
```diff
@@ -0,0 +1,559 @@
...
+        n_points, r_outer, depth_scale, transition_width.
+
+    Interpolation parameters (k_neighbors, distance_threshold, kernel,
+    etc.) are grouped in the ``interpolation`` field, an instance of
+    :class:`InterpolationConfig`.
+    """
+
+    def __post_init__(self):
+        """Validate fields common to all indicator configs."""
+        if hasattr(self, "n_points") and self.n_points < 100:
```
dsroberts: Bit of a magic number here. Maybe define e.g. `indicator_config_min_pts = 100` somewhere and use that instead.

gadopt/gplates/connectors.py
```diff
@@ -0,0 +1,559 @@
...
+    @classmethod
+    def from_dict(cls, config_dict: dict):
+        """Create configuration from dictionary, ignoring unknown keys.
+
+        Interpolation parameters (k_neighbors, distance_threshold, kernel,
+        gaussian_sigma, default_value) can appear either as top-level keys
+        or nested under an ``"interpolation"`` dict. The legacy key
+        ``default_thickness`` is accepted as an alias for ``default_value``.
+        """
+        d = dict(config_dict)
```
dsroberts: Must be careful here with copying a dict this way. I _think_ creating `merged` they way you've done on L149 is safe (unless you're expecting `nested` to contain values that are dicts), but in general any dict value of the top-level dict will only be copied by reference.

gadopt/gplates/connectors.py
```diff
@@ -0,0 +1,559 @@
...
+
+        if interp_overrides:
+            d["interpolation"] = InterpolationConfig(**{
+                **asdict(InterpolationConfig()),
+                **interp_overrides,
+            })
+
+        known_fields = {f.name for f in fields(cls)}
+        filtered = {k: v for k, v in d.items() if k in known_fields}
+        return cls(**filtered)
```
dsroberts: Correct me if I'm misreading this, but this looks like a bit of a foot-gun in that if a subclass does not have an `interpolation` field, it gets constructed then above then nuked by this filter here. I'm assuming every subclass needs an `interpolation` field? Maybe `IndicatorConfigBase` should provide that instead of the subclasses.

gadopt/gplates/connectors.py
```diff
@@ -0,0 +1,559 @@
...
+
+    def _broadcast_sources(self, sources: dict | None) -> dict:
+        """Broadcast source arrays from rank 0 to all MPI ranks."""
+        keys = self.comm.bcast(
+            list(sources.keys()) if self._is_root else None, root=0
+        )
+        return {
+            k: self.comm.bcast(sources[k] if self._is_root else None, root=0)
+            for k in keys
+        }
```
dsroberts: This function isn't necessary. `mpi4py` can broadcast dicts.

gadopt/gplates/connectors.py
```diff
@@ -0,0 +1,559 @@
...
+
+        if use_cache:
+            return self._cached_result
+
+        if self._is_root:
+            sources = self._prepare_sources(age)
+        else:
+            sources = None
+
+        sources = self._broadcast_sources(sources)
```
dsroberts:
```suggestion
        sources = self.comm.bcast(sources, root=0)
```
`mpi4py` can bcast dicts.

gadopt/gplates/connectors.py
```diff
@@ -0,0 +1,559 @@
...
+            (M,) array of indicator values in [0, 1].
+        """
+        # Verify subclass set all required attributes in __init__.
+        # Checked once per instance to give a clear TypeError rather than
+        # a confusing AttributeError deep in the template method.
+        if not getattr(self, "_init_checked", False):
+            for attr in ("gplates_connector", "config", "comm", "_is_root",
+                         "reconstruction_age", "_cached_result",
+                         "_cached_coords_hash", "_transition_width_nondim",
+                         "_initialized"):
```
dsroberts: If subclasses are also supposed to provide `_polygon_filter`, `_rotator` and `_PointCloud`, should those be checked for here?

gadopt/gplates/connectors.py
```diff
@@ -0,0 +1,559 @@
...
+            level=DEBUG)
+        cloud = self._polygon_filter.filter_inside(cloud, at_age=0.0)
+        # Guard against zero-point input; the f-string is evaluated eagerly
+        # so it would raise ZeroDivisionError even if the log level is off.
+        if n_before > 0:
+            log(f"{type(self).__name__}: After filtering: {cloud.n_points} points "
+                f"({100 * cloud.n_points / n_before:.1f}% retained)", level=DEBUG)
+        else:
+            log(f"{type(self).__name__}: Input had 0 points, nothing to filter.",
+                level=DEBUG)
```
dsroberts: We should revisit logging at some point. Standard practice is to let the logger perform the formatting to avoid exactly this, but `PETSc.Sys.Print` doesn't allow that.

dsroberts (commented):
I think we went over most of this earlier, so I only have a couple of minor things. I'm not done yet, I'll review the rest first thing tomorrow.

gadopt/gplates/gplates.py
```diff
@@ -240,6 +255,16 @@ def remove_radial_component(self):
...
         )
 
+    def _ad_mul(self, other):
+        r = GplatesVelocityFunction(self.function_space())
+        r.assign(other * self)
+        return r
+
+    def _ad_add(self, other):
+        r = GplatesVelocityFunction(self.function_space())
+        fd.Function.assign(r, self + other)
```
dsroberts: What? This is calling a class method as if it were a static method with `r` in place of `self`? Why?

gadopt/gplates/gplates.py
```diff
@@ -644,3 +669,83 @@ def _fibonacci_sphere(self, samples):
...
+            )
+
+        self.indicator_connector = indicator_connector
+
+        mesh = extract_unique_domain(self)
+
+        scalar_element = function_space.ufl_element()
+        vector_element = VectorElement(scalar_element)
+        coords_space = fd.FunctionSpace(mesh, vector_element)
+        coords_func = fd.Function(coords_space)
```
dsroberts: I tried this with a vector function space just to see what would happen and it failed on L720 with 
```
NotImplementedError: Can't make a Function defined on a <class 'finat.ufl.mixedelement.VectorElement'>
```
I get that its named `GplatesScalarFunction` but it can't hurt to have a check that `function_space` is actually a scalar function space before proceeding. Also:

```suggestion
        scalar_element = function_space.ufl_element()
        vector_element = fd.VectorElement(scalar_element)
        coords_space = fd.FunctionSpace(mesh, vector_element)
        coords_func = fd.Function(coords_space)
```

gadopt/gplates/gplates.py
```diff
@@ -644,3 +669,83 @@ def _fibonacci_sphere(self, samples):
...
+            self.create_block_variable()
+
+    def _ad_mul(self, other):
+        r = GplatesScalarFunction(self.function_space())
+        r.assign(other * self)
+        return r
+
+    def _ad_add(self, other):
+        r = GplatesScalarFunction(self.function_space())
+        fd.Function.assign(r, self + other)
```
dsroberts: Again, what? Why is this not `r.assign(self + other)`?

gadopt/gplates/gplates.py
```diff
@@ -1,17 +1,26 @@
 import warnings
 import firedrake as fd
 import numpy as np
+import pygplates
+from firedrake import VectorElement
```
dsroberts: Why does this get its own special import when it is bought in by `import firedrake as fd`?

tests/unit/test_gplates_adjoint.py
```diff
@@ -0,0 +1,468 @@
...
+
+        # ndtime=0 -> age=200 Ma
+        vf.update_plate_reconstruction(ndtime=0.0)
+        bv_1 = vf.block_variable
+
+        # ndtime=0.01 -> age=199 Ma, within delta_t=2.0
+        vf.update_plate_reconstruction(ndtime=0.01)
+        bv_2 = vf.block_variable
+
+        assert bv_1 is bv_2
```
dsroberts, line 452: In the `TestGplatesScalarFunctionDeltaTCaching` class you have tests on the call counts, but not in these tests which are fundamentally the same. I think for consistency call counts should be checked in this class too.

dsroberts (commented):
OK, that's the rest of the review. There are some things I found quite strange, but maybe that's just me?

sghelichkhani:
This is the structural rework your review pushed me towards, so a lot of the inline comments are either fixed or no longer apply against the new code. Quick map, dsroberts:

The repeated load/interpolate/broadcast flow is now a single template on the connector over a `Source` and an `OutputStrategy`, which is what collapsed the duplication you kept pointing at. The magic minimum-points number is a named constant, the redundant numpy import is gone, the dict broadcast uses `comm.bcast` directly rather than the custom helper, and the `_ad_*` helpers name their result clearly. The vector/tensor space check on `GplatesScalarFunction` is in, so a non-scalar space now fails with a clear message up front instead of deep inside Firedrake.

The shallow-copy footgun and the silently-dropped-`interpolation`-field comments don't have an analogue any more — the config-merging machinery they referred to was deleted; configs are now plain dataclasses passed explicitly. Same for the missing-subclass-attribute check: composition replaced the inheritance that needed it, so `Source` and `OutputStrategy` are constructor arguments rather than overrides.

Two efficiency changes while I was in here: the result cache keys on the identity of the mesh-coordinate buffer instead of hashing the whole thing every call, and sibling connectors that share a source reuse one kNN geometry build.

gadopt/gplates/sources.py
```diff
@@ -0,0 +1,177 @@
...
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import TYPE_CHECKING
+
+import numpy as np
+from mpi4py import MPI
+
+if TYPE_CHECKING:
```
sghelichkhani, line 23: I am not sure if this is what you guys @angus-g @dsroberts en/dis-courage. This would be a first to use, with clear (but maybe unnecessary) advantages.

dsroberts: I like that it provides added pedantry around type checking, but I dislike the introduction of circular dependencies. I guess since its gated behind `if TYPE_CHECKING` its OK. @angus-g sometimes has stronger feelings on this kind of thing than I do, so would be good to hear from him.

gadopt/gplates/connectors.py
```diff
@@ -0,0 +1,337 @@
...
+            raise ValueError(
+                f"distance_threshold must be positive, got {self.distance_threshold}"
+            )
+        if self.gaussian_sigma <= 0:
+            raise ValueError(
+                f"gaussian_sigma must be positive, got {self.gaussian_sigma}"
+            )
+
+
+_DEFAULT_INTERPOLATION_PARAMETERS = InterpolationConfig()
```
dsroberts: Creating objects like this and why it shouldn't be done came up in #366 (https://github.com/g-adopt/g-adopt/pull/366#discussion_r2875579568). Don't know why its back here.

gadopt/gplates/connectors.py
```diff
@@ -0,0 +1,337 @@
...
+import gc
+import weakref
+from dataclasses import dataclass
+
+import numpy as np
+from mpi4py import MPI
+from scipy.spatial import cKDTree
+
+from ..utility import log, DEBUG
+from .outputs import MeshConfig, OutputStrategy, _DEFAULT_MESH_PARAMETERS
```
dsroberts:
```suggestion
from .outputs import MeshConfig, OutputStrategy
```

gadopt/gplates/connectors.py
```diff
@@ -0,0 +1,337 @@
...
+        if gc_collect_frequency is not None and gc_collect_frequency < 1:
+            raise ValueError(
+                f"gc_collect_frequency must be >= 1 or None, "
+                f"got {gc_collect_frequency}"
+            )
+
+        self.source = source
+        self.output = output
+        self.mesh = mesh or _DEFAULT_MESH_PARAMETERS
+        self.interpolation = interpolation or _DEFAULT_INTERPOLATION_PARAMETERS
```
dsroberts:
```suggestion
        self.mesh = mesh or MeshConfig()
        self.interpolation = interpolation or InterpolationConfig()
```

gadopt/gplates/outputs.py
```diff
@@ -0,0 +1,83 @@
...
+    def __post_init__(self):
+        if self.r_outer <= 0:
+            raise ValueError(f"r_outer must be positive, got {self.r_outer}")
+        if self.depth_scale <= 0:
+            raise ValueError(f"depth_scale must be positive, got {self.depth_scale}")
+
+
+# Importable default used by the ScalarFieldConnector when the caller doesn't
+# pass a MeshConfig of their own.
+_DEFAULT_MESH_PARAMETERS = MeshConfig()
```
dsroberts:
```suggestion
```

gadopt/gplates/gplates.py
```diff
@@ -644,3 +689,89 @@ def _fibonacci_sphere(self, samples):
...
+            self.create_block_variable()
+
+    def _ad_mul(self, other):
+        result = GplatesScalarFunction(self.function_space())
+        result.assign(other * self)
+        return result
+
+    def _ad_add(self, other):
+        result = GplatesScalarFunction(self.function_space())
+        fd.Function.assign(result, self + other)
```
dsroberts:
```suggestion
        result.assign(self + other)
```

dsroberts (commented):
Looking good so far. The new structure makes it really clear that a connector actually connects stuff.

gadopt/gplates/gplates.py
sghelichkhani: Ha!!! We are at the point that you are only picking on pieces that I have written and are fine with Claude'd writing. 
So this is the way firedrake has implemented things. 
https://github.com/firedrakeproject/firedrake/blob/2018173c526c91db48cac43725b3d55ca81559bd/firedrake/adjoint_utils/function.py#L256
I am not really sure why they do it that way, but if not a mistake, I thought it is to avoid any subclass override of assign ... 
Not really sure. Slight preference to keep it consistent with upstream firedrake even though I think you are right?

gadopt/gplates/gplates.py
sghelichkhani: See comment above!

gadopt/gplates/connectors.py
sghelichkhani: Thanks Dale! Makes sense.

gadopt/gplates/sources.py
sghelichkhani: I don't know how to do the type checking w/o the circular deps. Maybe there is a better way? @angus-g ?

gadopt/gplates/gplates.py
connorjward: Going back through our history that code used to look like
```py
    @no_annotations
    def _ad_add(self, other):
        from firedrake import Function

        r = get_overloaded_class(Function)(self.function_space())
        Function.assign(r, self + other)
        return r
```
I think this is therefore a leftover from when we did not provide adjoint by default and had adjoint versions of all Firedrake types.

I think you can just do the normal thing here of `r.assign(...)`.

gadopt/gplates/sources.py
connorjward: As I understand it this is the recommended way of resolving circular type hints. I do this in my own code.

.github/workflows/test.yml
dsroberts:
```suggestion
          uv pip install .[demos,optimisation,gplates]
```
Not necessary now that `gadopt[gplates]` brings in `pygplates`. Also means this file doesn't have to change when `gtrack` becomes responsible for `pygplates`

gadopt/gplates/gplates.py
```diff
@@ -644,3 +689,89 @@ def _fibonacci_sphere(self, samples):
...
+            raise TypeError(
+                "GplatesScalarFunction requires a scalar function space; "
+                f"got {type(scalar_element).__name__}. Use one component or "
+                "switch to a scalar element."
+            )
+
+        self.indicator_connector = indicator_connector
+
+        mesh = extract_unique_domain(self)
+        vector_element = VectorElement(scalar_element)
```
dsroberts:
```suggestion
        vector_element = fd.VectorElement(scalar_element)
```

gadopt/gplates/gplates.py
```diff
@@ -644,3 +689,89 @@ def _fibonacci_sphere(self, samples):
...
+            raise TypeError(
+                f"indicator_connector must be a ScalarFieldConnector, "
+                f"got {type(indicator_connector).__name__}"
+            )
+
+        scalar_element = function_space.ufl_element()
+        # Tensor/vector spaces silently produce ill-shaped coords arrays and
+        # later fail deep inside Function construction with a cryptic
+        # NotImplementedError. Catch it up front.
+        if isinstance(scalar_element, (VectorElement, TensorElement)):
```
dsroberts:
```suggestion
        if isinstance(scalar_element, (fd.VectorElement, fd.TensorElement)):
```

gadopt/gplates/gplates.py
```diff
@@ -1,17 +1,27 @@
 import warnings
+from dataclasses import dataclass
+
 import firedrake as fd
 import numpy as np
+import pygplates
+from firedrake import TensorElement, VectorElement
```
dsroberts:
```suggestion
```

gadopt/gplates/gplatesfiles.py
dsroberts:
```suggestion
    # Return absolute paths of the files, normalised to lists
```
US spelling.

gadopt/gplates/sources.py
dsroberts: I'm going to assume no news from @angus-g is good news so I'll resolve this one.

tests/unit/test_outputs.py
```diff
@@ -0,0 +1,36 @@
+"""Pure-math tests for OutputStrategy subclasses and MeshConfig.
+
+No reconstruction data needed; everything in this file constructs its inputs
+directly. These tests cover the mathematical correctness of the indicator and
+geotherm transformations and the validation contracts of the small dataclasses.
+"""
```
dsroberts: This comment seems incongruous from what's actually going on here. Its a data validation test, there is no maths happening. We're making sure that the defaults are what we expect, that input parameters are honoured and that invalid values for those parameters are disallowed. Was this a larger set of tests in the past?

dsroberts: Oh, I see this gets expanded in the next PR. I'll reconsider this comment when I look at the other tests. Still, "Pure-math" means something, I don't think we're deriving any proofs here, and "math" is an Americanism.

dsroberts (commented):
Looking good. Still a few minor things but otherwise I think its ready to go. Also, some of the suggestions have come out showing a blank line instead of the original code. I'm not sure if that's going to break anything.

gadopt/gplates/gplates.py
sghelichkhani: Makes sense! I was just worried about overlooking some magic machinery from pyadjoint. Thanks @connorjward!

gadopt/gplates/outputs.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def __post_init__(self):
+        if self.r_outer <= 0:
+            raise ValueError(f"r_outer must be positive, got {self.r_outer}")
+        if self.depth_scale <= 0:
+            raise ValueError(f"depth_scale must be positive, got {self.depth_scale}")
+
+
+# Importable default used by the ScalarFieldConnector when the caller doesn't
+# pass a MeshConfig of their own.
```
dsroberts:
```suggestion
```
Remove comment corresponding to object that doesn't exist any more. Also, this change has not made it into #517 and the rest, please update those once this is done.

sghelichkhani: Yes! I am actually doing a bunch of cascading a fixing of different things now! Please wait for a few mins and I will come up with a few corrections fixes.

sghelichkhani: Done in bc719971 and merged up through the whole stack. Turned out the comment wasn't the only leftover: 03 still carried the `_DEFAULT_MESH_PARAMETERS` constant unused, and 04/05 still used it in connectors.py, so I converged everything on your earlier `mesh or MeshConfig()` suggestion and deleted the constant on all branches. Full suites green on 03, 04, 05 and the integration branch.

.github/workflows/test.yml
sghelichkhani: Done in bc719971 — the gplates extra already brings in pygplates, so the explicit one is gone.

gadopt/gplates/gplates.py
sghelichkhani: Removed in bc719971, together with switching the two uses over to fd.VectorElement and fd.TensorElement.

gadopt/gplates/gplates.py
sghelichkhani: Applied in bc719971.

gadopt/gplates/gplates.py
sghelichkhani: Applied in bc719971.

gadopt/gplates/gplatesfiles.py
sghelichkhani: Fixed in bc719971.

tests/unit/test_outputs.py
sghelichkhani: Fair point — on this branch the file really is just validation, so the docstring now says exactly that. On the downstream branches, where the numerical kernel tests live, it talks about the numerical behaviour of the transformations instead of "pure math". Both in bc719971 and the merge-ups.

gadopt/gplates/gplates.py
sghelichkhani: Now that Connor confirmed the `Function.assign` spelling was just a leftover from the pre-default-adjoint days, I've applied this — and the same in `GplatesVelocityFunction._ad_add` — in bc719971. The adjoint tests still pass.

dsroberts (approved):
Looks ready to go. Once the tests have passed I think this is good to merge.

sghelichkhani:
Because this is an ongoing multimaterial flaky test, I am going to bypass Regression failure and merge. 
image

PR #517, opened 2026-05-16
Add the lithosphere source, erf geotherm, and lithosphere factories

sghelichkhani:
The first concrete feature on top of the new core. `LithosphereSource` combines a forward-stepped `SeafloorAgeTracker` for the oceanic seeds with present-day continental thickness back-rotated by a `PointRotator`, concatenating the two into one cloud that carries both thickness and age. The ocean tracker is the only stateful piece, and the per-age cache on the base class keeps it to a single `step_to` per geological age no matter how many connectors share the source.

This is the PR that answers the "LithosphereGeotherm has me questioning the whole structure" note. The geotherm classes are gone: a normalised oceanic geotherm is now just `ScalarFieldConnector(LithosphereSource(...), GeothermERFOutput(...))`, sharing the same source as the indicator. The structural unease was the right instinct, and the Source/Output split in the previous PR is the answer.

A couple of robustness points worth flagging. The walk is forward-only and can't rewind, so an optional `walk_start_age` turns the old silently-locked ceiling into a declared, fail-fast contract checked up front. The first-call initialisation was lifted out of `_step_ocean_to` into `_initialize_walk` so the stepping path reads as a flat sequence, and `_find_best_checkpoint` is now a module-level helper rather than a private staticmethod. On the MPI side: I checked gtrack's source, and `step_to`, `reinitialize`, and the checkpoint calls are all rank-local with no internal collectives, so the rank-0-compute-then-broadcast pattern in `Source.prepare` is safe from deadlock.

Outputs gain `TanhOutput` (the smooth indicator) and `GeothermERFOutput` (the age-dependent oceanic geotherm). Two factories, `lithosphere_indicator` and `lithosphere_geotherm`, absorb the common wiring; to share one tracker between them you build the source once and pass it to both. The docstring-style, `gc_collect_frequency` placement, and `**kwargs`/spelling comments from the original review are addressed here.

### DIFF
--- a/gadopt/gplates/__init__.py
+++ b/gadopt/gplates/__init__.py
@@ -2,12 +2,28 @@ from .connectors import ScalarFieldConnector, InterpolationConfig
 from .gplates import (
     GplatesScalarFunction,
     GplatesVelocityFunction,
-    PlateModelFiles,
     ensure_reconstruction,
     pyGplatesConnector,
+    PlateModelFiles,
 )
-from .outputs import MeshConfig, OutputStrategy
-from .sources import Source
+from .outputs import (
+    GeothermERFOutput,
+    MeshConfig,
+    OutputStrategy,
+    QuinticOutput,
+    ocean_erf_normalized,
+    radial_quintic_step,
+)
+from .sources import (
+    CloudDataType,
+    LithosphereSource,
+    LithosphereSourceConfig,
+    Source,
+)
+from .factories import (
+    ConnectorFactory,
+    LithosphereConnectorFactory,
+)
 
 __all__ = [
     # Firedrake function wrappers
@@ -21,7 +37,18 @@ __all__ = [
     "ScalarFieldConnector",
     "InterpolationConfig",
     "MeshConfig",
-    # Abstract source / output contracts
+    # Sources
     "Source",
+    "LithosphereSource",
+    "LithosphereSourceConfig",
+    "CloudDataType",
+    # Outputs
     "OutputStrategy",
+    "QuinticOutput",
+    "GeothermERFOutput",
+    "ocean_erf_normalized",
+    "radial_quintic_step",
+    # Factories
+    "ConnectorFactory",
+    "LithosphereConnectorFactory",
 ]
--- a/gadopt/gplates/gplates.py
+++ b/gadopt/gplates/gplates.py
@@ -1,6 +1,7 @@
 import warnings
 from dataclasses import dataclass
 
+
 import firedrake as fd
 import numpy as np
 import pygplates
@@ -18,7 +19,6 @@ __all__ = [
     "ensure_reconstruction",
     "GplatesVelocityFunction",
     "GplatesScalarFunction",
-    "ScalarFieldConnector",
     "PlateModelFiles",
     "pyGplatesConnector",
 ]
@@ -695,9 +695,8 @@ class GplatesScalarFunction(fd.Function):
     reconstructions.
 
     Wraps any ``ScalarFieldConnector`` (built either directly via composition
-    of a ``Source`` and an ``OutputStrategy``, or via one of the convenience
-    factories ``lithosphere_indicator``, ``lithosphere_geotherm``,
-    ``polygon_indicator``, ``polygon_geotherm``). Calling
+    of a ``Source`` and an ``OutputStrategy``, or via a ``ConnectorFactory``
+    such as ``LithosphereConnectorFactory``). Calling
     ``update_plate_reconstruction(ndtime)`` recomputes the field and assigns
     it onto the underlying Firedrake DoFs.
 
--- a/gadopt/gplates/sources.py
+++ b/gadopt/gplates/sources.py
@@ -14,19 +14,175 @@ source points live and what properties do they carry at age X?", Outputs answer
 
 from __future__ import annotations
 
+import re
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING
+from dataclasses import dataclass
+from pathlib import Path
+from typing import TYPE_CHECKING, Callable
 
+import h5py
 import numpy as np
+import numpy.typing as npt
 from mpi4py import MPI
 
+from gtrack import PointCloud, PointRotator, PolygonFilter, SeafloorAgeTracker
+from gtrack.config import TracerConfig
+from gtrack.mesh import create_sphere_mesh_latlon
+
+from ..utility import log, DEBUG, INFO
+
 if TYPE_CHECKING:
-    from .gplates import pyGplatesConnector
+    from .gplates import pyGplatesConnector, PlateModelFiles
 
 
 # Minimum points to make any sphere-coverage source meaningful. Anything
 # below this and the kNN interpolator gives nonsense.
 SOURCE_MIN_POINTS = 100
+
+
+# ---------------------------------------------------------------------------
+# Source configs
+# ---------------------------------------------------------------------------
+
+@dataclass
+class LithosphereSourceConfig:
+    """Knobs for LithosphereSource — ocean tracker, checkpointing, data loading.
+
+    Mesh geometry (r_outer, depth_scale) lives on MeshConfig.
+    Interpolation (k_neighbors, kernel, ...) lives on InterpolationConfig.
+    Output knobs (transition_width, kappa, ...) live on the output strategy.
+
+    Checkpointing is enabled by setting ``checkpoint_interval_myr``;
+    ``checkpoint_dir`` then defaults to ``gtrack_checkpoints`` (resolved
+    against the cwd at construction time and created up front on rank 0).
+    """
+
+    time_step: float = 1.0
+    n_points: int = 10000
+    reinit_interval_myr: float = 50.0
+    property_name: str = "thickness"
+    gtrack_config: dict | None = None
+    checkpoint_interval_myr: float | None = None
+    checkpoint_dir: str | Path | None = None
+
+    def __post_init__(self):
+        if self.n_points < SOURCE_MIN_POINTS:
+            raise ValueError(
+                f"n_points must be at least {SOURCE_MIN_POINTS}, got {self.n_points}"
+            )
+        if self.time_step <= 0:
+            raise ValueError(f"time_step must be positive, got {self.time_step}")
+        if self.reinit_interval_myr <= 0:
+            raise ValueError(
+                f"reinit_interval_myr must be positive, "
+                f"got {self.reinit_interval_myr}"
+            )
+        if (self.checkpoint_interval_myr is not None
+                and self.checkpoint_interval_myr <= 0):
+            raise ValueError(
+                f"checkpoint_interval_myr must be positive or None, "
+                f"got {self.checkpoint_interval_myr}"
+            )
+
+
+# ---------------------------------------------------------------------------
+# Data-loading helpers (module-level — both Sources need them)
+# ---------------------------------------------------------------------------
+
+def _load_from_hdf5(filepath: str | Path, property_name: str) -> PointCloud:
+    """Read lat/lon/values from an HDF5 (or NetCDF) file into a PointCloud."""
+    with h5py.File(filepath, "r") as f:
+        lon = f["lon"][:]
+        lat = f["lat"][:]
+        if property_name in f:
+            values = f[property_name][:]
+        elif "z" in f:
+            values = f["z"][:]
+        else:
+            raise KeyError(
+                f"File must contain '{property_name}' or 'z'. "
+                f"Available datasets: {list(f.keys())}"
+            )
+
+    lon = np.where(lon > 180, lon - 360, lon)
+
+    if lon.ndim == 1 and lat.ndim == 1:
+        lon_grid, lat_grid = np.meshgrid(lon, lat)
+        latlon = np.column_stack([lat_grid.ravel(), lon_grid.ravel()])
+        values = values.ravel()
+    else:
+        latlon = np.column_stack([lat.ravel(), lon.ravel()])
+        values = values.ravel()
+
+    cloud = PointCloud.from_latlon(latlon)
+    cloud.add_property(property_name, values)
+    return cloud
+
+
+CloudDataType = PointCloud | tuple[npt.ArrayLike, npt.ArrayLike] | str | Path | int | float
+
+
+def _build_cloud(data: CloudDataType, property_name: str, n_points_fallback: int) -> PointCloud:
+    """Dispatch the input ``data`` to the right PointCloud constructor.
+
+    Accepts: an existing PointCloud, an HDF5 path, a (latlon, values) tuple,
+    or a scalar (broadcast onto a Fibonacci sphere mesh).
+    """
+    if isinstance(data, PointCloud):
+        return data
+    if isinstance(data, (str, Path)):
+        return _load_from_hdf5(data, property_name)
+    if isinstance(data, tuple) and len(data) == 2:
+        latlon, values = data
+        cloud = PointCloud.from_latlon(np.asarray(latlon))
+        cloud.add_property(property_name, np.asarray(values))
+        return cloud
+    if isinstance(data, (int, float)):
+        lats, lons = create_sphere_mesh_latlon(n_points_fallback)
+        latlon = np.column_stack([lats, lons])
+        cloud = PointCloud.from_latlon(latlon)
+        cloud.add_property(property_name, np.full(len(latlon), float(data)))
+        return cloud
+    raise TypeError(
+        f"Unsupported data type: {type(data)}. "
+        "Expected PointCloud, file path, (latlon, values) tuple, or scalar."
+    )
+
+
+# ---------------------------------------------------------------------------
+# Checkpoint helpers
+# ---------------------------------------------------------------------------
+
+# Matched against Path.stem, so no ".npz" suffix here; the suffix is
+# checked separately in _find_best_checkpoint.
+_CHECKPOINT_RE = re.compile(r"^ocean_checkpoint_(\d+)Ma$")
+
+
+def _find_best_checkpoint(checkpoint_dir: Path | None, target_age: float) -> Path | None:
+    """Return the checkpoint path whose age is the smallest value >= target_age.
+
+    Smallest-age-not-younger-than-target minimises forward stepping. Returns
+    None if the directory contains no matching files. The directory is
+    created at source construction time, so if it has disappeared since,
+    the resulting error propagates rather than being silenced.
+    """
+    if checkpoint_dir is None:
+        return None
+    best_path = None
+    best_age = None
+    for fname in checkpoint_dir.iterdir():
+        if fname.suffix != ".npz":
+            continue
+        m = _CHECKPOINT_RE.match(fname.stem)
+        if m is None:
+            continue
+        ckpt_age = int(m.group(1))
+        if ckpt_age < target_age:
+            continue
+        if best_age is None or ckpt_age < best_age:
+            best_age = ckpt_age
+            best_path = fname
+    return best_path
 
 
 # ---------------------------------------------------------------------------
@@ -36,15 +192,15 @@ SOURCE_MIN_POINTS = 100
 class Source(ABC):
     """Time-dependent source of (xyz, properties) on a sphere.
 
-    Subclasses expose a ``provides`` set listing the property keys (excluding
-    "xyz") that ``prepare(age)`` returns. ``provides`` is declared here as an
-    abstract property, which a subclass may satisfy either with a plain class
-    constant (the concrete sources do this, keeping class-level access like
-    ``LithosphereSource.provides``) or with an instance property (handy for
-    test doubles and dynamically-configured sources). The ABC implements a
-    per-age cache so that consumers sharing this Source can call ``prepare``
-    in any order and the underlying gtrack state advances at most once per
-    age.
+    Subclasses expose a ``provides`` set listing the keys (including "xyz")
+    that ``prepare(age)`` returns. ``provides`` is declared here as an
+    abstract property, which a subclass may satisfy either with a read-only
+    property (the concrete sources do this) or with a plain class constant
+    (handy for test doubles and dynamically-configured sources, and keeps
+    class-level access like ``SomeSource.provides`` working). The ABC
+    implements a per-age cache so that consumers sharing this Source can call
+    ``prepare`` in any order and the underlying gtrack state advances at most
+    once per age.
 
     ``prepare`` is collective across ``self.comm``. Subclasses implement
     ``_compute_sources(age)``, which runs on rank 0 only; the ABC handles
@@ -54,7 +210,7 @@ class Source(ABC):
     @property
     @abstractmethod
     def provides(self) -> frozenset[str]:
-        """The set of property keys (excluding 'xyz') that prepare(age) returns."""
+        """The set of keys (including 'xyz') that prepare(age) returns."""
         ...
 
     gplates_connector: "pyGplatesConnector"
@@ -81,8 +237,8 @@ class Source(ABC):
 
     @abstractmethod
     def _compute_sources(self, age: float) -> dict[str, np.ndarray]:
-        """Build the source-arrays dict on rank 0. Must include 'xyz' plus
-        every key in ``self.provides``."""
+        """Build the source-arrays dict on rank 0. Must include every key in
+        ``self.provides`` ('xyz' among them)."""
 
     def _load(self) -> None:
         """Rank-0 I/O: build the gtrack machinery and the present-day cloud.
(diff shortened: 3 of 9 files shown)


gadopt/gplates/gplates.py
```diff
@@ -313,12 +331,29 @@ def __init__(self,
                     the finer velocity variations will then result in more challenging Stokes solves.
             nneighbours (Optional[int]): Number of neighbouring points when interpolating velocity for each grid point. Default is 4.
             kappa: (Optional[float]): Diffusion constant used for don-dimensionalising time. Default is 1e-6.
+            continental_polygons (Optional[Union[str, List[str]]]): Path(s) to continental polygon files.
```
dsroberts: This PR is a good opportunity to move the `__init__` docstring up immediately under the class declaration.

gadopt/gplates/gplates.py
```diff
@@ -671,6 +706,575 @@ def _fibonacci_sphere(self, samples):
...
+            runs is safe.
+    """
+
+    # Ocean tracker parameters
+    time_step: float = 1.0
+    n_points: int = 10000
+    reinit_interval_myr: float = 50.0
+
+    # Interpolation
+    interpolation: InterpolationConfig = None  # set in __post_init__
```
dsroberts:
```suggestion
    interpolation: InterpolationConfig | None = None  # set in __post_init__
```

gadopt/gplates/gplates.py
```diff
@@ -671,6 +706,575 @@ def _fibonacci_sphere(self, samples):
...
+            raise ValueError(f"time_step must be positive, got {self.time_step}")
+        if self.reinit_interval_myr <= 0:
+            raise ValueError(
+                f"reinit_interval_myr must be positive, got {self.reinit_interval_myr}"
+            )
+        if self.gc_collect_frequency is not None and self.gc_collect_frequency < 1:
+            raise ValueError(
+                f"gc_collect_frequency must be >= 1 or None, "
+                f"got {self.gc_collect_frequency}"
+            )
```
dsroberts: I feel like this is the wrong place to validate `gc_collect_frequency`. Given any subclass of `IndicatorConnector` can provide it, it should probably be validated there.

gadopt/gplates/gplates.py
```diff
@@ -671,6 +706,575 @@ def _fibonacci_sphere(self, samples):
...
+        if self._initialized and self.reconstruction_age is not None:
+            if age > self.reconstruction_age:
+                raise ValueError(
+                    f"Requested age {age:.2f} Ma is older than the last computed "
+                    f"age ({self.reconstruction_age:.2f} Ma). The ocean tracker "
+                    f"can only evolve forward in time (decreasing age)."
+                )
+
+    @staticmethod
+    def _find_best_checkpoint(checkpoint_dir, target_age):
```
dsroberts: Seems weird to declare a private function `@staticmethod`.

gadopt/gplates/gplates.py
```diff
@@ -671,6 +706,575 @@ def _fibonacci_sphere(self, samples):
...
+
+        return best_path
+
+    def _save_checkpoint_if_due(self, age):
+        """Write a checkpoint if the interval has been reached.
+
+        Failures are logged but never crash the simulation.
+        """
+        if self._checkpoint_dir is None:
+            return
```
dsroberts: Redundant check. The only way `self._checkpoint_dir` can not be `None` is if `self.config.checkpoint_interval_myr` is also not `None`. Might also be worth adding some debug level logging at the other `return` statements.

gadopt/gplates/gplates.py
```diff
@@ -671,6 +706,575 @@ def _fibonacci_sphere(self, samples):
...
+        if allow_reinit:
+            self._save_checkpoint_if_due(age)
+
+        return cloud
+
+
+# ---------------------------------------------------------------------------
+# Default geotherm functions
+# ---------------------------------------------------------------------------
+def ocean_erf_normalized(depth_m, z_lab_m, age_myr=None, kappa=1e-6, **kwargs):
```
dsroberts: Is `**kwargs` for API purposes? It isn't passed on to anything, so I'm guessing its here for if you switch `ocean_erf_normalized` (US spelling...) to some other geotherm and not change any of the arguments?

dsroberts (commented):
I'm not finished with this one, `LithosphereGeotherm` has me questioning the entire structure of `connectors.py` and `gplates.py`. I need to think about it some more. Unfortunately I'm not as quick as the robots, so if it has to go in we can come back and fix it up later.

sghelichkhani:
@dsroberts  , this is the PR for "LithosphereGeotherm has me questioning the whole structure": the geotherm classes don't exist any more. A normalised oceanic geotherm is now `ScalarFieldConnector(LithosphereSource(...), GeothermERFOutput(...))`, sharing the same source as the indicator, so there's no hand-rolled "shared tracker" relationship left to question.

The smaller inline points are handled: the `__init__` docstring is at class level, `gc_collect_frequency` validation moved onto the connector where it belongs, `_find_best_checkpoint` is a module-level function rather than a private staticmethod, the redundant `None` check is gone, and the geotherm helper no longer takes `**kwargs` it never forwarded.

One thing I confirmed while reworking the MPI path: gtrack's `step_to`, `reinitialize` and checkpoint calls are all rank-local with no internal collectives, so the rank-0-compute-then-broadcast in `Source.prepare` can't deadlock. I am not sure how to properly do this. But for the time being the assumption is that gtrack remains rank agnostic! 

I also added a `walk_start_age` so the forward-only tracker fails fast on an out-of-range request rather than silently locking its ceiling at whichever connector happens to update first.

sghelichkhani:
@dsroberts  This is also ready for a review!  Instead of fixing the tests here, I moved the fix to gtrack package itself! Ran the CI now, so hopefully turns green (locally does)

gadopt/gplates/__init__.py
```diff
@@ -2,12 +2,25 @@
...
+    MeshConfig,
+    OutputStrategy,
+    TanhOutput,
+    ocean_erf_normalized,
+    radial_tanh_step,
+)
+from .sources import (
+    LithosphereSource,
+    LithosphereSourceConfig,
+    Source,
```
dsroberts, line 21: Is it likely that users are going to come up with custom sources? If not, does the abstract base class need to be visible to anything that imports `gadopt.gplates`?

gadopt/gplates/gplates.py
```diff
@@ -691,6 +710,138 @@ def _fibonacci_sphere(self, samples):
...
+#
+# When two consumers (an indicator and a geotherm) need to share the same
+# underlying Source (mandatory for LithosphereSource, whose SeafloorAgeTracker
+# is forward-only and expensive to construct), build the Source explicitly
+# and pass it to *both* factory calls via the ``source=`` keyword. The
+# per-age cache on the Source guarantees the tracker advances exactly once
+# per age, regardless of call order between the two connectors.
+
+
+def lithosphere_indicator(
```
dsroberts: So this isn't quite what I had in mind, and I think there may be issues of ambiguity with the arguments (e.g. Can I pass a `source` and then use this function to give it a new connector?). Do you mind if I have a go at putting together what I had in mind? I find it easier to put into code than into words...

dsroberts (commented):
I'll leave the review at this for now. Let me know if you're OK with me having a go at the factory functions.

gadopt/gplates/__init__.py
sghelichkhani: At the time of writing this up, I thought it should be possible to come up with new outputs at least When you look at the more recent commits on top of the downstream PRs you will see I ended up introducing at least two new outputs. But I can see why sources might be tricky, as it is mostly features that might end up being implemented in the backed. So happy to drop the `Source` at least.

gadopt/gplates/gplates.py
sghelichkhani: Of course! Have a go and push directory here. If I have comments I am happy to review the changes.

dsroberts:
@sghelichkhani, the factory implementation is where I want it. Please take a look. The idea is that the factory controls the ways in which Sources and Outputs can be combined. Once a factory has been created, you can either assign it a Source or have it create one, but not both. To have it create the Source and Outputs, it has to know which class to use, which are passed in at object creation time. This opens up the opportunity to create convenience subclasses with common combinations (see `LithosphereIndicator`). The other options is that you can just assign already constructed source and output objects to a `ConnectorFactory` object, or mix-and-match. A `ConnectorFactory` can create both an indicator and a geotherm, but doesn't have to. It only creates the objects on first access.

There is limited error checking, `Source` and `OutputStrategy` subclasses are responsible for error checking their input, `ConnectorFactory` just passes inputs straight through. Though it will throw an exception if a user attempts to assign multiple Sources or multiple indicator/geotherm Outputs. The only other fatal exception is attempting to construct an indicator or geotherm if the source class is not specified and a source has not been provided. It will do its best to construct those objects otherwise, and will issue a warning if it thinks you might have forgotten something.

The whole thing got a bit bulky so I moved it to a new file. The API has changed a little (see the tests), and I've added tests specifically for the factory class. I'm aware I may not have the relationship between connectors, indicators and geotherms right, so please rename anything I've messed up there.

gadopt/gplates/sources.py
```diff
@@ -175,3 +318,293 @@ def validate_age(self, age: float) -> None:
...
+    source and update at different cadences, whichever one fires first would
+    otherwise silently fix the walk's ceiling. Declare ``walk_start_age`` to
+    pin the oldest age you intend to request up front, so an out-of-range
+    request fails immediately at wiring time rather than locking the walk at
+    the first touch. Declaring ``walk_start_age`` does NOT let you revisit
+    ages already stepped past — it only declares the oldest age you intend to
+    request.
+    """
+
+    provides = frozenset({"xyz", "thickness", "age"})
```
dsroberts: I think that this is an incorrect override. Given its declared as a property in `Sources`, it should be here too. Doing this makes `provides` read-only as well, which is probably for the best.
```suggestion
    @property
    def provides(self):
        return frozenset({"xyz", "thickness", "age"})
```

gadopt/gplates/sources.py
```diff
@@ -14,21 +14,164 @@
...
+    return cloud
+
+
+def _build_cloud(data, property_name: str, n_points_fallback: int) -> PointCloud:
+    """Dispatch the input ``data`` to the right PointCloud constructor.
+
+    Accepts: an existing PointCloud, an HDF5 path, a (latlon, values) tuple,
+    or a scalar (broadcast onto a Fibonacci sphere mesh).
+    """
+    if hasattr(data, "xyz") and hasattr(data, "properties"):
```
dsroberts:
```suggestion
    if isinstance(data, PointCloud):
```

sghelichkhani:
The thing I really like ( and I think it's the right instinct) is factory makes the shared source the default rather than something you have to remember to set up. Having a single object hand you both `.indicator` and `.geotherm` off the same source means you basically can't get that wrong by construction.

sghelichkhani:
But I'm a bit more hesitant is around the surface the users see. Because `construct_source` and `construct_output` take through *args, **kwargs all the physical knobs that used to be in the signature, things like
  transition_width_km, kappa, walk_start_age, sort of disappear from view. You can't tab-complete them or read them off help() anymore, you have to go and read `LithosphereSource.__init__` by hand to know what's even accepted. For most of our code that'd be a nuisance, but for these particular parameters, which are physics rather than plumbing, I'd really like them to stay discoverable. In the same spirit I'm a little nervous about the path where reaching .indicator  without an output set just warns and builds a default. In an MPI run that warning tends to scroll past, and a silently defaulted thickness or transition width is exactly the kind of thing I'd rather have blow up loudly than guess. Because so much of the simulation depends on these choices and the reproducibility of them.

sghelichkhani:
So how about we keep the `lithosphere_indicator` and `lithosphere_geotherm` functions as the typed front door so the physics knobs stay visible, and then take the one-object-owns-the-shared-source bit, and fold it in as a thin helper that builds a single source and hands back both connectors already sharing it. 

In practice that's a (I THINK) handful of small additions that make the common case shorter, and it leaves everything else exactly as it is, the polygon side, the tests, the fixture generators, none of that has to move. So no churn for the rest of the stack.  Does that feel reasonable to you? Happy to have a go and you tell me if you reckon I'm missing something about how you pictured people calling it?

gadopt/gplates/sources.py
```diff
@@ -14,21 +14,164 @@
...
+    else:
+        latlon = np.column_stack([lat.ravel(), lon.ravel()])
+        values = values.ravel()
+
+    cloud = PointCloud.from_latlon(latlon)
+    cloud.add_property(property_name, values)
+    return cloud
+
+
+def _build_cloud(data, property_name: str, n_points_fallback: int) -> PointCloud:
```
dsroberts: Add a type hint for the `data` argument to `_build_cloud`
```suggestion
CloudDataType = PointCloud | tuple[np.ndarray, np.ndarray] | str | Path | int | float

def _build_cloud(data: CloudDataType, property_name: str, n_points_fallback: int) -> PointCloud:
```
Note that we're not typing ndarrays correctly, but that's not really important.

dsroberts (commented):
I have more to go, but I'm a bit pressed for time and want to respond to your latest comments.

sghelichkhani:
> I have more to go, but I'm a bit pressed for time and want to respond to your latest comments.

I am starting to sweat!

dsroberts:
> But I'm a bit more hesitant is around the surface the users see. Because `construct_source` and `construct_output` take through *args, **kwargs all the physical knobs that used to be in the signature, things like transition_width_km, kappa, walk_start_age, sort of disappear from view. You can't tab-complete them or read them off help() anymore, you have to go and read `LithosphereSource.__init__` by hand to know what's even accepted. For most of our code that'd be a nuisance, but for these particular parameters, which are physics rather than plumbing, I'd really like them to stay discoverable.

Oh, good point. In that sense we have a backward step in usability. Leave this to me, I think we can use the convenience classes to provide something like this.

> In the same spirit I'm a little nervous about the path where reaching .indicator without an output set just warns and builds a default. In an MPI run that warning tends to scroll past, and a silently defaulted thickness or transition width is exactly the kind of thing I'd rather have blow up loudly than guess. Because so much of the simulation depends on these choices and the reproducibility of them.

OK, then we'll make it will throw an exception.

sghelichkhani:
Sounds like you have a plan! 😄

dsroberts:
Latest commit is my idea to address discoverability concerns. I've overloaded the `construct_source`, `construct_output` and `construct_geotherm` functions in the `LithosphereIndicator` helper class to match the argument list of the respective class `__init__` functions. This way, when you use a helper class (which is optional but recommended) you now see argument lists as you would have for directly instantiating those classes. 
vscode_mouseover
Of course, you don't need to have the `ConnectorFactory` construct your source/output at all, you can construct manually and assign them.
```
source = LithosphereSource(...)
factory.source = source
```
Feel free to update the docstrings for those functions to match the respective `__init__`, but do leave the note about why those are there, otherwise I'll come back to this in the future and go 'what's the point of all of this repeated code?'.

gadopt/gplates/sources.py
```diff
@@ -175,3 +318,293 @@ def validate_age(self, age: float) -> None:
...
                 f"Ages must be >= 0 (present day)."
             )
+
+
+# ---------------------------------------------------------------------------
+# LithosphereSource
+# ---------------------------------------------------------------------------
+
+class LithosphereSource(Source):
+    """Combined oceanic + continental lithosphere source.
```
dsroberts, line 341: Please remind Claude of our docstring standards. `__init__` args (all of them) are documented at the class level.

gadopt/gplates/sources.py
```diff
@@ -175,3 +318,293 @@ def validate_age(self, age: float) -> None:
...
+        self._plate_files = plate_files
+
+        self._initialized = False
+        self._last_reinit_age: float | None = None
+        self._last_checkpoint_age: float | None = None
+        self._checkpoint_dir = (
+            self.config.checkpoint_dir or "./gtrack_checkpoints"
+            if self.config.checkpoint_interval_myr is not None
+            else None
+        )
```
dsroberts:
```suggestion
        cd = Path(self.config.checkpoint_dir or "gtrack_checkpoints").absolute()
        cd.mkdir(parents=True, exist_ok=True)
        self._checkpoint_dir = cd if config.checkpoint_interval_myr is not None else None
```
I recommend switching to `pathlib.Path` for this and handling directory creation here. You don't want to get part way through a simulation and have it fail to create the checkpoint directory because of a typo or something.

gadopt/gplates/sources.py
```diff
@@ -14,21 +14,164 @@
...
+    Output knobs (transition_width, kappa, ...) live on the output strategy.
+    """
+
+    time_step: float = 1.0
+    n_points: int = 10000
+    reinit_interval_myr: float = 50.0
+    property_name: str = "thickness"
+    gtrack_config: dict | None = None
+    checkpoint_interval_myr: float | None = None
+    checkpoint_dir: str | None = None
```
dsroberts:
```suggestion
    checkpoint_dir: Path | None = None
```

gadopt/gplates/sources.py
```diff
@@ -14,21 +14,164 @@
...
+                f"checkpoint_interval_myr must be positive or None, "
+                f"got {self.checkpoint_interval_myr}"
+            )
+
+
+# ---------------------------------------------------------------------------
+# Data-loading helpers (module-level — both Sources need them)
+# ---------------------------------------------------------------------------
+
+def _load_from_hdf5(filepath, property_name: str) -> PointCloud:
```
dsroberts:
```suggestion
def _load_from_hdf5(filepath: str | Path, property_name: str) -> PointCloud:
```

gadopt/gplates/sources.py
```diff
@@ -14,21 +14,164 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Checkpoint helpers
+# ---------------------------------------------------------------------------
+
+_CHECKPOINT_RE = re.compile(r"^ocean_checkpoint_(\d+)Ma\.npz$")
+
+
+def _find_best_checkpoint(checkpoint_dir: str | None, target_age: float):
```
dsroberts:
```suggestion
def _find_best_checkpoint(checkpoint_dir: Path | None, target_age: float) -> Path | None:
```

gadopt/gplates/sources.py
```diff
@@ -14,21 +14,164 @@
...
+    """Return the checkpoint path whose age is the smallest value >= target_age.
+
+    Smallest-age-not-younger-than-target minimises forward stepping. Returns
+    None if the directory is missing or contains no matching files.
+    """
+    if checkpoint_dir is None or not os.path.isdir(checkpoint_dir):
```
dsroberts:
```suggestion
    if checkpoint_dir is None:
```
If `checkpoint_dir` disappears after we've already confirmed its existence, we should allow an exception to be raised rather than silently returning.

gadopt/gplates/sources.py
```diff
@@ -14,21 +14,164 @@
...
+        return None
+    best_path = None
+    best_age = None
+    for fname in os.listdir(checkpoint_dir):
+        m = _CHECKPOINT_RE.match(fname)
```
dsroberts:
```suggestion
    for fname in checkpoint_dir.iterdir():
        m = _CHECKPOINT_RE.match(fname.stem)
```

gadopt/gplates/sources.py
```diff
@@ -14,21 +14,164 @@
...
+        if m is None:
+            continue
+        ckpt_age = int(m.group(1))
+        if ckpt_age < target_age:
+            continue
+        if best_age is None or ckpt_age < best_age:
+            best_age = ckpt_age
+            best_path = os.path.join(checkpoint_dir, fname)
```
dsroberts:
```suggestion
            best_path = fname
```

gadopt/gplates/sources.py
```diff
@@ -175,3 +318,293 @@ def validate_age(self, age: float) -> None:
...
+        if self._checkpoint_dir is None:
+            return
+        interval = self.config.checkpoint_interval_myr
+        if (self._last_checkpoint_age is not None
+                and abs(self._last_checkpoint_age - age) < interval):
+            return
+        rounded_age = int(round(age))
+        filepath = os.path.join(
+            self._checkpoint_dir, f"ocean_checkpoint_{rounded_age}Ma.npz"
+        )
```
dsroberts:
```suggestion
        filepath = self._checkpoint_dir / f"ocean_checkpoint_{rounded_age}Ma.npz"
```

gadopt/gplates/sources.py
```diff
@@ -175,3 +318,293 @@ def validate_age(self, age: float) -> None:
...
+        try:
+            os.makedirs(self._checkpoint_dir, exist_ok=True)
```
dsroberts:
```suggestion
```

dsroberts (commented):
Still not done but have to stop for meeting.

gadopt/gplates/__init__.py
sghelichkhani: Coming back to this one: I'd actually keep `Source` exported. The test doubles downstream subclass it to exercise the validation contract without any reconstruction I/O, and anyone writing a custom source for the factory needs it importable anyway. Dropping it would just push those imports onto a private path.

gadopt/gplates/sources.py
sghelichkhani: Done in 9138ec4f — it's a read-only property on the concrete sources now. While in there I noticed the docstrings disagreed with every implementation: the ABC said `provides` excludes "xyz" but `LithosphereSource` (and this suggestion) include it, so I've made the documented contract "includes xyz" throughout, with `requires` still excluding it. I kept the ABC permissive about property-vs-class-constant since the test doubles use both styles.

gadopt/gplates/sources.py
sghelichkhani: All of the pathlib/checkpoint suggestions are in (9138ec4f), with two amendments to the snippets as written. Matching on `Path.stem` means the regex can't keep its `\.npz$` anchor — as suggested it would never match and checkpoint resume would silently die — so the suffix is checked separately and a new `TestFindBestCheckpoint` pins the write-then-find round trip. And the `__init__` directory creation needed `self.config` (the local `config` is None whenever the caller relies on the default) plus a gate on `checkpoint_interval_myr`, otherwise every run grows a `gtrack_checkpoints/` directory even with checkpointing disabled; it's also rank-0 only so nothing collective touches the filesystem. Agreed on the fail-loud semantics: a directory that disappears mid-run now raises instead of silently re-initialising.

gadopt/gplates/sources.py
sghelichkhani: All eight constructor args are documented at class level now and the `__init__` docstring is gone — same treatment for `PolygonSource` on the downstream branch.

sghelichkhani:
The overload approach works well — I've kept it, along with your note about why the duplicated signatures exist. Three follow-ups on top, now pushed (e5a9cc0e and 9138ec4f): I renamed `LithosphereIndicator` to `LithosphereConnectorFactory`, taking you up on "rename anything I've messed up" — the object makes both an indicator and a geotherm, so calling it an indicator undersold it. I also closed the remaining `**kwargs` on the output overloads so the static checking is complete, and moved the connector-level parameters (`mesh`, `interpolation`, `gc_collect_frequency`) onto typed constructor arguments in place of `update_connector_params`, for exactly the discoverability reasons we discussed.

Accessing `.indicator` or `.geotherm` without an output now raises as agreed, and the tests grew coverage that was missing: the geotherm construction path, and an identity check that `factory.indicator.source is factory.geotherm.source` — the shared-tracker guarantee that motivated the whole class.

Since the functions are gone for good, I've cascaded the change through the stack: gtrack-04 gains a `PolygonConnectorFactory` in the same style (its `construct_output` keeps the polygon-specific `default_thickness_km=0` so too-far nodes still read as outside-the-region), with the regression tests and the reference generator ported so the pickled reference stays valid, and gtrack-05's demo now shows the factory route for the solo indicators alongside the explicit composition in Parts 1–2. Full suites are green locally on all three branches.

gadopt/gplates/sources.py
```diff
@@ -14,37 +14,193 @@
...
+        m = _CHECKPOINT_RE.match(fname.stem)
+        if m is None:
+            continue
+        ckpt_age = int(m.group(1))
+        if ckpt_age < target_age:
+            continue
+        if best_age is None or ckpt_age < best_age:
+            best_age = ckpt_age
+            best_path = fname
+    return best_path
```
dsroberts, line 185: I think checkpoint loading/saving may need a bit of a rethink. You have two sources of truth for the checkpoint file naming. The regular expression at the top and the hard-coded path on line 633. I'm thinking you could simplify this function if your checkpoint files were sortable, that is add leading zeros such that `sorted(checkpoints)` returns the list of files in order, then this function would only need to pick the last file it saw before `age` went above `target_age`.
```suggestion
_CHECKPOINT_PREFIX = "ocean_checkpoint_"
_CHECKPOINT_SUFFIX = "Ma.npz"


def _find_best_checkpoint(checkpoint_dir: Path | None, target_age: float) -> Path | None:
    """Return the checkpoint path whose age is the smallest value >= target_age.

    Smallest-age-not-younger-than-target minimises forward stepping. Returns
    None if the directory contains no matching files. The directory is
    created at source construction time, so if it has disappeared since,
    the resulting error propagates rather than being silenced.
    """
    if checkpoint_dir is None:
        return None
    checkpoint_files = sorted(checkpoint_dir.glob(f"{_CHECKPOINT_PREFIX}*{_CHECKPOINT_SUFFIX}"))
    prevpath = None
    for fpath in checkpoint_files:
        try:
            age = int(fpath.name.replace(_CHECKPOINT_PREFIX,'').replace(_CHECKPOINT_SUFFIX,''))
        except ValueError:
            # Found something that looks like checkpoint file without a valid age. Ignore?
            continue
        if age >= target_age:
            return prevpath
        prevpath = fpath
```

gadopt/gplates/sources.py
```diff
@@ -175,3 +331,311 @@ def validate_age(self, age: float) -> None:
...
+
+    def _save_checkpoint_if_due(self, age: float) -> None:
+        if self._checkpoint_dir is None:
+            return
+        interval = self.config.checkpoint_interval_myr
+        if (self._last_checkpoint_age is not None
+                and abs(self._last_checkpoint_age - age) < interval):
+            return
+        rounded_age = int(round(age))
+        filepath = self._checkpoint_dir / f"ocean_checkpoint_{rounded_age}Ma.npz"
```
dsroberts, line 637:
```suggestion
        filepath = self._checkpoint_dir / f"{_CHECKPOINT_PREFIX}{rounded_age:04}{_CHECKPOINT_SUFFIX}"
```

gadopt/gplates/sources.py
dsroberts, line 258: Not for this sequence of PRs, but there may be merit in restructuring this one day to remove the restriction of having all the IO done MPI rank 0. You may end up with a lot of indicators to load, and you could see improved performance by loading the data in parallel across a few ranks instead of loading everything serially on rank 0.

dsroberts, line 222: I don't see any checks on this, so probably less trouble to not have a `None` case.
```suggestion
    _cached_dict: dict[str, np.ndarray] = {}
```

dsroberts, line 231: Same here, though there are checks, they're easy enough to rewrite without a `None` case.
```suggestion
    _interp_geometry_cache: dict[???,???] = {}
```

dsroberts, line 280:
```suggestion
        self._interp_geometry_cache.clear()
```

dsroberts, line 299:
```suggestion
        return self._interp_geometry_cache.setdefault(key, build_fn())
```

gadopt/gplates/sources.py
```diff
@@ -175,3 +331,311 @@ def validate_age(self, age: float) -> None:
...
+                self._ocean_tracker.load_checkpoint(best)
+                loaded_age = self._ocean_tracker.current_age
+                log(f"LithosphereSource: loaded ocean checkpoint at "
+                    f"{loaded_age} Ma from {best}.", level=DEBUG)
+                self._last_reinit_age = loaded_age
+                self._last_checkpoint_age = loaded_age
+                loaded = True
+            except Exception as exc:
+                log(f"LithosphereSource: failed to load checkpoint "
+                    f"{best}: {exc}. Falling back to full init.", level=INFO)
```
dsroberts, line 604: I don't like how this whole block is wrapped in a try/except. If you're expecting `_initialize_walk` to carry on if it couldn't load a checkpoint, then only `self._ocean_tracker.load_checkpoint(best)` should be in the `try`. I think you should perhaps be specific about the exception too. Is there something that can happen here that would actually make you want to bail on the whole simulation?

tests/unit/test_gplates.py
```diff
@@ -75,3 +108,369 @@ def test_gplates(write_pvd=False):
...
+    """The checkpoint-discovery helper matches on Path stems, so the filename
+    pattern and the suffix are checked independently — guard both."""
+
+    def test_picks_smallest_age_not_younger_than_target(self, tmp_path):
+        from gadopt.gplates.sources import _find_best_checkpoint
+
+        for age in (50, 100, 200):
+            (tmp_path / f"ocean_checkpoint_{age}Ma.npz").touch()
+        (tmp_path / "ocean_checkpoint_75Ma.txt").touch()  # wrong suffix
+        (tmp_path / "other_120Ma.npz").touch()  # wrong stem
```
dsroberts, line 171:
```suggestion
        for age in (50, 100, 200):
            (tmp_path / f"ocean_checkpoint_{age:04}Ma.npz").touch()
        (tmp_path / "ocean_checkpoint_0075Ma.txt").touch()  # wrong suffix
        (tmp_path / "other_0120Ma.npz").touch()  # wrong stem
```

tests/unit/test_gplates.py
```diff
@@ -75,3 +108,369 @@ def test_gplates(write_pvd=False):
...
+
+    def test_none_dir_returns_none(self):
+        from gadopt.gplates.sources import _find_best_checkpoint
+
+        assert _find_best_checkpoint(None, 50.0) is None
+
+    def test_no_checkpoint_old_enough_returns_none(self, tmp_path):
+        from gadopt.gplates.sources import _find_best_checkpoint
+
+        (tmp_path / "ocean_checkpoint_50Ma.npz").touch()
```
dsroberts, line 184:
```suggestion
        (tmp_path / "ocean_checkpoint_0050Ma.npz").touch()
```

dsroberts (commented):
Some more thoughts on this one. I can't comment on the validity of the adjoint or output tests though.

dsroberts:
What's happening with this group of PRs in light of #534 and onwards?

sghelichkhani:
> What's happening with this group of PRs in light of #534 and onwards?

I was waiting to see if you are OK with it to close them down. Is that a yes?

dsroberts:
> > What's happening with this group of PRs in light of #534 and onwards?
> 
> I was waiting to see if you are OK with it to close them down. Is that a yes?

Sure, go right ahead.

sghelichkhani:
Closing this in favour of #535, which restructures the source layer onto the released gtrack 0.5.0 rather than building it inside gadopt. That's the direct answer to the "questioning the whole structure" point raised here: gadopt no longer defines its own source layer at all, it drives gtrack's producers. The lithosphere source work from this PR lives on there. Leaving the branch and this discussion in place for reference.

PR #518, opened 2026-05-16
Add the polygon source, linear geotherm, and polygon factories

sghelichkhani:
The second concrete feature. `PolygonSource` restricts thickness data to polygon-bounded regions such as cratons. Rather than a hard cut at the polygon edge, it keeps every seed but zeroes the thickness outside the polygons (mask-and-relabel), so the kNN interpolator smears smoothly across the boundary and the roll-off length is set by the interpolation sigma instead of the polygon outline. Undefined continental seeds in the sliver between the continental and static-polygon sets inherit the nearest defined-continental plate id, so they ride the adjacent continent through time instead of being dropped or drifting onto an ocean plate.

One subtlety worth a reviewer's eye: `PolygonSource` returns a single rotated cloud, and gtrack's property accessors hand back internal arrays by reference, so the per-age cache copies them on the way out. That copy is genuinely required here, unlike the lithosphere path whose concatenate already produces fresh arrays — the comment in the code spells out the asymmetry so the polygon copy isn't dropped by analogy later. `GeothermLinearOutput` and `continental_linear` add the linear continental geotherm, with mantle temperature (1.0) outside the region.

This PR also brings the full source and connector test suites. Their fixtures build both sources together, which is why they live here rather than split across the two feature PRs, and it makes this the natural place to pick the review back up end to end. Included are the four-factory lockstep regression, the shared-source single-step guarantee, and the kNN-geometry profiling note that closed out the question of whether to invert the interpolation tree (the build is effectively free; the inversion would change the numerics for no real gain).


sghelichkhani:
This one didn't get a pass last time, and it's the natural place to pick the review back up end to end — the full source and connector test suites live here because their fixtures build the lithosphere and polygon sources together. It carries the four-factory lockstep regression and the shared-source single-step guarantee, so the whole stack is exercised in one place.

The polygon edges are smoothed by a mask-and-relabel rather than a hard cut, and there's a sliver-inheritance step so seeds caught between the continental and static-polygon sets ride the adjacent continent instead of being dropped. The one place to look twice is the per-age cache copy in `PolygonSource`: gtrack hands back its internal arrays by reference for a single rotated cloud, so that copy is required, not defensive — the comment spells out why it differs from the lithosphere path.

sghelichkhani:
Closing this in favour of #535. The polygon source and its geotherm are folded into the gtrack 0.5.0 port there, so the whole source layer moves out of gadopt in one go rather than landing feature by feature. Branch and discussion kept for reference.

PR #519, opened 2026-05-16
Add the gplates_fields demo and CI data fetch

sghelichkhani:
The end-to-end demo. It drives surface velocities from the Muller 2022 reconstruction while building lithospheric and cratonic indicator and geotherm fields through the new connectors, and validates the run against a checked-in reference. It shows the shared-source pattern in practice: one `LithosphereSource` feeding both an indicator and an erf geotherm so the forward-only ocean tracker steps once per age, plus a `PolygonSource` for the cratons.

On Rhodri's documentation points from the old initial-condition demo: the new demo states the half-space cooling model and references it, the geotherm helpers note they return a normalised fraction in [0, 1], and the interpolation knobs now carry their units inline — `distance_threshold` is an angular reach in radians, which previously read as dimensionless, with the rough km equivalent alongside. The regression and render CI jobs unzip the reconstruction into the demo directory and pull the continental-thickness and craton boundary data; the render job installs the gplates extra so gtrack is present when the notebook runs.

Auto-downloading the craton and thickness data via an `ensure_reconstruction`-style helper stays out of scope here by the earlier agreement; the Makefile targets fetch it explicitly for now.


sghelichkhani:
Rhodri, this demo replaces the old initial-condition one and I've folded in your documentation points. It states and references the half-space cooling model, the geotherm helpers note they return a normalised [0, 1] fraction, and the interpolation knobs carry their units inline now — `distance_threshold` reads as an angular reach in radians with the rough km equivalent alongside, rather than looking dimensionless. It runs both sources in lockstep and validates against a checked-in reference.

Auto-downloading the craton and thickness data through an `ensure_reconstruction`-style helper is still out of scope here per our earlier agreement; the Makefile targets fetch it explicitly for now.

sghelichkhani:
Closing this in favour of #536, which rebuilds the same demo on the gtrack-backed API from #535. Same reconstruction and CI data fetch, driven through the ported producers instead of gadopt's own source classes. Branch kept for reference.

PR #520, opened 2026-05-19
Allow use of any Firedrake container (Angus's patch)

sghelichkhani:
Cherry-picks @angus-g's commit g-adopt/g-adopt@bc1cb88 ("Allow use of any Firedrake container"), which restructures the container image resolution in `.github/workflows/test.yml` so that the `dev-` prefix is built inside the expression via `format('dev-{0}', ...)` rather than hardcoded outside it. This lets `workflow_dispatch` accept a free-form container tag (the input changes from `choice` to `string`) and brings the self-hosted benchmark job to label-handling parity with the main test job.

One small follow-up of my own on top: `types: [opened, synchronize, reopened, labeled]` on the `pull_request:` trigger, so that applying a `container:main` (or `container:release`) label to a stacked PR re-runs CI without having to push a fresh commit. Useful for the gtrack and richards stacks where the base is another feature branch and the default image tag wouldn't otherwise exist.

Cherry-pick was clean modulo a trivial conflict on the second job's image line — main had since gained the same label-handling block, so I took Angus's `format()`-based version for both jobs.

Full credit to Angus for the original fix.


angus-g (commented):
Thanks for doing this, sorry I didn't get around to opening the PR myself! One possible addition: allow the `render.yml` workflow to use similar logic, so you can test notebook rendering from other containers? This would just be removing the `dev-` prefix and allowing a string base rather than option.

sghelichkhani:
> Thanks for doing this, sorry I didn't get around to opening the PR myself! One possible addition: allow the `render.yml` workflow to use similar logic, so you can test notebook rendering from other containers? This would just be removing the `dev-` prefix and allowing a string base rather than option.

Very good idea! Thanks Angus!

sghelichkhani:
One the GIA tests failed randomly. We are just making sure it's just the usual intermittent failures. Thanks @dsroberts  for rerunning.

dsroberts:
> Thanks for doing this, sorry I didn't get around to opening the PR myself! One possible addition: allow the `render.yml` workflow to use similar logic, so you can test notebook rendering from other containers? This would just be removing the `dev-` prefix and allowing a string base rather than option.

Second run passed, so this is good to go, but I do think this is worth adding to `render.yml` as part of this PR.

dsroberts (approved):

PR #522, opened 2026-05-22
Add soil curve models for variably saturated flow

sghelichkhani:
Second of five PRs decomposing the Richards-equation foundation work. Stacked on #521.

Three constitutive soil-water relations used in variably saturated flow:

- `HaverkampCurve`: rational forms for `theta(h)` and `K(h)`, Haverkamp (1977) parameters for the sand-clay column benchmark.
- `VanGenuchtenCurve`: van Genuchten (1980) / Mualem `theta(h)` and `K(theta)`, the de facto standard in unsaturated zone hydrology.
- `ExponentialCurve`: `K(h) = K_s * exp(alpha * h)`, used for the Tracy analytical infiltration benchmark in the follow-up PR.

All three derive from a common `SoilCurve` abstract base class so downstream code can treat them uniformly. Pure constitutive models with no solver dependency; sixteen unit tests cover parameter validation, the saturated/residual limits, the `theta`-`h`-`K` consistency relations, and round-trip behaviour for `h(theta)` where applicable.

### DIFF
--- a/gadopt/soil_curves.py
+++ b/gadopt/soil_curves.py
@@ -52,6 +52,12 @@ class SoilCurve(ABC):
     #: Default dimensionless smoothing of the saturation magnitude near h = 0.
     DEFAULT_REG_EPS = 1e-6
 
+    #: Parameters required by every soil model, validated in the base class.
+    BASE_REQUIRED_PARAMS = ('theta_r', 'theta_s', 'Ks', 'Ss')
+
+    #: Model-specific required parameters; each subclass lists its own.
+    REQUIRED_PARAMS = ()
+
     def __init__(self, theta_r, theta_s, Ks, Ss, reg_eps=DEFAULT_REG_EPS, **kwargs):
         """
         Initialise soil curve with model parameters.
@@ -134,9 +140,23 @@ class SoilCurve(ABC):
         eps = self.reg_eps * scale
         return fd.sqrt(x * x + eps * eps)
 
-    @abstractmethod
     def _validate_parameters(self) -> None:
-        """Validate that all required parameters are provided."""
+        """Check all required parameters are present, then run model-specific checks.
+
+        The required set is the union of BASE_REQUIRED_PARAMS (common to every
+        model) and the subclass REQUIRED_PARAMS. Subclasses add range or
+        consistency checks by overriding _validate_model_parameters.
+        """
+        for param in (*self.BASE_REQUIRED_PARAMS, *self.REQUIRED_PARAMS):
+            if param not in self.parameters:
+                raise ValueError(f"Missing required parameter: {param}")
+        self._validate_model_parameters()
+
+    def _validate_model_parameters(self) -> None:
+        """Hook for model-specific validation (e.g. parameter ranges).
+
+        Default is a no-op; subclasses override to add their own checks.
+        """
         pass
 
     @abstractmethod
@@ -197,6 +217,8 @@ class HaverkampCurve(SoilCurve):
         Ss: Specific storage coefficient $[L^{-1}]$
     """
 
+    REQUIRED_PARAMS = ('alpha', 'beta', 'A', 'gamma')
+
     @property
     def alpha(self):
         """Haverkamp alpha fitting parameter."""
@@ -216,13 +238,6 @@ class HaverkampCurve(SoilCurve):
     def gamma(self):
         """Haverkamp gamma fitting parameter."""
         return self.parameters['gamma']
-
-    def _validate_parameters(self) -> None:
-        """Validate Haverkamp model parameters."""
-        required_params = ['theta_r', 'theta_s', 'alpha', 'beta', 'Ks', 'A', 'gamma', 'Ss']
-        for param in required_params:
-            if param not in self.parameters:
-                raise ValueError(f"Missing required parameter: {param}")
 
     def _abs_h_theta(self, h):
         r"""Regularised $|h|$ for the moisture branch (head scale $\alpha^{1/\beta}$)."""
@@ -284,6 +299,8 @@ class VanGenuchtenCurve(SoilCurve):
         Ss: Specific storage coefficient $[L^{-1}]$
     """
 
+    REQUIRED_PARAMS = ('alpha', 'n')
+
     @property
     def alpha(self):
         """van Genuchten alpha parameter (inverse air-entry pressure)."""
@@ -294,14 +311,13 @@ class VanGenuchtenCurve(SoilCurve):
         """van Genuchten n parameter (pore-size distribution)."""
         return self.parameters['n']
 
-    def _validate_parameters(self) -> None:
-        """Validate van Genuchten model parameters."""
-        required_params = ['theta_r', 'theta_s', 'alpha', 'n', 'Ks', 'Ss']
-        for param in required_params:
-            if param not in self.parameters:
-                raise ValueError(f"Missing required parameter: {param}")
+    def _validate_model_parameters(self) -> None:
+        """Require n > 1 when n is a scalar Constant.
 
-        # Validate parameter ranges when n is a scalar Constant
+        A UFL-expression n (e.g. a spatially varying field or an inversion
+        control) is left to the caller, so the range guard only applies when n
+        is a scalar Constant.
+        """
         n = self.parameters['n']
         if isinstance(n, fd.Constant) and n.values()[0] <= 1.0:
             raise ValueError("Parameter n must be > 1.0")
@@ -378,17 +394,12 @@ class ExponentialCurve(SoilCurve):
         Ss: Specific storage coefficient $[L^{-1}]$
     """
 
+    REQUIRED_PARAMS = ('alpha',)
+
     @property
     def alpha(self):
         """Exponential alpha decay parameter."""
         return self.parameters['alpha']
-
-    def _validate_parameters(self) -> None:
-        """Validate exponential model parameters."""
-        required_params = ['theta_r', 'theta_s', 'alpha', 'Ks', 'Ss']
-        for param in required_params:
-            if param not in self.parameters:
-                raise ValueError(f"Missing required parameter: {param}")
 
     def moisture_content(self, h: fd.Function | ufl.core.expr.Expr) -> ufl.core.expr.Expr:
         """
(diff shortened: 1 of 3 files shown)


tests/unit/test_soil_curves.py
```diff
@@ -0,0 +1,385 @@
...
+        # Get unsaturated values
+        theta_h = evaluate_soil_curve(haverkamp, h_unsat, "moisture_content").dat.data[0]
+        theta_vg = evaluate_soil_curve(vg, h_unsat, "moisture_content").dat.data[0]
+        theta_exp = evaluate_soil_curve(exp, h_unsat, "moisture_content").dat.data[0]
+
+        K_h = evaluate_soil_curve(haverkamp, h_unsat, "hydraulic_conductivity").dat.data[0]
+        K_vg = evaluate_soil_curve(vg, h_unsat, "hydraulic_conductivity").dat.data[0]
+        K_exp = evaluate_soil_curve(exp, h_unsat, "hydraulic_conductivity").dat.data[0]
+
+        # Models should give different results (not testing specific values,
```
w3168: The philosophy of this test seems a bit funny :) Is it worth just checking the individual tests results?

gadopt/soil_curves.py
```diff
@@ -0,0 +1,345 @@
...
+
+    @property
+    def gamma(self):
+        """Haverkamp gamma fitting parameter."""
+        return self.parameters['gamma']
+
+    def _validate_parameters(self) -> None:
+        """Validate Haverkamp model parameters."""
+        required_params = ['theta_r', 'theta_s', 'alpha', 'beta', 'Ks', 'A', 'gamma', 'Ss']
+        for param in required_params:
```
w3168: Maybe the part checking for required params could be in the base class to reduce code repetition?

gadopt/soil_curves.py
```diff
@@ -0,0 +1,345 @@
...
+    def moisture_content(self, h: fd.Function | ufl.core.expr.Expr) -> ufl.core.expr.Expr:
+        """
+        van Genuchten moisture content relationship.
+
+        $\theta(h) = \theta_r + (\theta_s - \theta_r) / (1 + |\alpha h|^n)^m$
+        where $m = 1 - 1/n$
+        """
+        m = 1 - 1/self.n
+
+        theta = self.theta_r + (self.theta_s - self.theta_r) / ((1 + abs(self.alpha * h)**self.n)**m)
```
w3168: Presumably `abs` is not differentiable? Thought I would flag this for later adjoint applications in case it is important...

w3168 (commented):
Hi @sghelichkhani just a few questions. Looking good though!

gadopt/soil_curves.py
w3168: Out of interest how does Firedrake handle this to work with newton solves?

sghelichkhani: Good catch, and you're right to flag it. `abs` isn't differentiable at zero, and the real problem here isn't `abs` on its own but the fractional power of the magnitude sitting on top of it. UFL differentiates `|h|` to `sign(h)·dh` and `|h|^p` to `p·|h|^(p-1)·sign(h)`, so the moment a DOF lands exactly on the water table at h=0 you get `pow(0, p-1)` blowing up to Inf and then `sign(0)·Inf = 0·Inf = NaN`. The branch-wise `conditional` doesn't save you because the singular term lives inside the unsaturated branch, which is exactly what gets differentiated, and Firedrake's adjoint reassembles the very same `derivative(F, u)` and `derivative(F, m)`, so a NaN in the Jacobian shows up identically in the gradient. So for adjoint work it really does matter, as you suspected.

I've just pushed a commit that handles this by regularising the magnitude as `sqrt(h² + ε²)` with a `reg_eps` parameter (default 1e-6, dimensionless, scaled to the relevant head). That keeps the base bounded away from zero so every fractional power stays smooth, and I rewrote the `water_retention` derivatives by hand so they're the exact derivative of the regularised θ rather than the singular analytic form. The forward values are unchanged to O(reg_eps²) away from the water table, and you can set `reg_eps=0` to recover the exact-but-singular form if you ever want it. There's test coverage for all of this in `test_soil_curves.py`: one test proves the value and the UFL derivative stay finite at h=0, a complementary regression test confirms the old `reg_eps=0` form really does go NaN there, and a Taylor-style test checks the UFL derivative matches a finite difference to second order on an unsaturated state. The one honest caveat is that K picks up an O(reg_eps) offset right at h=0; θ and C are clean to O(reg_eps²).

tests/unit/test_soil_curves.py
sghelichkhani: You were right, that one wasn't earning its place. It just re-ran saturation checks the per-model tests already did. Looking closer, a lot of the file was tautological, and the derivative test differentiated each expression against itself, so a wrong formula could pass. I've reworked it: values now check against an independent numpy reimplementation of the published closed forms (anchored to the 1980 paper's Fig 1/3 for van Genuchten), C is finite-differenced against theta, plus coverage of the seam, heterogeneous parameters and Ss routing. Mutation-tested to confirm it now catches a wrong formula.

There's also a real pyadjoint Taylor test now, checking second-order gradient convergence over the head field, so the adjoint path through the regularised curves is actually exercised (the regularisation rationale itself is in the thread on the abs comment above). All in the latest push.

gadopt/soil_curves.py
sghelichkhani: Good call, done. The base class now drives the missing-parameter check from a `REQUIRED_PARAMS` tuple per model plus the shared base params, so each model only lists its own and there's one place raising the error. The van Genuchten `n > 1` guard moved into an overridable hook so model-specific checks still have a home. Latest push.

sghelichkhani:
@w3168 whenever you get a chance, could you give this one a green light? All three of your comments are addressed and resolved. The test file now checks against numpy with a real pyadjoint Taylor test. All items by Angus are now addressed. This is the bottom of the Richards branch so it could get a green tick so we can focus on the rest.

tests/unit/test_soil_curves.py
```diff
@@ -0,0 +1,548 @@
...
+        )
+
+        model = _make_curve(cls, reg_eps=SoilCurve.DEFAULT_REG_EPS)
+        mesh = fd.UnitIntervalMesh(4)
+        V = fd.FunctionSpace(mesh, "CG", 1)
+        x, = fd.SpatialCoordinate(mesh)
+
+        # Strictly unsaturated head field (range -0.3 .. -1.3), set before
+        # annotation so it is a clean leaf control.
+        h = fd.Function(V).interpolate(-0.3 - x)
```
w3168, line 440: Did you need to keep h unsaturated for the taylor test to work? 

I'm guessing that if you cross the water table the conditional statements will break the adjoint...? Anyway I don't think we need to add more premature tests now if you are have not got adjoint results for your paper...

tests/unit/test_soil_curves.py
```diff
@@ -0,0 +1,548 @@
...
+
+        tape = get_working_tape()
+        tape.clear_tape()
+        continue_annotation()
+        try:
+            J = fd.assemble(getattr(model, property_name)(h) * fd.dx)
```
w3168, line 446: maybe not important but dont we usually square the functional?

tests/unit/test_soil_curves.py
```diff
@@ -0,0 +1,548 @@
...
+            theta_r=0.10, theta_s=theta_s, Ks=1e-5, Ss=Ss, alpha=alpha, n=2.0,
+        )
+        return mesh, P, model, alpha, theta_s, Ss
+
+    def test_fields_pass_through_untouched(self):
+        """A Function parameter is stored as-is, never collapsed to a scalar."""
+        _, _, model, alpha, theta_s, Ss = self._vg_heterogeneous()
+        assert model.parameters["alpha"] is alpha
+        assert model.parameters["theta_s"] is theta_s
+        # Ss is routed through ensure_constant, which passes Functions through.
```
w3168, line 488: is this comment still valid?

gadopt/soil_curves.py
```diff
@@ -0,0 +1,422 @@
...
+            'Ks': Ks,
+            'Ss': Ss
+        }
+        params.update(kwargs)
+
+        # Wrap scalars in fd.Constant. Anything already a ufl Expr (a Function or
+        # Function(R), e.g. heterogeneous soil or an inversion control) passes
+        # through untouched.
+        self.parameters = {
+            key: value if isinstance(value, ufl.core.expr.Expr) else fd.Constant(value)
```
w3168, line 96: Could you reuse `ensure_constant(value)` here and avoid the if statement? (feel free to ignore)

gadopt/soil_curves.py
```diff
@@ -0,0 +1,422 @@
...
+    """
+    Haverkamp soil curve model.
+
+    The Haverkamp model provides empirical relationships for soil-water
+    retention and hydraulic conductivity based on fitting parameters.
+
+    Parameters:
+        theta_r: Residual water content $[L^3/L^3]$
+        theta_s: Saturated water content $[L^3/L^3]$
+        alpha: Fitting parameter $[L^{-\\beta}]$
```
w3168, line 205: are the units of alpha right? maybe `L^\beta`

gadopt/soil_curves.py
```diff
@@ -0,0 +1,422 @@
...
+
+        Returns:
+            Volumetric moisture content $\theta$ $[L^3/L^3]$
+        """
+        pass
+
+    @abstractmethod
+    def hydraulic_conductivity(self, h: fd.Function | ufl.core.expr.Expr) -> ufl.core.expr.Expr:
+        """
+        Calculate relative hydraulic conductivity $K(h)$.
```
w3168, line 171: is the `relative` right here? maybe just hydraulic conductivity if it is dimensional

w3168 (changes requested):
Thanks @sghelichkhani for making the changes a while ago. Sorry I didn't get back to this before. 

I think the tests look better now. I found a few more small things (also with the help of chatgpt)  and messaged you offline about the regularisation. I think the regularisation is fine as is but chatgpt suggested regularising the exponent directly.

w3168:
Hi Sia just checking if you want another review of this yet?

PR #523, opened 2026-05-22
Add Richards equation and core solver

sghelichkhani:
Third of five PRs decomposing the Richards-equation foundation work. **This is the headline PR**: equations, solver, and a real end-to-end test. Stacked on #522.

The Richards equation governs water movement in variably saturated porous media,

```
(S_s * S(h) + C(h)) * dh/dt - div(K(h) * grad h) - div(K(h) * grad z) = 0
```

with `h` the pressure head, `theta(h)` the moisture content, `K(h)` the hydraulic conductivity, and `z` the vertical coordinate.

**`gadopt.richards_equation`** provides the Richards-specific residual terms (mass and gravity) as plain UFL callables. Diffusion and source are reused from `scalar_equation` via the `nonlinear_coefficients` mechanism from #521: `K(h)` is built as a UFL expression in the equation's `solution` and is resolved against each Irksome stage's trial via `ufl.replace`.

**`gadopt.RichardsSolver`** drives the equation in time. Two PETSc presets cover the common geometries:

- `direct_richards_solver_parameters`: MUMPS LU, the 2-D default.
- `iterative_richards_solver_parameters`: Newton + GMRES + BoomerAMG via Hypre, the 3-D default. Fails fast at construction if PETSc was not built with Hypre.

The default time discretisation uses Irksome's stage-value stepper (`stage_type='value'`), which treats the nonlinear mass term `Dt(theta(h))` as the exact finite difference rather than applying the chain rule. Mass-conservative for any Butcher tableau, both stiffly- and non-stiffly-accurate (the latter via Irksome's conservative-update path). Users who opt out get a runtime warning that the chain-rule discretisation accumulates mass-balance error.

The SIPG penalty assembly uses Hillewaert's `C(p)` trace constant to match the momentum equation. Because `scalar_equation.diffusion_term` assembles with the Shahbazi `C(p-1)` form internally, `RichardsSolver` multiplies the user-facing safety factor by the `C(p)/C(p-1)` ratio so the assembled penalty matches the Hillewaert convention. Flipping `scalar_equation` to `shift=0` would let us delete this compensation, but would force regenerating stored reference data on every DG scalar diffusion test (`viscoplastic_case_DG`, `multi_material`, `2d_cylindrical_TALA_DG`, `adjoint_2d_cylindrical`, and the adjoint mantle-convection demos), so the compensation lives in `RichardsSolver` instead.

**Tracy (2006) 2-D infiltration benchmark** lands under `tests/richards/`. Tracy provides a closed-form steady-state solution for the exponential-K model; the test verifies L2 convergence at DQ degrees 0, 1, 2 (expected `O(h^(p+1))`) for both Dirichlet specified-head and no-flux boundary configurations.

### DIFF
--- a/gadopt/__init__.py
+++ b/gadopt/__init__.py
@@ -23,6 +23,11 @@ from .level_set_tools import (
 from .limiter import VertexBasedP1DGLimiter
 from .nullspaces import create_stokes_nullspace, rigid_body_modes
 from .preconditioners import FreeSurfaceMassInvPC, SPDAssembledPC
+from .richards_solver import (
+    RichardsSolver,
+    direct_richards_solver_parameters,
+    iterative_richards_solver_parameters,
+)
 from .soil_curves import (
     SoilCurve,
     HaverkampCurve,
--- a/gadopt/__init__.py
+++ b/gadopt/__init__.py
@@ -23,11 +23,7 @@ from .level_set_tools import (
 from .limiter import VertexBasedP1DGLimiter
 from .nullspaces import create_stokes_nullspace, rigid_body_modes
 from .preconditioners import FreeSurfaceMassInvPC, SPDAssembledPC
-from .richards_solver import (
-    RichardsSolver,
-    direct_richards_solver_parameters,
-    iterative_richards_solver_parameters,
-)
+from .richards_solver import RichardsSolver
 from .soil_curves import (
     SoilCurve,
     HaverkampCurve,
--- a/gadopt/richards_equation.py
+++ b/gadopt/richards_equation.py
@@ -1,0 +1,157 @@
+r"""Richards equation terms for unsaturated flow.
+
+This module implements the individual terms for the Richards equation for
+variational unsaturated flow. The Richards equation describes movement of
+the wetting phase in a two-phase flow system, if we ignore the
+compressibility of the dry phase.
+
+All terms are considered as if they were on the left-hand side of the equation,
+leading to the following UFL expression returned by the ``Equation``'s
+``residual`` method:
+
+$$
+  \frac{\partial \theta}{\partial t} + F(h) = 0.
+$$
+
+The Richards equation in mixed form:
+
+$$
+  \frac{\partial \theta}{\partial t} + S_s S \frac{\partial h}{\partial t}
+  - \nabla \cdot (K \nabla h) - \nabla \cdot (K \nabla z) = 0
+$$
+
+where:
+- $h$ is the pressure head
+- $S_s$ is the specific storage coefficient
+- $S(h) = (\theta - \theta_r)/(\theta_s - \theta_r)$ is the effective saturation
+- $C(h) = d\theta/dh$ is the specific moisture capacity
+- $K(h)$ is the hydraulic conductivity
+- $z$ is the vertical coordinate (gravity term)
+
+"""
+
+from firedrake import *
+from irksome import Dt
+from ufl.indexed import Indexed
+
+from .equations import Equation
+from .utility import is_continuous
+
+__all__ = [
+    "richards_mass_term",
+    "richards_gravity_term",
+]
+
+
+def richards_mass_term(
+    eq: Equation, trial: Argument | Indexed | Function
+) -> Form:
+    r"""Richards equation mass term for moisture storage.
+
+    The mass term accounts for water storage changes and has two parts:
+
+    $$
+    \frac{\partial \theta}{\partial t} + S_s S \frac{\partial h}{\partial t}
+    $$
+
+    The moisture content change d(theta)/dt is expressed as Dt(theta(h)).
+    When used with a StageValueTimeStepper, this is automatically discretised
+    as a conservative finite difference (theta(h_new) - theta(h_old))/dt
+    rather than the chain-rule expansion C(h)*dh/dt, preserving exact mass
+    conservation for DG discretisations.
+
+    The specific storage term Ss*S*dh/dt uses Dt(h) since it is already
+    formulated directly in terms of h.
+
+    Args:
+        eq: G-ADOPT Equation instance
+        trial: Firedrake trial function for pressure head
+
+    Returns:
+        UFL form for the mass term
+    """
+    soil_curve = eq.soil_curve
+
+    # Moisture content change: d(theta)/dt
+    F = inner(eq.test, Dt(soil_curve.moisture_content(trial))) * eq.dx
+
+    # Specific storage: Ss * S * dh/dt (standard Dt, linear in h for given S).
+    # Only include when Ss is nonzero to avoid polluting the form with a
+    # zero-coefficient TimeDerivative that can confuse the nonlinear solver.
+    if float(soil_curve.Ss) != 0.0:
+        theta = soil_curve.moisture_content(trial)
+        S = (theta - soil_curve.theta_r) / (soil_curve.theta_s - soil_curve.theta_r)
+        F += inner(eq.test, soil_curve.Ss * S * Dt(trial)) * eq.dx
+
+    return F
+
+
+def richards_gravity_term(
+    eq: Equation, trial: Argument | Indexed | Function
+) -> Form:
+    r"""Richards gravity term for gravity-driven flow with upwinding.
+
+    The gravity term represents gravity-driven water flow:
+
+    $$
+    \\nabla \\cdot (K \\nabla z) = \\frac{\\partial K}{\\partial z}
+    $$
+
+    where $z$ is the vertical coordinate. For DG discretizations, we use upwinding
+    on interior facets to ensure stability.
+
+    The weak form is:
+
+    $$
+    \\int_\\Omega K \\frac{\\partial \\phi}{\\partial z} dx
+    - \\int_{\\mathcal{I}} \\text{jump}(\\phi) \\cdot (q_n^+ - q_n^-) dS
+    $$
+
+    where $q_n = 0.5(q \\cdot n + |q \\cdot n|)$ is the upwinded flux and
+    $q = K \\nabla z$ is the gravity-driven flux.
+
+    Args:
+        eq: G-ADOPT Equation instance
+        trial: Firedrake trial function for pressure head
+
+    Returns:
+        UFL form for the gravity term
+    """
+    soil_curve = eq.soil_curve
+
+    # Evaluate hydraulic conductivity at trial function
+    K = soil_curve.hydraulic_conductivity(trial)
+
+    # Get mesh dimension and spatial coordinates
+    dim = eq.mesh.geometric_dimension
+    x = SpatialCoordinate(eq.mesh)
+
+    # Gradient in vertical direction (last coordinate)
+    # grad(z) = e_z, where e_z is the unit vector in z-direction
+    n_down = grad(x[dim - 1])
+
+    # Volume integral: \int K * \partial v/\partial z dx = \int K * \grad z \cdot \grad v dx
+    F = inner(K * n_down, grad(eq.test)) * eq.dx
+
+    # Upwinding on interior facets for DG
+    if not is_continuous(eq.trial_space):
+        # Gravity-driven flux
+        q = K * n_down
+
+        # Upwinded flux: take the value from the upwind side
+        # q\cdot n > 0 means flow from '-' to '+', so use '-' side (upwind)
+        # q\cdot n < 0 means flow from '+' to '-', so use '+' side (upwind)
+        q_n = 0.5 * (dot(q, eq.n) + abs(dot(q, eq.n)))
+
+        # Add upwind flux term
+        F -= jump(eq.test) * (q_n('+') - q_n('-')) * eq.dS
+
+    return F
+
+
+# Set required and optional attributes for each term
+richards_mass_term.required_attrs = {'soil_curve'}
+richards_mass_term.optional_attrs = set()
+
+richards_gravity_term.required_attrs = {'soil_curve'}
+richards_gravity_term.optional_attrs = set()
--- a/gadopt/richards_equation.py
+++ b/gadopt/richards_equation.py
@@ -28,19 +28,43 @@ where:
 - $K(h)$ is the hydraulic conductivity
 - $z$ is the vertical coordinate (gravity term)
 
+The discretisation deliberately carries the mass term as the conservative
+$\partial \theta / \partial t$ form rather than expanding it via the chain rule
+into $C(h)\,\partial h/\partial t$. With the stage-value stepper this yields the
+exact finite difference $(\theta_{new} - \theta_{old})/\Delta t$, which keeps mass
+balance exact for DG. As a consequence $C(h) = d\theta/dh$ is the physical moisture
+capacity listed above but never appears explicitly in the residual.
+
 """
+
+import numbers
 
 from firedrake import *
 from irksome import Dt
 from ufl.indexed import Indexed
 
 from .equations import Equation
-from .utility import is_continuous
+from .utility import is_continuous, upward_normal
 
 __all__ = [
     "richards_mass_term",
     "richards_gravity_term",
 ]
+
+
+def _is_scalar_zero(value) -> bool:
+    """Return True only when ``value`` is provably the scalar zero.
+
+    A plain Python number or a scalar ``Constant`` can be safely evaluated to a
+    float and compared against zero. Anything else (most importantly a ``Function``
+    or ``Function(R)``, which must never be float()'d) is treated as non-zero so
+    that a spatially varying coefficient or an inversion control is never dropped.
+    """
+    if isinstance(value, numbers.Real):
+        return float(value) == 0.0
+    if isinstance(value, Constant):
+        return float(value) == 0.0
+    return False
 
 
 def richards_mass_term(
@@ -76,9 +100,13 @@ def richards_mass_term(
     F = inner(eq.test, Dt(soil_curve.moisture_content(trial))) * eq.dx
 
     # Specific storage: Ss * S * dh/dt (standard Dt, linear in h for given S).
-    # Only include when Ss is nonzero to avoid polluting the form with a
-    # zero-coefficient TimeDerivative that can confuse the nonlinear solver.
-    if float(soil_curve.Ss) != 0.0:
+    # We skip this term when Ss is identically zero, to avoid polluting the form
+    # with a zero-coefficient TimeDerivative that can confuse the nonlinear solver.
+    # That skip is only safe for values we can provably evaluate to the scalar
+    # zero: a plain Python number or a scalar Constant. A Function or Function(R)
+    # (e.g. a spatially varying storage, or an inversion control) must never be
+    # float()'d, so we always include the term for those and never silently drop it.
+    if not _is_scalar_zero(soil_curve.Ss):
         theta = soil_curve.moisture_content(trial)
         S = (theta - soil_curve.theta_r) / (soil_curve.theta_s - soil_curve.theta_r)
         F += inner(eq.test, soil_curve.Ss * S * Dt(trial)) * eq.dx
@@ -91,24 +119,27 @@ def richards_gravity_term(
 ) -> Form:
     r"""Richards gravity term for gravity-driven flow with upwinding.
 
-    The gravity term represents gravity-driven water flow:
+    The gravity term represents the gravity-driven water flux:
 
     $$
-    \\nabla \\cdot (K \\nabla z) = \\frac{\\partial K}{\\partial z}
+    \nabla \cdot (K \nabla z)
     $$
 
-    where $z$ is the vertical coordinate. For DG discretizations, we use upwinding
+    where $z$ is the elevation (upward coordinate); $\nabla z$ is supplied by
+    ``upward_normal``, so the term is correct both in a Cartesian box and on the
+    sphere. (In a Cartesian box with $z$ the vertical coordinate this reduces to
+    the familiar $\partial K/\partial z$.) For DG discretizations, we use upwinding
     on interior facets to ensure stability.
 
     The weak form is:
 
     $$
-    \\int_\\Omega K \\frac{\\partial \\phi}{\\partial z} dx
-    - \\int_{\\mathcal{I}} \\text{jump}(\\phi) \\cdot (q_n^+ - q_n^-) dS
+    \int_\Omega K \frac{\partial \phi}{\partial z} dx
+    - \int_{\mathcal{I}} \text{jump}(\phi) \cdot (q_n^+ - q_n^-) dS
     $$
 
-    where $q_n = 0.5(q \\cdot n + |q \\cdot n|)$ is the upwinded flux and
-    $q = K \\nabla z$ is the gravity-driven flux.
+    where $q_n = 0.5(q \cdot n + |q \cdot n|)$ is the upwinded flux and
+    $q = K \nabla z$ is the gravity-driven flux.
 
     Args:
         eq: G-ADOPT Equation instance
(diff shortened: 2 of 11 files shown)


gadopt/richards_equation.py
```diff
@@ -0,0 +1,157 @@
...
+$$
+  \frac{\partial \theta}{\partial t} + S_s S \frac{\partial h}{\partial t}
+  - \nabla \cdot (K \nabla h) - \nabla \cdot (K \nabla z) = 0
+$$
+
+where:
+- $h$ is the pressure head
+- $S_s$ is the specific storage coefficient
+- $S(h) = (\theta - \theta_r)/(\theta_s - \theta_r)$ is the effective saturation
+- $C(h) = d\theta/dh$ is the specific moisture capacity
```
stephankramer, line 27: Does not appear in your equation?

gadopt/richards_equation.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    soil_curve = eq.soil_curve
+
+    # Moisture content change: d(theta)/dt
+    F = inner(eq.test, Dt(soil_curve.moisture_content(trial))) * eq.dx
+
+    # Specific storage: Ss * S * dh/dt (standard Dt, linear in h for given S).
+    # Only include when Ss is nonzero to avoid polluting the form with a
+    # zero-coefficient TimeDerivative that can confuse the nonlinear solver.
+    if float(soil_curve.Ss) != 0.0:
```
stephankramer: Is `Ss` always going to be a Constant? Might it not be space varying? Also, could it be control-dependent in an inversion set-up? In which case it needs to be a `Function(R)` which also doesn't allow evaluating to float

gadopt/richards_equation.py
```diff
@@ -0,0 +1,157 @@
...
+
+def richards_gravity_term(
+    eq: Equation, trial: Argument | Indexed | Function
+) -> Form:
+    r"""Richards gravity term for gravity-driven flow with upwinding.
+
+    The gravity term represents gravity-driven water flow:
+
+    $$
+    \\nabla \\cdot (K \\nabla z) = \\frac{\\partial K}{\\partial z}
```
stephankramer: why are the backslashes escaped. It's a raw (`r"..."`) string, so shouldn't be necessary?

gadopt/richards_equation.py
```diff
@@ -0,0 +1,157 @@
...
+
+    The weak form is:
+
+    $$
+    \\int_\\Omega K \\frac{\\partial \\phi}{\\partial z} dx
+    - \\int_{\\mathcal{I}} \\text{jump}(\\phi) \\cdot (q_n^+ - q_n^-) dS
+    $$
+
+    where $q_n = 0.5(q \\cdot n + |q \\cdot n|)$ is the upwinded flux and
+    $q = K \\nabla z$ is the gravity-driven flux.
```
stephankramer: same here

gadopt/richards_equation.py
```diff
@@ -0,0 +1,157 @@
...
+    # Evaluate hydraulic conductivity at trial function
+    K = soil_curve.hydraulic_conductivity(trial)
+
+    # Get mesh dimension and spatial coordinates
+    dim = eq.mesh.geometric_dimension
+    x = SpatialCoordinate(eq.mesh)
+
+    # Gradient in vertical direction (last coordinate)
+    # grad(z) = e_z, where e_z is the unit vector in z-direction
+    n_down = grad(x[dim - 1])
```
stephankramer: That's up, no? Also, can you not use `vertical_normal()`? Just in the off-chance you want to do this on the sphere :-)

gadopt/richards_equation.py
```diff
@@ -0,0 +1,157 @@
...
+    F = inner(K * n_down, grad(eq.test)) * eq.dx
+
+    # Upwinding on interior facets for DG
+    if not is_continuous(eq.trial_space):
+        # Gravity-driven flux
+        q = K * n_down
+
+        # Upwinded flux: take the value from the upwind side
+        # q\cdot n > 0 means flow from '-' to '+', so use '-' side (upwind)
+        # q\cdot n < 0 means flow from '+' to '-', so use '+' side (upwind)
```
stephankramer: This is a bit confusing. `n` does not have a fixed direction, there's a `n('+')` and a `n('-')` without opposing signs: pointing outwards from the cell on + or - resp. - with the orientation of the inbetween facet, i.e. which side is labeled + or - being arbitrary. So the expression `q\cdot n > 0` being positive _on one side of the facet_ (whether it's + or -) means that `q` is going out, i.e. we're on the upwind side. So below only one of `q_n('+')` and `q_n('-')` is zero. The minus sign is needed because `jump(eq.test)` is a bit weird: its sign depends on the arbitrary orientation of the facet as it gives `test_+ - test_-`, what we want is `test_up - test_down` so if `-` is up, we need an extra minus sign.

gadopt/richards_solver.py
```diff
@@ -0,0 +1,462 @@
+"""Solver class for the Richards equation governing variably saturated flow.
+
+This module provides the `RichardsSolver` class for solving the Richards equation,
+which describes movement of fluid phase in a two-phase flow system, if we ignore the
+compressibility of the dry phase. The solver uses time integrators and together with
```
stephankramer:
```suggestion
compressibility of the dry phase. The solver uses time integrators together with
```

gadopt/richards_solver.py
```diff
@@ -0,0 +1,462 @@
...
+    "direct_richards_solver_parameters",
+    "iterative_richards_solver_parameters",
+]
+
+
+_newton_common: dict[str, Any] = {
+    "snes_type": "newtonls",
+    "snes_linesearch_type": "bt",
+    "snes_rtol": 1e-8,
+    "snes_atol": 1e-8,
```
stephankramer: Do you always use the same units for this equation? Is `1e-8` always small?

gadopt/richards_solver.py
```diff
@@ -0,0 +1,462 @@
...
+
+
+direct_richards_solver_parameters: dict[str, Any] = {
+    "mat_type": "aij",
+    "snes_type": "newtonls",
+    "snes_linesearch_type": "bt",
+    "ksp_type": "preonly",
+    "pc_type": "lu",
+    "pc_factor_mat_solver_type": "mumps",
+    "snes_atol": 1e-15,
```
stephankramer: Why is this different when using direct - that's a bit unexpected

gadopt/richards_solver.py
```diff
@@ -0,0 +1,462 @@
...
+
+iterative_richards_solver_parameters: dict[str, Any] = {
+    "mat_type": "aij",
+    "ksp_type": "gmres",
+    "snes_linesearch_type": "bt",
+    "ksp_rtol": 1e-5,
+    "ksp_max_it": 200,
+    "ksp_gmres_restart": 30,
+
+    # BoomerAMG via Hypre -- strong classical AMG for scalar elliptic
```
stephankramer: your operator is not elliptic though :-)

gadopt/richards_solver.py
```diff
@@ -0,0 +1,462 @@
...
+"""
+
+
+class RichardsSolver(SolverConfigurationMixin):
+    """Advances Richards equation in time for variably saturated flow.
+
+    The Richards equation describes water movement in variably saturated porous media:
+
+    $$
+    (S_s S + C) \\frac{\\partial h}{\\partial t} - \\nabla \\cdot (K \\nabla h) - \\nabla \\cdot (K \\nabla z) = 0
```
stephankramer: Here you do need `\\` - but I prefer `r"""` with single `\`

gadopt/richards_solver.py
```diff
@@ -0,0 +1,462 @@
...
+        For scalar Richards the near-nullspace is the constants -- the
+        single mode the diffusion operator nearly annihilates. Providing
+        it to BoomerAMG / vlumping's coarse AMG ensures the prolongation
+        preserves constants, which is critical for optimal AMG convergence.
+        """
+        ones = Function(self.solution_space).assign(1.0)
+        basis = VectorSpaceBasis([ones])
+        basis.orthonormalize()
+        return basis
+
```
stephankramer: I'm surprised this is needed. I would have thought any amg preserves constant modes. Or is this just some AI "interpolation"? It's right that the constant 1 is the equivalent "near null space" for scalar Laplace, it's just that you normally wouldn't have to tell AMG that in the scalar case - unless you have some weird FEM basis in which the constant function is not represented by a discrete vector with constant coefficients I guesss....

gadopt/richards_solver.py
```diff
@@ -0,0 +1,462 @@
...
+        # though Hillewaert's coercivity proof itself assumes bounded
+        # diffusivity. Absorb the C(p)/C(p-1) ratio into the safety factor so
+        # the assembled penalty matches the Hillewaert convention; the
+        # user-facing `interior_penalty` stays a bare safety factor.
+        #
+        # Flipping scalar_equation to shift=0 would let us delete this block,
+        # but would force regenerating stored reference data on every DG scalar
+        # diffusion test (viscoplastic_case_DG, multi_material,
+        # 2d_cylindrical_TALA_DG, adjoint_2d_cylindrical, and the adjoint
+        # mantle-convection demos).
```
stephankramer: I don't know who I'm talking to here: if for some reason you think the polynomial order used in the penalty term is wrong - or should be configurable then it should be fixed in the diffusion term. Is this because we need to take the order of the diffusion expression into account? These comments are gobledygook to me: what does shift=-1 mean? what is the "Shahbazi floor"?

gadopt/richards_solver.py
sghelichkhani: Yeah, sorry @stephankramer  you're right. This essentially AI slop. Originally marked by my LLM as necessary for AMG to work, but is definitely not. 

I traced the hypre path and this is essentially a no-op the way we've set it up. For BoomerAMG PETSc only forwardsa near-nullspace to hypre through the GM/LN interpolation, which is gated behind `vec_interp_variant` (defaulting to 0) plus nodal coarsening. We set neither, so
PETSc copies the constant vector into hypre and hypre just drops it. 

See: [the near-nullspace is ignored by BoomerAMG unless you
also set `nodal_coarsen` and `vec_interp_variant`](https://gitlab.com/petsc/petsc/-/blob/c73c6e7a756c80f1d8c836a87416649adc13d07a/src/ksp/pc/impls/hypre/impl/ihypre.c#L2272-L2273).

But I am still correct about the plain `nullspace`/`transpose_nullspace` handling for the pure-Neumann case stays right?

gadopt/richards_solver.py
sghelichkhani: Ok! So this is following a long discussion I had with @w3168. The comment above, which @stephankramer finds confusing, was meant to be left there as a reminder that I think the way we define the SIPG lower bound coefficient for the scalar diffusion term is wrong, but working for whatever reason. So I was trying my best to avoid changing other parts of the code while I am dealing with a million PRs on my hand.  

Currently we use`scalar_equation.diffusion_term` in multiple places, which uses `interior_penalty_factor(shift=-1)`. This implies a -1 shift in the polynomial degree we are passing to the formulation of the lower bound. And this is by far the most we are using an SIPG term, by using it for energy, generic transport, diffusive smoothing, Richards. Note that Richards' uses the same term `scalar_equation.diffusion_term`. 

The other place that we use SIPG is in imposing weak boundary condition for velocity in Stokes, where we use `momentum_equation.viscosity_term` . There we are using `interior_penalty_factor(shift=0)`,  where we actually pass the actual polynomial degree without a -1 shift to the SIPG formula. 

So I agree with @stephankramer that this needs to be changed in `scalar_equation.diffusion_term`; but my thinking was that once the Richards module goes in that can be a separate PR to figure out why we are picking values just randomly without any explanation. 

Now I had a look at three references that we actually cite: 

- http://www.math.utah.edu/~epshteyn/penalty.pdf
- https://www.academia.edu/download/90419982/shahbazi_05.pdf

And the one that Shabazi cites as the formua https://infoscience.epfl.ch/bitstreams/a541b153-f562-48eb-baa1-b42f9335269d/download

  $$\sigma_f^* = \mu,\frac{(p+1)(p+d)}{d},\max_{e\supset f}\frac{1}{\mathcal V(e)} [...].$$


Which if I am not wrong is consistent with `shift=0` in gadopt. 

Now why is this important? Because Richards solver using DG2 with the default of `scalar_equation.diffusion_term` does not work for a few of my benchmarks.

sghelichkhani: And this has been inconsistent from day zero. This is May 2022. How we define sigma for scalar equation
 
https://github.com/g-adopt/g-adopt/commit/4d1860a02af390213bd241657fb6863046badecd#diff-09b6f8a3713915f82474ec20b725ae64e2a80aaa9ed76aa75912d8f936e5448eR120

This is how we define sigma for momentum: 
https://github.com/g-adopt/g-adopt/commit/4d1860a02af390213bd241657fb6863046badecd#diff-d4479f3306f4a479e474732394928e2af49f8f4df4bdbb92596bcaa3167ae5a2R76

@Patol75 Claude really wanted to blame this on you and #113 ! I am happy I double checked this!

gadopt/richards_solver.py
sghelichkhani: So far we have always used the dimensional form. So this has worked. The atol was introduced for the cases where we were close to steady state and did not want the solver to misbehave thinking it did not converge, but I realised now I need to use the stol for that. Added it now

gadopt/richards_solver.py
sghelichkhani: Fixed now.

gadopt/richards_solver.py
sghelichkhani: Correct! Since the benchmarks show good performance, I am assuming the comment I have now put there makes it correct :-)

gadopt/richards_solver.py
sghelichkhani: Corrected!

gadopt/richards_solver.py
stephankramer: I might be wrong, but you always have a time derivative, no? So then you never have a nullspace - even with pure Neumann?

gadopt/richards_solver.py
stephankramer: Ah ok, I just didn't have any idea what `shift=-1` meant (as this was indeed introduced in #113) - so I just assumed it was just AI talking gibberish at me. So if I understand correctly #113 introduced interior_penalty_factor() to avoid duplication in the penalty term between diffusion and viscosity (a good thing) but then it added this `shift=-1` in the diffusion-case to be consistent with the formula that was (wrongly) used before that? Then yes, sure, let's deal with it in a separate PR.

gadopt/richards_equation.py
sghelichkhani: Added a comment below here so that it is clear C is not beig used because of the mixed formulation. I thought it is cleaner to have the parameter there in the documentation

sghelichkhani: Please resolve this if happy!

gadopt/richards_equation.py
sghelichkhani: Fair point. I have now added more checks on this. Please see if you are happy and resolve if so.

gadopt/richards_equation.py
sghelichkhani: Resolved!

gadopt/richards_equation.py
sghelichkhani: Resolved!

gadopt/richards_solver.py
sghelichkhani: I need to think about this as the only remaining point! @MorrowL any thoughts?

gadopt/richards_solver.py
sghelichkhani: Good ! so resolved!

stephankramer:
@angus-g : could you take a look at the tests/ stuff in this?

gadopt/__init__.py
```diff
@@ -23,6 +23,11 @@
 from .limiter import VertexBasedP1DGLimiter
 from .nullspaces import create_stokes_nullspace, rigid_body_modes
 from .preconditioners import FreeSurfaceMassInvPC, SPDAssembledPC
+from .richards_solver import (
+    RichardsSolver,
+    direct_richards_solver_parameters,
+    iterative_richards_solver_parameters,
```
angus-g: We don't expose these for Stokes/energy solvers, maybe we shouldn't pollute the global namespace with these and use the same paradigm... Either use the string-valued presets, or if you need to override import from the module directly.

gadopt/richards_equation.py
```diff
@@ -0,0 +1,187 @@
...
+    A plain Python number or a scalar ``Constant`` can be safely evaluated to a
+    float and compared against zero. Anything else (most importantly a ``Function``
+    or ``Function(R)``, which must never be float()'d) is treated as non-zero so
+    that a spatially varying coefficient or an inversion control is never dropped.
+    """
+    if isinstance(value, numbers.Real):
+        return float(value) == 0.0
+    if isinstance(value, Constant):
+        return float(value) == 0.0
+    return False
```
angus-g: I guess this makes sense, but we only use it once for the case of `Ss` which is already forced to be a `Constant` or `Function` (twice!) during the curve initialisation.

gadopt/soil_curves.py
```diff
@@ -46,18 +61,28 @@ def __init__(self, theta_r, theta_s, Ks, Ss, **kwargs):
...
+        # Build parameters dictionary from positional and keyword arguments.
+        # Route Ss through ensure_constant so a bare float/int becomes a Constant
+        # while a Function or Function(R) (e.g. a spatially varying storage or an
+        # inversion control) passes through untouched and is never float()'d.
         params = {
             'theta_r': theta_r,
             'theta_s': theta_s,
             'Ks': Ks,
-            'Ss': Ss
+            'Ss': ensure_constant(Ss)
```
angus-g: This seems redundant with the block below that already wraps non-Expr terms in `Constant`. As a reminder, `ensure_constant` wraps `float` and `int` so the same outcome is achieved.

```python
def ensure_constant(f):
    if isinstance(f, float) or isinstance(f, int):
        return Constant(f)
    else:
        return f
```

gadopt/richards_solver.py
```diff
@@ -0,0 +1,454 @@
...
+            'source': self.source_term,
+        }
+
+        # scalar_equation.diffusion_term builds the SIPG penalty with shift=-1
+        # (the sharper C(p-1) trace constant), but we want C(p), which appears commonly
+        # in the literature (shift=0), as used by the momentum equation. As a stopgap we
+        # scale the safety factor by C(p)/C(p-1) here to cancel the shift back
+        # out. This is a hack; the proper fix is to default scalar diffusion to
+        # shift=0, deferred for now because it forces regenerating DG reference data.
+        alpha = self.interior_penalty if self.interior_penalty is not None else 2.0
```
angus-g, line 331:
```suggestion
        alpha = self.interior_penalty or 2.0
```

Maybe?

tests/richards/meta.py
```diff
@@ -0,0 +1,21 @@
+from .richards import cases
+
+steps = {}
+
+for l1, v1 in cases.items():
+    for l2, v2 in v1.items():
+        for l3, config in v2.items():
```
angus-g: I would prefer more descriptive names for these levels, as in the module itself.


```suggestion
for module_name, bc_dict in cases.items():
    for bc_type, degree_dict in bc_dict.items():
        for degree_key, config in degree_dict.items():
```

tests/unit/test_soil_curves.py
```diff
@@ -383,3 +435,125 @@ def test_unsaturated_behavior_differences(self):
...
+
+class TestRegularisationAdjointAccuracy:
+    """Finite-difference vs UFL-derivative agreement on an unsaturated state.
+
+    A lightweight stand-in for a full pyadjoint Taylor test: it confirms the
+    regularised constitutive output is differentiable and that the UFL
+    derivative (the object pyadjoint propagates) matches a central finite
+    difference to second order at a strictly unsaturated point (h < 0, away
+    from the water table).
+    """
```
angus-g: I find this a bit of a weird test. Also I'm potentially surprised that the expressions are differentiable? E.g. the properties are all `conditional` for `VanGenuchtenCurve` -- is this going to fall over when an _actual_ Taylor test is done?

angus-g (changes requested):
Thanks, couple of smallish things. I appreciate that you at least demonstrate that `reg_eps` is required during the unit testing, usually these things just test something obvious.

gadopt/soil_curves.py
```diff
@@ -90,6 +115,25 @@ def Ss(self):
         """Specific storage coefficient [1/L]."""
         return self.parameters['Ss']
 
+    def _reg_abs(self, x, scale=1):
```
w3168, line 121: Hi @sghelichkhani did you put this in because of my comment in #522 in the other pr? here: https://github.com/g-adopt/g-adopt/pull/522#discussion_r3345607403  

I think having the epsilon term probably is the right thing to do. Not sure if it should live here though? Maybe in gadopt's utility...

gadopt/richards_equation.py
```diff
@@ -0,0 +1,221 @@
...
+    # translates the user-facing 'h' in `set_boundary_conditions`), and its
+    # consistency term carries K grad(h).n only, so the K grad(z).n part is
+    # added here to complete the total flux. K is taken from the interior
+    # trace, which makes the pair a single Nitsche treatment of the potential
+    # h + z, and is identical to what setting `reference_for_diffusion = z`
+    # on the diffusion term would produce. Deliberately outside the DG branch
+    # above: it applies wherever a weak Dirichlet head exists, though in
+    # practice RichardsSolver only generates those for discontinuous spaces
+    # (continuous ones get a strong DirichletBC instead).
+    for bc_id, bc in eq.bcs.items():
```
sghelichkhani, line 197: @stephankramer After going through your comments I discovered this was a term that was missing from our implmementation! Now that you have read the equations, would be good if you have a look at this again.

gadopt/soil_curves.py
sghelichkhani: You're right, it does nothing. Dropped it. Note the soil curve files moved down the stack since you commented, so the fix is in #522.

gadopt/soil_curves.py
sghelichkhani: Left it here for now. One consumer, and the `scale` argument is soil-specific (a head scale like `1/alpha`), so it doesn't generalise cleanly yet. Easy to promote to utility the moment something else needs it. Note the soil curve files moved down the stack since you commented, so this lives in #522 now.

tests/unit/test_soil_curves.py
sghelichkhani: There's a real pyadjoint `taylor_test` now, over the head field as control, across both curves and all three properties, and it converges at second order. The `conditional` isn't the problem: UFL differentiates each branch separately, and the singular bit inside the unsaturated branch is exactly what `reg_eps` smooths.

gadopt/richards_solver.py
sghelichkhani: I think that is essentially any case that we'll run in the future. The Jacobian of the time term contributes `(C + Ss*S)/dt` times the mass matrix, and that keeps the constant out whatever the boundary conditions are. But for the speciall case the domain being fully saturated and `Ss = 0`, the mass term drops out entirely and we would have the saturated Darcy flow. With a flux condition on every boundary then constant  becomes a null space.

So I think you are correct in generall that in likelihood we would never use that. So  I've narrowed the docstring to spell out the three conditions.

gadopt/richards_equation.py
sghelichkhani: To save you re-deriving it: on a weakly imposed Dirichlet head the Nitsche consistency term that `scalar_equation.diffusion_term` supplies is built from `K grad(h).n` alone, so the elevation half was simply missing and the scheme was enforcing zero gravity flux through those boundaries. Adding `- phi K grad(z).n` completes it to the real total flux `K grad(h+z).n`. The symmetry and penalty terms need no change because `(h+z) - (h_D+z) = h - h_D`, so the pair ends up being one Nitsche treatment of the potential `h+z`.

It's deliberately not applied on flux boundaries or on boundaries left unspecified. Leaving it off there is what makes `flux` the total flux and makes an unspecified boundary no-flow, which are both the conditions you'd want. I've also documented the sign convention, since `flux` is minus the outward Darcy flux and a positive value means water entering.

gadopt/richards_solver.py
sghelichkhani: What actually decides convergence here is `snes_rtol` and `snes_stol`, and both are relative, so they're unit-free. `snes_atol` is down at `1e-12` purely as a floor, so it only fires when the residual is genuinely negligible

gadopt/richards_solver.py
sghelichkhani: I'd rather keep the explicit `is not None` here. `or` treats `0.0` as falsy, so anyone passing `interior_penalty=0` would silently get `2.0` instead.

gadopt/richards_equation.py
sghelichkhani: Both halves fixed.

gadopt/__init__.py
sghelichkhani: Agreed, that was inconsistent.

gadopt/richards_equation.py
sghelichkhani: Fixed, along with the one above it.

tests/richards/meta.py
sghelichkhani: Agreed! Thanks Angus!

PR #524, opened 2026-05-22
Add vertically lumped preconditioners and presets for Richards

sghelichkhani:
Fourth of five PRs decomposing the Richards-equation foundation work. Stacked on #523.

Two new Python preconditioners target high-aspect-ratio extruded meshes typical of water-table problems:

- **`VerticallyLumpedPC`**: two-level multigrid (Kramer et al. 2010). The fine operator is collapsed onto a 2D-like coarse problem via Galerkin projection; the coarse solve is MUMPS LU. Fine-level smoother is Chebyshev wrapping block-Jacobi ILU, which damps the stiff vertical modes cheaply and is insensitive to vertical resolution.
- **`VerticallyLumpedHMGPC`**: same two-level structure but the coarse solve descends a geometric multigrid on a 2D base `MeshHierarchy`. Fine-level smoother is `ASMLinesmoothPC` (exact per-column solves) for one-sweep damping of vertical modes. Requires the base mesh to live in a `MeshHierarchy` with at least one refinement level.

`RichardsSolver` gains two presets that wire these in (`vlumping`, `vlumping_hmg`) and `_auto_select_preset` now picks `vlumping` on 3-D extruded Cartesian meshes (where vertical anisotropy is most pronounced), falling back to `iterative` (BoomerAMG) on other 3-D meshes. `_validate_preset` fails fast when structural prerequisites aren't met: `vlumping` requires an extruded mesh; `vlumping_hmg` additionally requires the base mesh hierarchy.

Both presets bake in inexact Newton (`ksp_rtol=1e-4`), which consistently wins over `1e-5` in the Cockett and Murrumbidgee scaling studies; tighter linear solves don't reduce Newton counts.

The standalone PCs and the wired-in presets land together because vlumping is only used by Richards in practice, and the natural reproducer (`test_vlumping_variants.py`, carried from `richardson`) exercises the presets through `RichardsSolver` on tiny extruded problems (8x8 base, 4 layers, 2 steps) so the file stays serial-laptop friendly.

### DIFF
--- a/gadopt/preconditioners.py
+++ b/gadopt/preconditioners.py
@@ -1,10 +1,26 @@
 r"""This module contains classes that augment default Firedrake preconditioners.
 
+Includes preconditioners for:
+- Stokes: FreeSurfaceMassInvPC, SPDAssembledPC
+- Richards: VerticallyLumpedPC, VerticallyLumpedHMGPC
+
 """
 
 import firedrake as fd
-from ufl.indexed import Indexed
+from firedrake import (
+    FiniteElement,
+    FunctionSpace,
+    MixedFunctionSpace,
+    PCBase,
+    TensorProductElement,
+    TrialFunction,
+)
+from firedrake.assemble import assemble
+from firedrake.dmhooks import get_function_space
+from firedrake.interpolation import interpolate
+from firedrake.mg.utils import get_level
 from firedrake.petsc import PETSc
+from ufl.indexed import Indexed
 from .utility import InteriorBC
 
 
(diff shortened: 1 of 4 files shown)


gadopt/preconditioners.py
```diff
@@ -64,3 +80,320 @@ def initialize(self, pc: PETSc.PC):
...
+        Fine level   : V = V_horiz x V_vert  (full 3D tensor-product space)
+        Coarse level : V_c = V_horiz x R     (vertically constant)
+
+    The prolongation P is the natural injection from V_c into V (a coarse
+    DOF is replicated along its column). The coarse operator is formed by
+    Galerkin projection, A_c = P^T A P, collapsing the entire vertical
+    dimension onto a 2D-like problem solvable cheaply with MUMPS or GAMG.
+
+    Solver options for the coarse level use the prefix ``lumped_mg_coarse_``
+    and the fine-level smoother uses ``lumped_mg_levels_``.
```
stephankramer: I guess the double backtick is again escaping? Prefer `r"""`

gadopt/preconditioners.py
```diff
@@ -64,3 +80,320 @@ def initialize(self, pc: PETSc.PC):
...
+    def initialize(self, pc):
+        options_prefix = pc.getOptionsPrefix()
+        A, P = pc.getOperators()
+        V = get_function_space(pc.getDM())
+
+        # Handle both scalar and mixed spaces
+        if len(V) == 1:
+            V = FunctionSpace(V.mesh(), V.ufl_element())
+        else:
+            V = MixedFunctionSpace([V_ for V_ in V])
```
stephankramer: Why is this needed? How is the resulting `V` expected to be different from the `V` we get from the dm?

gadopt/preconditioners.py
```diff
@@ -64,3 +80,320 @@ def initialize(self, pc: PETSc.PC):
...
+
+        # Build vertically constant function space (V_horiz x R)
+        mesh = V.mesh()
+        _, vcell = mesh.ufl_cell().sub_cells
```
stephankramer: Maybe add a `try:`, `except AttributeError` with a helpful message as this will fail for non-extruded

gadopt/preconditioners.py
```diff
@@ -64,3 +80,320 @@ def initialize(self, pc: PETSc.PC):
...
+
+        # Build prolongation: interpolation from V_coarse to V
+        trial = TrialFunction(V_coarse)
+        Prol = assemble(interpolate(trial, V)).petscmat
+
+        # Set up 2-level multigrid
+        self.pc = PETSc.PC().create(comm=pc.comm)
+        self.pc.setOptionsPrefix(options_prefix + "lumped_")
+        self.pc.incrementTabLevel(1, parent=pc)
+        # Propagate the outer DM so level smoothers (e.g. ASMLinesmoothPC)
```
stephankramer: Nice!

gadopt/preconditioners.py
```diff
@@ -64,3 +80,320 @@ def initialize(self, pc: PETSc.PC):
...
+            raise RuntimeError(
+                "VerticallyLumpedHMGPC requires an extruded fine mesh "
+                "(mesh._base_mesh not found). Use VerticallyLumpedPC for "
+                "non-extruded meshes."
+            )
+
+        hierarchy, level = get_level(base_mesh)
+        if hierarchy is None or level is None:
+            raise RuntimeError(
+                "VerticallyLumpedHMGPC requires the fine mesh's base mesh "
```
stephankramer:
```suggestion
                "VerticallyLumpedHMGPC requires the fine mesh's 2D base mesh "
```

gadopt/preconditioners.py
```diff
@@ -64,3 +80,320 @@ def initialize(self, pc: PETSc.PC):
...
+        """Assemble horizontal interpolation matrices for the base MG."""
+        mats = []
+        for k in range(1, fine_level + 1):
+            mesh_c = hierarchy[k - 1]
+            mesh_f = hierarchy[k]
+            V_c = FunctionSpace(mesh_c, hele)
+            V_f = FunctionSpace(mesh_f, hele)
+            trial = TrialFunction(V_c)
+            P_k = assemble(interpolate(trial, V_f)).petscmat
+            mats.append(P_k)
```
stephankramer, line 815:
```suggestion
        V_c = FunctionSpace(hierarchy[0], hele)
        for k in range(1, fine_level + 1):
            mesh_f = hierarchy[k]
            V_f = FunctionSpace(mesh_f, hele)
            trial = TrialFunction(V_c)
            P_k = assemble(interpolate(trial, V_f)).petscmat
            mats.append(P_k)
            V_c = V_f  # fine becomes coarse in next round
```

gadopt/preconditioners.py
stephankramer: Same further down below: search for double backticks

gadopt/preconditioners.py
```diff
@@ -64,3 +80,320 @@ def initialize(self, pc: PETSc.PC):
...
+        I.setType("aij")
+        I.setPreallocationNNZ(1)
+
+        row_start, row_end = I.getOwnershipRange()
+
+        owned_pairs = {}
+        for c in range(Vc_cnm.shape[0]):
+            for k in range(Vc_cnm.shape[1]):
+                glob_R = int(lg_R[int(VR_cnm[c, k])])
+                glob_c = int(lg_c[int(Vc_cnm[c, k])])
```
stephankramer: Ouch, this looks expensive in python!

gadopt/preconditioners.py
```diff
@@ -64,3 +80,320 @@ def initialize(self, pc: PETSc.PC):
...
+    def _build_identity_coarse_to_R(V_coarse, V_R):
+        """Build the V_coarse -> V_R permutation matrix, parallel-safe.
+
+        ``cell_node_map().values_with_halo`` holds local DOF indices for
+        this rank (including ghost entries from halo cells). To insert
+        into a parallel PETSc Mat we translate local indices to global
+        via the ``dof_dset`` LGMap and only write rows this rank owns;
+        off-rank rows are covered by their owning rank.
+        """
+        import numpy as np
```
stephankramer: imports should (in principle) be on top

gadopt/preconditioners.py
```diff
@@ -64,3 +80,320 @@ def initialize(self, pc: PETSc.PC):
...
+                "VerticallyLumpedHMGPC requires the fine mesh's base mesh "
+                "to be part of a MeshHierarchy so the coarse PCMG can "
+                "descend a geometric hierarchy. Build the mesh with "
+                "ExtrudedMeshHierarchy(MeshHierarchy(base, L), ...) with "
+                "L >= 1, or fall back to VerticallyLumpedPC."
+            )
+
+        hele, _ = V.ufl_element().factor_elements
+        V_coarse = FunctionSpace(base_mesh, hele)
+        self.V_coarse = V_coarse
```
stephankramer: I find the name `V_coarse` a bit confusing. IIUC this is actually the *finest* horizontal mesh, right? I guess in the relationship between `V` (the top level full 3D function space) and this 2D one, `V` plays the role of "fine" and so this would be "coarse", but because you're here also building the 2D hierarchy it gets a bit confusing. So maybe `V_base_2D` or so?

gadopt/preconditioners.py
```diff
@@ -64,3 +80,320 @@ def initialize(self, pc: PETSc.PC):
...
+
+        # Compose the prolongation Prol = Q . I (see class docstring).
+        vcell = mesh.ufl_cell().sub_cells[1]
+        vele_R = FiniteElement("R", vcell, 0)
+        V_R = FunctionSpace(mesh, TensorProductElement(hele, vele_R))
+
+        trial_R = TrialFunction(V_R)
+        Q = assemble(interpolate(trial_R, V)).petscmat
+        I = self._build_identity_coarse_to_R(V_coarse, V_R)
```
stephankramer: same in the name and argument of this function: prefer 2D instead of coarse

stephankramer (changes requested):
This is really exciting, and most of the implementation looks good. The HMG variant is a nice idea. Initially I thought, well you do vertical lumping and then you just throw gamg at the horizontal, but if your horizontal mesh has a nice multigrid hierarchy it's not clear how you would do GMG in the horizontal, so this - with gmg-like fem interpolation operators and then doing a Galerkin projection to the coarser meshes - is a nice approach.

The thing I'm a little less enthusiastic about is `_build_identity_coarse_to_R()`. I suspect it might be quite slow (but maybe you care less if you only this once), but more importantly it's quite complex and touches deep into the Firedrake hierarchy with things like `V_R.dof_dset.lgmap.indices` - so I suspect for instance it will need to be rewritten when pyop3 lands (in a few months is the current planning).

So I'm wondering whether we could use the recent improvements in cross-mesh interpolation here, which allows interpolation between completely unrelated meshes. So I would have thought this should allow you to assemble an interpolation matrix between V_coarse (which I prefer to call V_2d) and V_R (the 3D fs extruded with R) - which should then give you the 1-to-1 mapping between nodes in V_2d and the corresponding column DOF in V_R. It depends a bit what that's doing, if it searches the location of the V_R dofs (which maybe located halfway down the column?) in the 2D mesh then maybe that doesn't work, but maybe if you ask for the interpolation matrix from V_R to V_coarse does work if that indeed searches mesh 2d physical node locations in the cells of V_R (which span the entire column) - and then you can just take the transpose (this doesn't work for general interpolation matrices, but does work for 1-to-1 mappings). You'd still want to do the same composing of that "identity" mapping and Q (the map from V_R to V) - as directly interpolating V_2d to V does not give what you want....

sghelichkhani:
Fair point Steph. So I think just a simple cross mesh interpolation does not work as the dimensions of the two meshes need to be the same.  both the cross-mesh and the newer submesh paths bail out unless the geometric (and topological) dimensions match so `interpolate(TrialFunction(V_R), V_coarse)` just raises straight away. 

Now doing a bit of AI assissted exploration: the relabelling here is the identity. `V_R = hele x Real` and `V_coarse = hele` on the base share the exact same dof layout, so the same-mesh matrix Q is already the prolongation we want. That is an extruded topology reuses the base mesh's `topology_dm`, renumbering and entity classes, and the Real vertical factor collapses the vertical extent away in the section and node-class computations, so V_R ends up numbered by the same computation over the same DM as V_coarse. I checked that the owned sizes, ownership ranges and the full lgmap including ghost ordering all agree in serial and in parallel, for CG and DG horizontals on both simplex and quad bases, and that the coarse operator and the KSP solve come out bit-identical to the old `Q . I` path. So the lucky outomce is that `_build_identity_coarse_to_R` is gone and the prolongation is just `assemble(interpolate(...)).petscmat`, which keeps us on public API that should survive pyop3. @stephankramer happy with that? I left a small guard that trips if the V_R/V_coarse layouts ever diverge, so a future change would fail loudly here instead of surfacing as something cryptic in the galerkin setup.

gadopt/preconditioners.py
sghelichkhani: Not sure maybe I am confused, but that's reST inline-literal markup for the PETSc option prefix, so it renders as code in the docs? It's what we use everywhere else we have 16 in `utility.py`. The backslash ones you flagged earlier were real though, and those are fixed.

gadopt/preconditioners.py
sghelichkhani: It wasn't needed, and the mixed branch was actually broken. For a scalar space the rebuild gives you back an identical space, so it did nothing. For a mixed space the result has no `factor_elements`, so it would have died two lines down at the element factorisation anyway. I've dropped both branches and replaced them with an explicit check that rejects a mixed space with a message pointing at fieldsplit.

gadopt/preconditioners.py
sghelichkhani: Done, wrapped both lookups since either can be the one that fails, and the message names the cell it actually got and points at `ExtrudedMesh`.

gadopt/preconditioners.py
sghelichkhani: Agreed.

gadopt/preconditioners.py
sghelichkhani: Taken as written

gadopt/preconditioners.py
sghelichkhani: that whole thing is gone.  replaced by a same-mesh interpolation

gadopt/preconditioners.py
sghelichkhani: No function-level imports left in the module.

gadopt/preconditioners.py
sghelichkhani: "coarse" was doing two jobs in the same file, coarse relative to the 3D space and coarse within the base hierarchy, and this one is the *finest* base level. Renamed to `V_base_2d` everywhere including the test, and I've put a line in the class docstring saying which sense of coarse is meant and that it sits at `level`, not at 0.

gadopt/preconditioners.py
sghelichkhani: That one was `_build_identity_coarse_to_R`, which went with the permutation rewrite. The surviving helper is `_build_base_hierarchy_interpolations`, no "coarse" in it.

PR #527, opened 2026-06-09
Use explicit Riesz maps for adjoint Controls

angus-g:
Because our adjoint examples are on non-moving meshes, we can construct the Riesz map ahead of time and cache the mass matrix. Otherwise, a RieszMap object is constructed on every derivative evaluation and incurs an LU factorisation and solve.

Something to consider for larger applications is to use an iterative rather than direct solver for this process.

If this is indeed a useful thing to do:
- [ ] add a statement to the accompanying documentation cells in the demos

### DIFF
--- a/demos/PDE_constrained_optimisation/PDE_constrained_boundary.py
+++ b/demos/PDE_constrained_optimisation/PDE_constrained_boundary.py
@@ -138,7 +138,8 @@ T0.project(T_wrong)
 # Note that L2 is the default Riesz map if none is explicitly specified.
 
 # +
-m = Control(T0, riesz_map="L2")
+riesz_map = RieszMap(Q, "L2", constant_jacobian=True)
+m = Control(T0, riesz_map=riesz_map)
 
 J = AdjFloat(0.0)  # Initialise functional
 factor = AdjFloat(0.5)  # First & final boundary integral weighted by 0.5 to implement mid-point rule time-integration.
--- a/demos/PDE_constrained_optimisation/PDE_constrained_field.py
+++ b/demos/PDE_constrained_optimisation/PDE_constrained_field.py
@@ -141,7 +141,8 @@ T0.project(T_target)
 # ensuring that gradients are computed in the $L^2$ Hilbert space with
 # respect to the inner product.
 
-m = Control(T0, riesz_map="L2")
+riesz_map = RieszMap(Q, "L2", constant_jacobian=True)
+m = Control(T0, riesz_map=riesz_map)
 
 # Based on our guess for the initial temperature, we run the model for a specified number of timesteps:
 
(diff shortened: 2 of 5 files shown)


w3168:
Thanks @angus-g. Would the default direct solver parameters have worked in 3d? I guess inverting a mass matrix with LU might be doable...

angus-g:
> Would the default direct solver parameters have worked in 3d?

Good question! Maybe, but in that case it definitely seems preferable to set the solver parameters manually.

stephankramer:
I think perceived wisdom says to solve a mass-matrix with CG + a "standard" cheap preconditioner. It's always a trivial solve so should converge in only a couple of iterations - so might be cheaper even in 2D. This is also the default solver parameters for projection (cg + icc) in Firedrake.

w3168:
I guess I was just surprised that @sghelichkhani didn't bump into this already in 3d in terms of memory? But maybe he did, as I'm pretty sure we've discussed the projection solver parameters in 3d before!

drhodrid:
I also have a recollection of Sia having to do something to work around these default solver parameters…

sghelichkhani:
Yes! So @drhodrid is correct that the default solver parameters for riesz mapping is one of the issues that I currently get around by changing FD's default solver parameters. There is this problem with the riesz representation, and the other is that the cloned projector object drops the solver parameters and uses the FD's default parameters (direct) instead of the projector's default which is gc+icc as @stephankramer mentions. (I am going to put a PR for that imminently)

In that spirit, my original thought was that this should be a fix in FD's side: riesz_representation should have the same default solve_parameters as projection, which would be gc+icc. So instead of changing this here, that would be in main FD?  Unless the H1 projection would not work nicely with gc+icc?

sghelichkhani:
since H1's `RieszMap._inner_product_form` is still SPD I am assuming the gc part would be OK, but not sure if icc a good pre-conditioner. probably fine @stephankramer ?

sghelichkhani:
What if a `Control` lives on a mixed space (the Stokes-type cases)? Does that even work? or should we handle mat_type first too?

stephankramer:
Yeah, unfortunately H1 requires more thought. It's a bit funny because its conditioning depends on your choice of length units: (1+1/dx^2) / (1 + 1/L^2) - which is to say that using standard "H1" (without a coefficient in front of the derivative term) you should really non-dimensionalise. In any case, for dx -> 0, it's still going to be 1/dx^2 i.e. scale like Poisson where multigrid would be the most appropriate, which unfortunately is always a bit problem dependent (e.g. amg vs. gmg). Maybe CG+icc is still a reasonable default? For large 3D problems the user would have to replace it with their multigrid of choice.

But yeah, I agree we should try to discuss this on the FD side, as the current defaults are probably not ideal

angus-g:
I think we've resolved to merge this as is, and push the solver parameter discussion to later (if ever): this PR mostly demonstrates that we *can* build the explicit map; if you need non-default parameters this gives a schematic to do so.

sghelichkhani (commented):
LGTM

sghelichkhani (approved):

PR #533, opened 2026-08-02
Generalise nodal control-volume measures across mesh types

drhodrid:
Add reusable routines for calculating nodal control-volume measures on
Firedrake meshes and expose them through the public G-ADOPT API.

The original layerwise control-volume algorithm and prototype code were
provided by Stephan Kramer. I subsequently updated and adapted that code,
then used GPT/Codex to help refactor and generalise the implementation
across different geometries, dimensions, element layouts and mesh types.
The resulting code was independently reviewed, corrected, optimised and
tested locally.

Add two public routines:

- nodal_control_volumes() calculates nodal measures over a complete mesh.
- layerwise_nodal_control_volumes() calculates horizontal nodal lengths or
  areas at every level of a non-periodically extruded mesh.

Support:

- 2-D and 3-D Cartesian meshes;
- triangular, quadrilateral, prismatic, tetrahedral and hexahedral cells;
- cylindrical and spherical geometries;
- embedded circle and spherical-manifold meshes;
- extruded and non-extruded meshes;
- CG1 and CG2 nodal layouts;
- meshes reconstructed from checkpoints or copied coordinate fields; and
- serial and distributed-memory MPI execution.

For degree-two fields, use a positive P1-iso-P2 control-volume
construction. Preserve the source coordinate degree so that requesting a
CG1 result does not reduce the order of curved geometry.

For radially extruded meshes, calculate layer geometry from the actual
coordinates. This accounts for both increasing and decreasing radius.
Cylindrical layer measures therefore scale with radius, while spherical
layer areas scale with radius squared.

Layer totals represent the measure of the discrete finite-element
geometry. Curved spherical totals converge to 4*pi*r**2 but are not
silently normalised to the analytical sphere area.

Optimise translated and homothetically scaled layers by assembling the
first layer once and scaling its nodal measures. Detect warped layers and
assemble those independently. Use a round-off-level geometric criterion
so small deformations and meshes located far from the origin cannot be
incorrectly classified as homothetic.

Stream Q1/Q2 vertical coordinate levels in bounded blocks instead of
constructing a complete target-degree coordinate field. This reduces peak
memory substantially for high-resolution spherical meshes. Reuse the
assembled Cofunction for warped layers and fill homothetic output levels
contiguously.

Add validation and regression tests covering:

- full-domain and layerwise measures;
- Cartesian, cylindrical and spherical cases;
- outward and inward radial extrusion;
- analytical layer scaling and discrete surface totals;
- independently assembled pointwise values;
- physical-coordinate-based horizontal DOF mapping;
- warped and nearly homothetic layers;
- large coordinate offsets;
- checkpoint loading and PVD output;
- reconstructed extruded meshes;
- positive nodal measures;
- assembly reuse and fast-path behaviour; and
- execution with two MPI ranks.

Also add a short spherical-checkpoint example that calls the reusable
routine and writes cv_area_3d.pvd.

The focused test suite passes in serial and on two MPI ranks. The code was
also checked with Ruff, Flake8 and Python compilation.


sghelichkhani:
Just a few obvious things upfront: 
1 - Testing goes into doit and not in the workflow directly.
2 - Demos, need a lot of wiring (to be added to testing and rendering). 
3 - And demos are percentage python files. 
4 - The tests have a lot of visualisations in them! Redundant and harmful to testing TBH. 
Will look at the actual code/math next.

sghelichkhani:
I think the core problem is that this doesn't really use the FEM machinery. It reshapes the coordinate array into (column, level), slices horizontal levels out by hand, and rebuilds a base mesh for each one. That only works when every base entity carries exactly one coordinate node per level. As soon as the base mesh has degree 3 geometry an edge carries two, the block is twice as long, and the reshape interleaves them. 

On a `CircleManifoldMesh(16, radius=1.2, degree=3)` extruded radially the level totals come out as 21.2, 43.8, 24.1 where they should be 7.54, 10.68, 13.82. Nothing raises, everything stays positive, and the numbers look plausible. Degree 4 is wrong too. The only guard we have is the shape comparison before `result.dat.data[:] = ...`, and it can't catch this: at degree 3 the iso and CG spaces have identical dof counts and disagree about where 32 of 49 nodes are. 

**the good part**
For degree 1 none of this is needed. The horizontal levels of a CG1 field on an extruded mesh are exactly the mesh's horizontal facets, so the whole thing is one form: 

```python
v = TestFunction(FunctionSpace(mesh, "CG", 1))
assemble(v * ds_b + v * ds_t + avg(v) * dS_h)
```

`ds_b` and `ds_t` pick up the two boundary levels and `dS_h` visits every interior interface once. (Note that the `avg` matters). For a continuous space the two restrictions on a horizontal facet land on the same global dofs, so `v('+') + v('-')` counts every interior level twice: on a unit cube with three layers you get 1, 2, 2, 1 instead of 1, 1, 1, 1. With `avg` it agrees with the current code to 2e-13 on a spherical shell, and it costs about the same as `assemble(Constant(1)*dx)` on the same mesh, which is the cheapest thing that can touch every cell.  The form also gets curved geometry of any degree right and handles periodic extrusion, **both for free**, because it never looks at the array layout. 

**the bad news is that this does not work for CG2** And I think the issue is a bug (I think) in  FIAT.

PR #534, opened 2026-08-05
Key the interpolation-geometry cache on the config by value

sghelichkhani:
**Stack:** **#534** ← #535 ← #536.  This PR is a standalone bugfix on `main`; the two above it build on it.

---

**Base: `main`.** A standalone bugfix in already-merged code (#516), independent of
the refactor above it.

`ScalarFieldConnector` shares one kNN interpolation-geometry bundle across the
outputs that hang off a single source. That cache was keyed on `id(config)`, which
is unique only among *live* objects: once an `InterpolationConfig` is garbage
collected its address is freed, and the next allocation can inherit the geometry
built for the old one — a different kernel or bandwidth, silently, with no error.

This keys the cache on the config *by value*: `InterpolationConfig` is now frozen
(so it hashes on its fields, and `__post_init__` validation can't be bypassed by
post-hoc assignment), and the cache key is the config object itself. Two configs
holding the same numbers correctly share one build; a collected config can no
longer hand its geometry to an unrelated one.

Ships with a deterministic positive test: one source shared by two connectors
holding equal-valued but distinct configs ends with a single cached geometry
(two under `id()` keying), plus companions covering distinct kernels, frozen-ness,
and hashability. No plate data or Firedrake solve required.

---


gadopt/gplates/connectors.py
```diff
@@ -251,7 +257,18 @@ def _compute(
...
+        # be `id(config)`, which is unique only among LIVE objects — a config
+        # that has been garbage collected frees its address for the next
+        # allocation, so an identity key could silently hand a new config the
+        # geometry built for an old one, with the wrong kernel or bandwidth and
+        # no error. By value is also the honest key, since the geometry depends
+        # on nothing else: two configs holding the same numbers should share one
+        # build. The coords half is a content HASH rather than the content, so
+        # it is collidable in principle; at 2^-64 that is not worth addressing,
+        # but it is not the same guarantee.
+        key = (hash(target_coords.tobytes()), self.interpolation)
```
sghelichkhani: I am removing this since based on what I understood id is unstable. if object A is collected and then a new object with the same ide is allocated (what are the chances?!) then id(B) == id(A). Making it frozen allows us to use value, instead of id to check self.interpolation

dsroberts: Unlikely but possible. I think the better motivation for this change is that your `InterplationConfig` undergoes validity checking on creation, and having it frozen prevents a user from doing anything that compromises its initial validity.

tests/unit/test_gplates_interpolation_config.py
```diff
@@ -0,0 +1,196 @@
...
+        # a field holding a list, dict or set would raise here, and would raise
+        # again on every geometry-cache insertion.
+        hash(InterpolationConfig())
+        hash(InterpolationConfig(kernel="gaussian", k_neighbors=4))
+
+    def test_equal_values_hash_equal(self):
+        assert hash(InterpolationConfig()) == hash(InterpolationConfig())
+        assert InterpolationConfig() == InterpolationConfig()
+
+
```
dsroberts, line 155: Is it really up to us to test that built-in python functionality does what it says it does?

tests/unit/test_gplates_interpolation_config.py
```diff
@@ -0,0 +1,196 @@
...
+# ---------------------------------------------------------------------------
+# Geometry cache keyed by value
+# ---------------------------------------------------------------------------
+
+class TestGeometryCacheKey:
+    def test_equal_configs_share_one_geometry(self):
+        # Two distinct config objects holding identical values. Under the old
+        # id()-based key this produces two cache entries and two cKDTree
```
dsroberts: Do not reference code that no longer exists in comments

tests/unit/test_gplates_interpolation_config.py
```diff
@@ -0,0 +1,196 @@
...
+
+        field_a = conn_a.get_indicator(targets, ndtime=0.0)
+        field_b = conn_b.get_indicator(targets, ndtime=0.0)
+
+        assert len(source._interp_geometry_cache) == 1
+        np.testing.assert_array_equal(field_a, field_b)
+
+    def test_differing_configs_get_their_own_geometry(self):
+        # The half that catches the original symptom: two configs that differ
+        # must not share a build, and the fields they produce must differ.
```
dsroberts: What?

tests/unit/test_gplates_interpolation_config.py
```diff
@@ -0,0 +1,196 @@
...
+        xyz = rng.normal(size=(n_points, 3))
+        self.xyz = xyz / np.linalg.norm(xyz, axis=1)[:, np.newaxis]
+        self.thickness = rng.uniform(50.0, 200.0, size=n_points)
+        self.gplates_connector = FakeGplatesConnector()
+        self.comm = MPI.COMM_SELF
+        self._is_root = True
+        self.compute_calls = 0
+
+    def _compute_sources(self, age):
+        self.compute_calls += 1
```
dsroberts: Consider using `mock` for these sorts of things. Have a look in `test_diags.py` for some examples.

tests/unit/test_gplates_interpolation_config.py
```diff
@@ -0,0 +1,196 @@
...
+    Args:
+        n_points: number of target nodes.
+        seed: RNG seed, chosen distinct from the source cloud's.
+
+    Returns:
+        An ``(n_points, 3)`` array of points on the unit sphere.
+    """
+    rng = np.random.default_rng(seed)
+    xyz = rng.normal(size=(n_points, 3))
+    return xyz / np.linalg.norm(xyz, axis=1)[:, np.newaxis]
```
dsroberts: Using PRNG in a test feels a bit sketchy unless you can be sure you're getting the same distribution each time. How are you going to be able to replicate a failure if the input changes every time?

dsroberts (commented):
A sensible change, though I do have some questions/comments on the tests.

tests/unit/test_gplates_interpolation_config.py
dsroberts: OK, I've re-read the comments and done my own testing I think what you're doing here makes sense, but I'm not sure its necessary. This test doesn't care what is supposed to be happening in the interpolation, it just needs to know whether cache entries were created or not, and whether the cache was hit when it should have been. I can see that you're using the field outputs as a proxy for whether the cache is storing the correct result, but I think it would make more sense to inspect the cache directly rather than assuming a numerical difference in a field implies the cache contains different fields. Direct `is` and `is not` on the objects should be sufficient.

sghelichkhani:
Thanks Daniel :-), a real improvement and Mock is my new favorite thing. I also rewrote the docstrings, since a few still described the old field-comparison test and the _interpolator attribute that isn't on this branch. Both commits are pushed.

dsroberts (approved):
I think this is in good shape. I don't fully understand how stacks are supposed to work, so I guess this gets approved but not merged yet?

sghelichkhani:
Yes! That's what I would say, let's go ahead and review the rest and once everything is green we merge all of them at once with main!

dsroberts:
Cool, let me know when #535 is ready for review and I'll get on it.

PR #535, opened 2026-08-05
Port the gadopt gplates source layer onto gtrack 0.7.0

sghelichkhani:
This PR is the middle layer of #534 ← #535 ← #536. It supersedes the source-layer changes in #517 and #518.

## Summary

This PR moves plate reconstruction and point-cloud generation from G-ADOPT to gtrack 0.7.0. G-ADOPT retains Firedrake fields, MPI coordination, interpolation, and connector factories.

`PointCloudSource` adapts gtrack producers to the collective `prepare(age)` interface. Rank zero validates each age, and every rank receives the result before the collective source call.

The connector checks the `requires` set for each output against the `provides` set for the source. `SphericalKNNInterpolator` returns `outside_source_range` and `neighbor_coverage`.

Outputs use explicit `thickness`, `masked_thickness`, `membership`, and `lateral_weight` channels. `IndicatorOutput` separates the radial transition, base depth, and lateral weight.

Existing quintic outputs remain presets. The polygon geotherm uses mantle temperature outside the selected region. The `gplates` extra requires `gtrack>=0.7.0`.

## Verification

The focused GPlates suite passes with the Müller 2022 reconstruction data: 159 passed and 4 skipped. This includes the reconstruction download and velocity regression tests.

GitHub lint and smoke tests pass.


.gitignore
```diff
@@ -8,4 +8,5 @@
 *.cereal
 *.ipynb
 __pycache__/
-.doit.db.db
\ No newline at end of file
+.doit.db.db
+NOTES/
```
sghelichkhani:
```suggestion
```

gadopt/gplates/connectors.py
```diff
@@ -1,129 +1,38 @@
...
 # ---------------------------------------------------------------------------
 
 class ScalarFieldConnector:
-    """Composition of a Source and an OutputStrategy into a single time-
-    varying scalar field on a mesh.
+    """Evaluate source channels as one time-dependent target field.
 
-    Construction validates ``output.requires <= source.provides``; this
-    catches the obvious mis-pairings (e.g. PolygonSource + GeothermERFOutput,
-    which needs ``"age"``) at the point of wiring rather than at the first
```
sghelichkhani, line 107: @dsroberts I know the rewrite of the comments is making this even more confusing. So might make sense for us to sit down togehter and we go through things together if things are confusing. 
But basically I tried to make things more concise and not so verbose (confuisng). Will have another roud of review on the commenting todya.

gadopt/gplates/__init__.py
dsroberts, line 1: I think we can simplify this file by correct use of `__all__` in the modules being imported here. If every module has `__all__` specified, this file can be reduced to 
```
from .connectors import *
from .interpolation import *
from .gplates import *
from .outputs import *
from .sources import *
from .factories import *
```

gadopt/gplates/connectors.py
```diff
@@ -214,30 +159,24 @@ def get_indicator(
 
     # Cache
     def _check_cache(self, age: float, target_coords: np.ndarray) -> bool:
```
dsroberts, line 161: Docstrings

gadopt/gplates/connectors.py
```diff
@@ -214,30 +159,24 @@ def get_indicator(
...
+        # A released coordinate array gives a dead weak reference and a miss.
         if self._cached_coords_ref() is not target_coords:
             return False
         log(f"{type(self).__name__}: age {age:.2f} Ma unchanged "
             f"(within delta_t={self.delta_t}); reusing cached result.",
             level=DEBUG)
         return True
 
-    # Update the cache
     def _update_cache(
```
dsroberts, line 177: Docstrings

gadopt/gplates/factories.py
```diff
@@ -0,0 +1,435 @@
...
+``PolygonConnectorFactory`` are presets that fix those classes for the global
+and bounded cases and narrow the signatures to the arguments that apply.
+
+Every slot is write-once, and the connectors are cached properties. Reassigning
+a source after a connector exists would leave the two disagreeing about which
+points they read, so the factory refuses instead.
+"""
+
+from mpi4py import MPI
+from typing import Callable, TYPE_CHECKING
```
dsroberts, line 22:
```suggestion
from typing import TYPE_CHECKING
from collections.abc import Callable
```
typing.Callable has been [deprecated since Python 3.9](https://docs.python.org/3/library/typing.html#typing.Callable)

gadopt/gplates/factories.py
```diff
@@ -0,0 +1,435 @@
...
+
+        Raises:
+            RuntimeError: If this factory already has a geotherm output.
+            TypeError: If no geotherm output class was configured.
+        """
+        if self._geotherm_output is not None:
+            raise RuntimeError("This factory already has a geotherm output.")
+        if self._geotherm_output_class is None:
+            raise TypeError("The geotherm output class is not configured.")
+        self._geotherm_output = self._geotherm_output_class(**output_kwargs)
```
dsroberts, line 184:
```suggestion
        if self._geotherm_output_class is None:
            raise TypeError("The geotherm output class is not configured.")
        self.geotherm_output = self._geotherm_output_class(**output_kwargs)
```
Use the setter defined 20 lines up.

gadopt/gplates/factories.py
```diff
@@ -0,0 +1,435 @@
...
+
+        Raises:
+            RuntimeError: If this factory already has an indicator output.
+            TypeError: If no indicator output class was configured.
+        """
+        if self._output is not None:
+            raise RuntimeError("This factory already has an indicator output.")
+        if self._output_class is None:
+            raise TypeError("The indicator output class is not configured.")
+        self._output = self._output_class(**output_kwargs)
```
dsroberts, line 157:
```suggestion
        if self._output_class is None:
            raise TypeError("The indicator output class is not configured.")
        self.output = self._output_class(**output_kwargs)
```
Use the setter defined 20 lines up.

gadopt/gplates/factories.py
```diff
@@ -0,0 +1,435 @@
...
+
+        Raises:
+            RuntimeError: If this factory already has a source.
+            TypeError: If no source class was configured.
+        """
+        if self._source is not None:
+            raise RuntimeError("This factory already has a source.")
+        if self._source_class is None:
+            raise TypeError("The source class is not configured.")
+        self._source = self._source_class(*source_args, **source_kwargs)
```
dsroberts, line 130:
```suggestion
        if self._source_class is None:
            raise TypeError("The source class is not configured.")
        self.source = self._source_class(*source_args, **source_kwargs)
```
Use the setter defined 20 lines up.

gadopt/gplates/factories.py
```diff
@@ -0,0 +1,435 @@
...
+        """Return the output strategy that creates the geotherm field."""
+        return self._geotherm_output
+
+    @geotherm_output.setter
+    def geotherm_output(self, geotherm_output: OutputStrategy | None):
+        if self._geotherm_output is not None:
+            raise RuntimeError("This factory already has a geotherm output.")
+        self._geotherm_output = geotherm_output
+
+    def create_geotherm(self, **output_kwargs):
```
dsroberts, line 170: Potentially confusing name? `geotherm()` on line 220 creates a geotherm. This creates a geotherm output.

gadopt/gplates/interpolation.py
```diff
@@ -0,0 +1,252 @@
...
+where the interpolation is untrustworthy. ``outside_source_range`` for target
+nodes whose nearest source point is further away than
+``InterpolationConfig.max_source_separation_rad``, which is how a bounded
+source (a continental polygon, say) tells an output to use the "fallback" value.
+``neighbor_coverage`` decides the fraction of the
+queried neighbors that fell inside that range, which fades smoothly to zero
+across the edge of a source cloud.
+"""
+
+from __future__ import annotations
```
dsroberts, line 23: Unnecessary and deprecated.
```suggestion
```

gadopt/gplates/interpolation.py
```diff
@@ -0,0 +1,252 @@
...
+
+from dataclasses import dataclass
+
+import numpy as np
+from scipy.spatial import cKDTree
+
+
```
dsroberts, line 30:
```suggestion
__all__ = [
    "InterpolationConfig",
    "SphericalKNNInterpolator"
]
```

gadopt/gplates/interpolation.py
```diff
@@ -0,0 +1,252 @@
...
+            )
+        if self.gaussian_width_rad > np.pi:
+            raise ValueError(
+                "gaussian_width_rad must be at most pi radians, "
+                f"got {self.gaussian_width_rad}"
+            )
+        # Forcing an unhashable field (a list, a numpy array) to fail.
+        # This is to avoid silent breaking of the geometry cache
+        # the first time this config is used as a key.
+        hash(self)
```
dsroberts, line 110:
```suggestion
```
This is, on the surface, a sensible thing to check, however, for this to fail, the user would need to have ignored the type hinting then somehow passed a non-hashable type that has valid `__lt__`, `__le__` and `__gt__` operators with floats and ints. Input validation is important, but I think this class already has enough.

gadopt/gplates/connectors.py
```diff
@@ -257,92 +196,18 @@ def _compute(
...
+        geometry = self.source.get_or_build_geometry(
+            key, lambda: self._interpolator.geometry(source_xyz, target_coords)
         )
 
-        prop_keys = sorted(self.output.requires)
+        channel_keys = sorted(self.output.requires)
         interpolated = {
-            k: self._interp_gather(bundle, sources_dict[k]) for k in prop_keys
+            k: SphericalKNNInterpolator.gather(geometry, sources_dict[k])
+            for k in channel_keys
```
dsroberts, line 209: I'm confused as to why this has been changed to a `staticmethod`. What was wrong with how it was implemented before?

gadopt/gplates/interpolation.py
```diff
@@ -0,0 +1,252 @@
...
+        return {
+            "k1": False,
+            "idx": idx,
+            "outside_source_range": outside_source_range,
+            "neighbor_coverage": neighbor_coverage,
+            "exact_match": exact_match,
+            "weights": weights,
+        }
+
+    @staticmethod
```
dsroberts, line 226: Why is this static now? If it is generic and can be used by any interpolator it should probably be a free function in this module.

gadopt/gplates/outputs.py
```diff
@@ -1,40 +1,52 @@
...
-set; the connector validates this against the source's ``provides`` set at
-construction time so, e.g. a polygon source paired with an erf geotherm fails loudly
-rather than silently dropping the missing key.
+Each output declares the source channels it reads in ``requires``, which the
+connector checks against the source's ``provides`` when the two are wired
+together. That turns a mismatched pairing into an error at construction time
+rather than a ``KeyError`` deep inside ``compute`` on the first timestep.
 """
 
 from __future__ import annotations
```
dsroberts, line 22:
```suggestion
```
Unnecessary and deprecated

gadopt/gplates/outputs.py
```diff
@@ -1,40 +1,52 @@
...
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
-from typing import ClassVar
+from typing import Callable, Protocol
```
dsroberts, line 26:
```suggestion
from typing import Protocol
from collections.abc import Callable
```
typing.Callable has been [deprecated since Python 3.9](https://docs.python.org/3/library/typing.html#typing.Callable)

gadopt/gplates/outputs.py
```diff
@@ -1,40 +1,52 @@
...
 
 import numpy as np
+from scipy.special import erf
```
dsroberts, line 31:
```suggestion
__all__ = [
    "InterpolatedBaseDepth",
    "MembershipCorrectedBaseDepth",
    "FixedBaseDepth",
    "HalfSpaceCoolingGeotherm",
    "LinearGeotherm",
    "LayerIndicator",
    "LateralWeight",
    "MembershipField",
    "BoundedLinearGeotherm",
    "MembershipLateralWeight",
    "MeshConfig",
    "UniformLateralWeight",
    "OutputStrategy",
    "BoundedLayerIndicator",
    "GlobalLayerIndicator",
    "RadialQuinticTransition",
    "SourceLateralWeight",
    "MappedMembershipWeight",
    "continental_linear",
    "ocean_erf_normalized",
    "radial_quintic_step",
]
```

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+    Returns:
+        The normalised temperature, clipped to the interval from zero to one.
+        Nodes with a non-positive base depth return zero.
+    """
+    depth_m = np.asarray(depth_m, dtype=float)
+    z_lab_m = np.asarray(z_lab_m, dtype=float)
+    result = np.zeros_like(depth_m)
+    valid = z_lab_m > 0
+    result[valid] = depth_m[valid] / z_lab_m[valid]
+    return np.clip(result, 0.0, 1.0)
```
dsroberts, line 126:
```suggestion
    return np.clip(np.where(z_lab_m > 0, depth_m / z_lab_m, 0.0), 0.0, 1.0)
```
Can do this without creating named temporary variables, also I just think this looks a little neater.

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+        Args:
+            interpolated: Arrays keyed by channel name.
+            outside_source_range: Boolean mask, True where no source point is
+                in range.
+
+        Returns:
+            The weight at each target node, or a single float that applies
+            everywhere.
+        """
+        ...
```
dsroberts, line 228: I don't understand what this is for. Its a do-nothing class that is never subclassed. I don't believe it fits the definition of a protocol from here: https://docs.python.org/3/library/typing.html#typing.Protocol

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+
+    def __init__(self, base_transition_width_km: float = 10.0):
+        if base_transition_width_km <= 0:
+            raise ValueError(
+                "base_transition_width_km must be positive, "
+                f"got {base_transition_width_km}"
+            )
+        self.base_transition_width_km = base_transition_width_km
+
+    def step(self, r_target, base_r, mesh):
```
dsroberts, line 271: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+                in range.
+
+        Returns:
+            The weight at each target node, or a single float that applies
+            everywhere.
+        """
+        ...
+
+
+def _clip_membership(interpolated, outside_source_range):
```
dsroberts, line 231: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+    depth_m = np.asarray(depth_m, dtype=float)
+    z_lab_m = np.asarray(z_lab_m, dtype=float)
+    result = np.zeros_like(depth_m)
+    valid = z_lab_m > 0
+    result[valid] = depth_m[valid] / z_lab_m[valid]
+    return np.clip(result, 0.0, 1.0)
+
+
+# Shared radial primitive (used by every indicator output)
+def radial_quintic_step(r_target, base_r, width_nondim):
```
dsroberts, line 130: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+    erf_zlab = erf(z_lab_m[valid] / a[valid])
+    safe = erf_zlab > 1e-10
+    result[valid] = np.where(
+        safe, erf_z / np.maximum(erf_zlab, 1e-10),
+        depth_m[valid] / z_lab_m[valid],
+    )
+    return np.clip(result, 0.0, 1.0)
+
+
+def continental_linear(depth_m, z_lab_m):
```
dsroberts, line 106: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
             raise ValueError(f"depth_scale must be positive, got {self.depth_scale}")
 
 
+# Geotherm functions (used by HalfSpaceCoolingGeotherm / LinearGeotherm)
+def ocean_erf_normalized(depth_m, z_lab_m, age_myr, thermal_diffusivity_m2_per_s):
```
dsroberts, line 63: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+    requires = frozenset()
+
+    def __init__(self, fixed_base_depth_km: float):
+        if fixed_base_depth_km <= 0:
+            raise ValueError(
+                f"fixed_base_depth_km must be positive, got {fixed_base_depth_km}"
+            )
+        self.fixed_base_depth_km = fixed_base_depth_km
+
+    def base_r(self, interpolated, outside_source_range, mesh):
```
dsroberts, line 299: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+
+    def __init__(self, fallback_thickness_km: float = 0.0):
+        if fallback_thickness_km < 0:
+            raise ValueError(
+                f"fallback_thickness_km must be non-negative, "
+                f"got {fallback_thickness_km}"
+            )
+        self.fallback_thickness_km = fallback_thickness_km
+
+    def base_r(self, interpolated, outside_source_range, mesh):
```
dsroberts, line 325: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+    margins. It carries ``masked_thickness = membership * thickness`` instead,
+    and dividing the interpolated product by the interpolated membership
+    recovers the physical thickness inside the region. Below
+    ``MEMBERSHIP_FLOOR`` the division is meaningless, so those nodes get a base
+    depth of zero and the layer vanishes there.
+    """
+
+    requires = frozenset({"masked_thickness", "membership"})
+
+    def base_r(self, interpolated, outside_source_range, mesh):
```
dsroberts, line 345: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+class UniformLateralWeight:
+    """Return a lateral weight of one at all target nodes.
+
+    This is the strategy for a layer that covers the whole sphere and varies
+    only in thickness.
+    """
+
+    requires = frozenset()
+
+    def weight(self, interpolated, outside_source_range):
```
dsroberts, line 367: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+class MembershipLateralWeight:
+    """Use the interpolated membership channel as the lateral weight.
+
+    The layer fades out across the edge of the bounded region in step with the
+    interpolated membership, which avoids a hard polygon boundary in the field.
+    """
+
+    requires = frozenset({"membership"})
+
+    def weight(self, interpolated, outside_source_range):
```
dsroberts, line 380: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+    sharpen a margin or to hold full strength over most of a region.
+
+    Args:
+        mapping: Callable applied to membership between zero and one. Its
+            result is clipped back to the same interval.
+    """
+
+    requires = frozenset({"membership"})
+
+    def __init__(self, mapping):
```
dsroberts, line 397: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+        self.mapping = mapping
+
+    def weight(self, interpolated, outside_source_range):
```
dsroberts, line 400: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+
+    A node outside the source range falls back to a weight of one rather than
+    zero, because here a missing source point means the source has nothing to
+    say about that node, not that the layer is absent. Zeroing it would let
+    gaps in the source cloud punch holes in the field.
+    """
+
+    requires = frozenset({"lateral_weight"})
+
+    def weight(self, interpolated, outside_source_range):
```
dsroberts, line 416: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+
+    Args:
+        radial_transition: Strategy with a ``step(r_target, base_r, mesh)``
+            method.
+        base_depth: Strategy with a
+            ``base_r(interpolated, outside_source_range, mesh)`` method.
+        lateral_weight: Strategy following the ``LateralWeight`` protocol.
+    """
+
+    def __init__(
```
dsroberts, line 440: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+        self.radial_transition = radial_transition
+        self.base_depth = base_depth
+        self.lateral_weight = lateral_weight
+        self.requires = (
+            radial_transition.requires
+            | base_depth.requires
+            | lateral_weight.requires
+        )
+
+    def compute(self, interpolated, r_target, outside_source_range, mesh):
```
dsroberts, line 455: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+        if fallback_age_myr <= 0:
+            raise ValueError(
+                f"fallback_age_myr must be positive, got {fallback_age_myr}"
+            )
+        self.thermal_diffusivity_m2_per_s = thermal_diffusivity_m2_per_s
+        self.fallback_thickness_km = fallback_thickness_km
+        self.fallback_age_myr = fallback_age_myr
+        self._geotherm = geotherm or ocean_erf_normalized
+
+    def compute(self, interpolated, r_target, outside_source_range, mesh):
```
dsroberts, line 630: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+
+    requires = frozenset({"thickness"})
+
+    def __init__(
+        self,
+        geotherm: Callable | None = None,
+    ):
+        self._geotherm = geotherm or continental_linear
+
+    def compute(self, interpolated, r_target, outside_source_range, mesh):
```
dsroberts, line 666: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+
+    requires = frozenset({"masked_thickness", "membership"})
+
+    def __init__(
+        self,
+        geotherm: Callable | None = None,
+    ):
+        self._geotherm = geotherm or continental_linear
+
+    def compute(self, interpolated, r_target, outside_source_range, mesh):
```
dsroberts, line 695: Missing type annotations

gadopt/gplates/outputs.py
```diff
@@ -47,32 +59,675 @@ def __post_init__(self):
...
+    """Return the membership channel as a field between zero and one.
+
+    Useful on its own for diagnostics and for masking other fields: it shows
+    where a bounded source considers itself present, with no radial dependence
+    at all. Nodes outside the source range are zero.
+    """
+
+    requires = frozenset({"membership"})
+
+    def compute(self, interpolated, r_target, outside_source_range, mesh):
```
dsroberts, line 730: Missing type annotations

gadopt/gplates/sources.py
```diff
@@ -1,112 +1,130 @@
...
+geometry whenever the points move to a new age.
+
+``prepare`` is collective across ``comm``. The producer runs on rank zero only
+and the result is broadcast, which keeps one copy of the gtrack state in the
+job and keeps every rank in step. Age validation follows the same rule for the
+same reason: a rank-zero exception on its own leaves the other ranks blocked in
+the next collective call, so failures are broadcast before they are raised.
 """
 
 from __future__ import annotations
```
dsroberts, line 23:
```suggestion
```
Still unnecessary and deprecated

gadopt/gplates/sources.py
```diff
@@ -1,112 +1,130 @@
...
 
 import numpy as np
+import numpy.typing as npt
 from mpi4py import MPI
 
+from gtrack import PointCloud
+from gtrack.age_sources import AgeCloudSource
+
 if TYPE_CHECKING:
     from .gplates import pyGplatesConnector
```
dsroberts, line 39:
```suggestion
__all__ = [
    "CloudDataType",
    "PointCloudSource",
    "Source",
]
```

gadopt/gplates/sources.py
```diff
@@ -117,30 +135,35 @@ def prepare(self, age: float) -> dict[str, np.ndarray]:
...
         self._cached_dict = sources
-        # The source cloud just changed, so any cached interpolation geometry
-        # (cKDTree indices/weights over the previous cloud) is stale. Clear it
-        # in lockstep with _cached_dict so the two caches never disagree on age.
+        # New source points invalidate geometry that refers to the previous age.
         if self._interp_geometry_cache is not None:
             self._interp_geometry_cache.clear()
         return sources
 
     def get_or_build_geometry(self, key, build_fn):
```
dsroberts, line 143: Missing type annotation

gadopt/gplates/sources.py
```diff
@@ -175,3 +204,88 @@ def validate_age(self, age: float) -> None:
...
+        # exception alone can leave the other ranks blocked in ``prepare``.
+        message = None
+        if self._is_root:
+            try:
+                self.producer.validate_age(age)
+            except Exception as exc:
+                message = str(exc) or type(exc).__name__
+        message = self.comm.bcast(message, root=0)
+        if message is not None:
+            raise ValueError(message)
```
dsroberts, line 283: Not sure about this strategy. Are we expecting this error to be caught at some point? That's about the only situation that comes to mind that would cause a hang somewhere. I think trying to handle this here is opening a bit of a can of worms about how MPI and python exceptions should work together. I think just let the exception go uncaught and if someone down the line tries to catch it and their code hangs that's on them.

tests/unit/test_chord_threshold.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+# ---------------------------------------------------------------------------
+# The conversion itself
+# ---------------------------------------------------------------------------
+
+class TestAngleToChord:
+    def test_matches_the_closed_form(self):
+        assert _angle_to_chord(0.5) == 2.0 * np.sin(0.25)
+        assert _angle_to_chord(0.02) == 2.0 * np.sin(0.01)
```
dsroberts, line 149: Floating point comparisons like this can be dangerous. Sometimes `0.02 / 2.0 == 0.010000000000000001`. You've acknowledged this 8 lines below in `TestGaussianWidthConversion`.

tests/unit/test_chord_threshold.py
```diff
@@ -0,0 +1,284 @@
...
+        # The inclusive limit lets callers disable the source-range test.
+        cfg = InterpolationConfig(max_source_separation_rad=np.pi)
+        assert cfg.max_source_separation_rad == np.pi
+
+    def test_above_pi_is_rejected(self):
+        # Beyond pi, a larger angle gives a smaller chord threshold.
+        with pytest.raises(ValueError, match="at most pi"):
+            InterpolationConfig(max_source_separation_rad=np.pi + 1e-9)
+        with pytest.raises(ValueError, match="at most pi"):
+            InterpolationConfig(max_source_separation_rad=4.0)
```
dsroberts, line 200: This is the same assert twice?

tests/unit/test_chord_threshold.py
```diff
@@ -0,0 +1,284 @@
...
+        ],
+        ids=["single-neighbour", "multiple-neighbours"],
+    )
+    def test_threshold_and_mask_contract(self, source_xyz):
+        # The cases exercise the single-neighbour and multi-neighbour branches.
+        thetas = [0.497, 0.503, 0.51]
+        mask = outside_source_range_mask(source_xyz, thetas, 0.5)
+        np.testing.assert_array_equal(mask, [False, True, True])
+        assert mask.dtype == np.bool_
+        assert mask.shape == (len(thetas),)
```
dsroberts, line 228: Doesn't `np.testing.assert_array_equal(mask, [False, True, True])` imply ` mask.dtype == np.bool_` and `mask.shape == (len(thetas),)`?

tests/unit/test_docstring_references.py
dsroberts, line 1: No, just no. Wildly out of scope and yet too narrow in scope to be useful.

tests/unit/test_connectors.py
```diff
@@ -0,0 +1,934 @@
...
+
+# ---------------------------------------------------------------------------
+# Paths and constants
+# ---------------------------------------------------------------------------
+
+REPO_ROOT = Path(__file__).resolve().parents[2]
+GPLATES_GLOBAL = REPO_ROOT / "demos/mantle_convection/gplates_global"
+GPLATES_FIELDS = REPO_ROOT / "demos/mantle_convection/gplates_fields"
+CONTINENTAL_DATA = GPLATES_FIELDS / "continental_lithospheric_thickness_mesh.h5"
+CRATON_SHAPEFILE = GPLATES_FIELDS / "Craton_Boundaries_Inferred.shp"
```
dsroberts, line 73: Hello I'm new to G-ADOPT and I want to make sure its working on my system. Where can I get these files from?

tests/unit/test_connectors.py
```diff
@@ -0,0 +1,934 @@
...
+    POLYGON_BACKGROUND_POINT_COUNT,
+    POLYGON_SCALAR_INPUT_POINT_COUNT,
+    TEST_AGES,
+    REGRESSION_LAYER_HEIGHTS,
+)
+
+
+def _require_data():
+    if not (GPLATES_GLOBAL / "Muller_etal_2022_SE_1Ga_Opt_PlateMotionModel_v1.2").exists():
+        pytest.skip("Muller 2022 SE reconstruction not downloaded; run `make data`.")
```
dsroberts, line 93: `make data` does not work (nor should it).

tests/unit/test_connectors.py
```diff
@@ -0,0 +1,934 @@
...
+        )
+        conn = ScalarFieldConnector(
+            _DataSource(), GlobalLayerIndicator(), gc_collect_frequency=frequency
+        )
+        target = _target_coords()
+        # Distinct ages (spaced > delta_t=1.0, all <= oldest_age=100) so every
+        # call is a cache miss and runs _compute (where the gc counter lives).
+        for age in range(90, 90 - 10 * n_calls, -10):
+            conn.get_indicator(target, conn.source.age2ndtime(float(age)))
+        return calls["n"]
```
dsroberts, line 234:
```suggestion
        mock_gc = Mock()
        monkeypatch.setattr("gadopt.gplates.connectors.gc.collect", mock_gc)
        conn = ScalarFieldConnector(
            _DataSource(), GlobalLayerIndicator(), gc_collect_frequency=frequency
        )
        target = _target_coords()
        # Distinct ages (spaced > delta_t=1.0, all <= oldest_age=100) so every
        # call is a cache miss and runs _compute (where the gc counter lives).
        for age in range(90, 90 - 10 * n_calls, -10):
            conn.get_indicator(target, conn.source.age2ndtime(float(age)))
        return mock_gc.call_count
```
Please stop re-inventing this particular wheel

tests/unit/test_connectors.py
```diff
@@ -0,0 +1,934 @@
...
+        directions = _fibonacci_directions(n_points)
+        inside = np.arccos(np.clip(directions[:, 2], -1.0, 1.0)) < self.CAP_RADIUS_RAD
+        self._fixed = {
+            "xyz": 6.371e6 * directions,
+            "membership": inside.astype(float),
+            "masked_thickness": self.DEPTH_KM * inside.astype(float),
+        }
+
+
+class _CountingCKDTree:
```
dsroberts, line 375: I don't like that we have yet another implementation of `Mock()`. We don't need `tree.query` to actually do anything. You can mock the return value of a specific class member function with 
```
mock_ckdtree = Mock()
mock_ckdtree.return_value.query.return_value = (np.ones((15,2),dtype=np.int32),np.ones((15,2),dtype=np.int32))
```
This will satisfy the return requirements of `tree.query` in `SphericalKNNInterpolator.geometry` without having to go through the actual interpolation, and therefore do away with the entire `_CountingCKDTree` class.

tests/unit/test_connectors.py
```diff
@@ -0,0 +1,934 @@
...
+@pytest.fixture(scope="module")
+def Q(regression_mesh):
+    return fd.FunctionSpace(regression_mesh, "CG", 1)
+
+
+# ---------------------------------------------------------------------------
+# Regression: four factory pairings + GplatesScalarFunction wrapper
+# ---------------------------------------------------------------------------
+
+def _reduced_quantities(values, mesh):
```
dsroberts, line 730: What is this function for? It takes some raw data, turns it into a firedrake function and calculates some things. But its used on things that are already firedrake functions? Why the extra steps?

dsroberts (commented):
Structurally I think this is all good. This is a pretty sizeable PR, I still need to look at `test_gplates.py`, `test_gplates_interpolation_config.py`, `test_outputs.py` and `test_sources.py`.

gadopt/gplates/connectors.py
```diff
@@ -214,30 +159,49 @@ def get_indicator(
 
     # Cache
```
sghelichkhani:
```suggestion
```

gadopt/gplates/sources.py
sghelichkhani: So just that I know, you want me to remove the tickery with message? So that raise is raise?

tests/unit/test_chord_threshold.py
sghelichkhani: Two different values: the first is just over pi (pi + 1e-9), the second is well past it (4.0). The first checks the boundary is strict, the second that it is not only a rounding edge case. Happy to fold them into a parametrize if you prefer.

tests/unit/test_connectors.py
sghelichkhani: Good point, this was skipping in CI all along. The test workflow now fetches the thickness h5 and the craton shapefile from data.gadopt.org, and a missing file is an assertion failure whose message gives the download URL, same as test_gplates.py.

tests/unit/test_chord_threshold.py
dsroberts: I see that, but having both suggests you may not trust Python's ability to handle small delta floating point operations. From a testing standpoint, more is more, it doesn't really matter how much. You could put `1e10` there and its still the same test.

gadopt/gplates/sources.py
dsroberts: Yeah, I think this is one of the uncomfortable collision points between Python and MPI that mpi4py gives us. There is a fundamental disagreement between how they each handle error states. MPI prefers to fail fast, whereas Python gives the user a chance to recover the situation. For the rest of G-ADOPT I think we err on the side of 'fail fast', just throw the exception and expect the runtime to deal with it. For exceptions raised on a single rank, the Python process will end, MPI will pick that up and kill the rest of the processes. There is kind of an implicit understanding we don't intend for exceptions to be caught, and if you do, you do so at your own risk because it will most likely cause deadlocks. This solves the potential deadlock issue, but why? What are the circumstances a user script would want to recover from this error? Is there some kind of fallback mechanism a user can employ? I can't think of a reason a job should continue if the age is invalid, so I think its fair to just let the exception go uncaught like we do for all but 3 other exceptions in the entire G-ADOPT codebase.

tests/unit/test_connectors.py
dsroberts: Not really what I was going for. Not for this PR but I really think we need to treat Gadi (i.e. running offline) as the exception rather than the rule. `gplatesfiles.py` should be extended to include the continental data and cratons, each of the structs in there should contain a URL that `ensure_reconstruction()` and its (not yet written) equivalents for continents and cratons should attempt to download if the files aren't already present. Only if that download step fails should the user be prompted to manually download the required files. I will gladly take this on myself.

tests/unit/test_chord_threshold.py
sghelichkhani: Fair enough, dropped the 4.0 one.

gadopt/gplates/sources.py
sghelichkhani: Agreed, and removed.

sghelichkhani:
Addressed

tests/unit/test_connectors.py
sghelichkhani: All good! Happy to discuss this later. I still think the only time we use these files other than testing is on gadi, and offline. But happy to be overruled.

sghelichkhani:
Ready for another look @dsroberts

gadopt/gplates/interpolation.py
```diff
@@ -0,0 +1,251 @@
...
+
+        Returns:
+            A dictionary with the neighbor indices ``idx``, the masks
+            ``outside_source_range`` and ``neighbor_coverage``, and a ``k1``
+            flag. When ``k1`` is False the geometry also carries the normalised
+            ``weights`` and the ``exact_match`` mask.
+            Should be treated as read-only.
+        """
+        cfg = self.config
+        chord_threshold = _angle_to_chord(cfg.max_source_separation_rad)
```
dsroberts:
```suggestion
        chord_threshold = _angle_to_chord(self.config.max_source_separation_rad)
```
Unnecessary temporary variable.

gadopt/gplates/interpolation.py
```diff
@@ -0,0 +1,251 @@
...
+
+        r_source = np.linalg.norm(source_xyz, axis=1)
+        unit_source = source_xyz / np.maximum(r_source[:, np.newaxis], _EPSILON)
+
+        r_target = np.linalg.norm(target_coords, axis=1)
+        unit_target = target_coords / np.maximum(r_target[:, np.newaxis], _EPSILON)
+
+        tree = cKDTree(unit_source)
+        k = min(cfg.neighbor_count, len(source_xyz))
```
dsroberts:
```suggestion
        k = min(self.config.neighbor_count, len(source_xyz))
```

gadopt/gplates/interpolation.py
```diff
@@ -0,0 +1,251 @@
...
+                "neighbor_coverage": neighbor_coverage,
+            }
+
+        exact_match = source_chord_distances[:, 0] < _EPSILON
+        outside_source_range = source_chord_distances[:, 0] > chord_threshold
+        # Neighbor coverage is zero when the nearest source point is out of range.
+        neighbor_coverage = np.mean(source_chord_distances <= chord_threshold, axis=1)
+
+        if cfg.kernel == "gaussian":
+            gaussian_chord_width = 2.0 * np.sin(cfg.gaussian_width_rad / 2.0)
```
dsroberts:
```suggestion
        if self.config.kernel == "gaussian":
            gaussian_chord_width = 2.0 * np.sin(self.config.gaussian_width_rad / 2.0)
```

gadopt/gplates/outputs.py
```diff
@@ -47,32 +83,744 @@ def __post_init__(self):
...
+    result = np.zeros_like(depth_m)
+    valid = z_lab_m > 0
+    erf_z = erf(depth_m[valid] / a[valid])
+    erf_zlab = erf(z_lab_m[valid] / a[valid])
+    safe = erf_zlab > 1e-10
+    result[valid] = np.where(
+        safe, erf_z / np.maximum(erf_zlab, 1e-10),
+        depth_m[valid] / z_lab_m[valid],
+    )
+    return np.clip(result, 0.0, 1.0)
```
dsroberts, line 129:
```suggestion
    result = np.zeros_like(depth_m)
    np.divide(depth_m, z_lab_m, result, where=z_lab_m > 0)
    erf_zlab = erf(z_lab_m / a)
    np.divide(erf(depth_m / a), erf_zlab, result, where=erf_zlab > 1e-10)
    return np.clip(result, 0.0, 1.0)
```
A bit of a simplification that reduces use of temporary arrays.

gadopt/gplates/outputs.py
```diff
@@ -47,32 +83,744 @@ def __post_init__(self):
...
+            range.
+
+    Returns:
+        Membership between zero and one, zero outside the source range.
+    """
+    m = np.clip(interpolated["membership"], 0.0, 1.0)
+    return np.where(outside_source_range, 0.0, m)
+
+
+class RadialQuinticTransition:
```
dsroberts: I think you should do what you've done for the lateral weights and create an abstract base class for the radial transition (even though there is only one of those for now) and for the base depths. It makes it clear to anyone working on this in future what is expected from these classes.

dsroberts (commented):
Overall looking good. A few minor things, but I think the `gplates` layer is looking just about ready. I'll take a look at the tests later this afternoon.

gadopt/gplates/outputs.py
```diff
@@ -47,32 +83,744 @@ def __post_init__(self):
...
+        depth_m: Depth below the surface, in metres.
+        z_lab_m: Depth of the lithosphere base, in metres.
+
+    Returns:
+        The normalised temperature, clipped to the interval from zero to one.
+        Nodes with a non-positive base depth return zero.
+    """
+    depth_m = np.asarray(depth_m, dtype=float)
+    z_lab_m = np.asarray(z_lab_m, dtype=float)
+    return np.clip(np.where(z_lab_m > 0, depth_m / z_lab_m, 0.0), 0.0, 1.0)
```
dsroberts: Oops, I led you astray here. Annoyingly this produces a runtime warning if any element `z_lab_m` is 0. Here is another method that masks first, then performs the the division.
```suggestion
    result = np.zeros_like(depth_m)
    return np.clip(np.divide(depth_m, z_lab_m, where=z_lab_m > 0, out=result), 0.0, 1.0)
```

tests/unit/data/generate_expected_connectors.py
```diff
@@ -0,0 +1,198 @@
...
+        layers=len(REGRESSION_LAYER_HEIGHTS),
+        layer_height=np.array(REGRESSION_LAYER_HEIGHTS, dtype=float),
+        extrusion_type="radial",
+    )
+    mesh.cartesian = False
+    return mesh
+
+
+def reduced(values):
+    f_space = fd.FunctionSpace(_mesh, "CG", 1)
```
dsroberts, line 82:
```suggestion
def reduced(f_space, values):
```

tests/unit/data/generate_expected_connectors.py
```diff
@@ -0,0 +1,198 @@
...
+        name: GplatesScalarFunction(Q, indicator_connector=c, name=name)
+        for name, c in connectors_by_name.items()
+    }
+    out = {name: {} for name in connectors_by_name}
+    sample_connector = next(iter(connectors_by_name.values()))
+    for age in ages:
+        ndtime = sample_connector.source.age2ndtime(float(age))
+        for name in connectors_by_name:
+            sfs[name].update_plate_reconstruction(ndtime)
+            out[name][age] = reduced(sfs[name].dat.data_ro_with_halos.copy())
```
dsroberts, line 114:
```suggestion
            out[name][age] = reduced(Q, sfs[name].dat.data_ro_with_halos.copy())
```

tests/unit/data/generate_expected_connectors.py
```diff
@@ -0,0 +1,198 @@
...
+            print(f"    {name} age={age} Ma  ->  "
+                  f"vol={out[name][age]['volume']:.6e}, "
+                  f"surf={out[name][age]['surface']:.6e}")
+    return out
+
+
+def main():
+    global _mesh
```
dsroberts, line 122:
```suggestion
```

tests/unit/data/generate_expected_connectors.py
dsroberts, line 1: If `f_space` is a CG1 function space on `_mesh` and `Q` is also a CG1 function space on `_mesh` why not just pass `Q` to `reduce`?

tests/unit/test_chord_threshold.py
```diff
@@ -0,0 +1,281 @@
...
+
+class ExplicitCloudSource(Source):
+    """A source whose cloud is given verbatim and never changes with age.
+
+    Args:
+        xyz: ``(n, 3)`` array of source points. Not normalised here; the
+            connector normalises both clouds itself.
+    """
+
+    provides = frozenset({"thickness"})
```
dsroberts, line 58:
```suggestion
```

tests/unit/test_chord_threshold.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def __init__(self, xyz):
+        self.xyz = np.asarray(xyz, dtype=float)
+        self.thickness = np.linspace(50.0, 200.0, len(self.xyz))
+        self.gplates_connector = FakeGplatesConnector()
+        self.comm = MPI.COMM_SELF
+        self._is_root = True
+
```
dsroberts, line 66:
```suggestion
    @property
    def provides(self) -> frozenset[str]:
        return frozenset({"thickness"})

```
It doesn't hurt to correctly override `provides()`

tests/unit/test_chord_threshold.py
```diff
@@ -0,0 +1,281 @@
...
+            np.vstack([NORTH_POLE[None, :], SOUTHERN_DECOYS]),
+        ],
+        ids=["single-neighbour", "multiple-neighbours"],
+    )
+    def test_threshold_and_mask_contract(self, source_xyz):
+        # The cases exercise the single-neighbour and multi-neighbour branches.
+        thetas = [0.497, 0.503, 0.51]
+        mask = outside_source_range_mask(source_xyz, thetas, 0.5)
+        np.testing.assert_array_equal(mask, [False, True, True])
+        assert mask.dtype == np.bool_
```
dsroberts, line 225:
```suggestion
        np.testing.assert_array_equal(mask, [False, True, True], strict=True)
```

dsroberts: Instead of a separate `assert`, `strict=True` will cause `assert_array_equal` to raise if the types are different.

gadopt/gplates/outputs.py
sghelichkhani: Done: RadialTransition and BaseDepth ABCs, with the four concrete classes inheriting from them and LayerIndicator annotated with the two.

tests/unit/test_connectors.py
```diff
@@ -0,0 +1,926 @@
...
+    @pytest.mark.parametrize(
+        "frequency, n_calls, expected_collects",
+        [
+            (3, 9, 3),      # collect every Nth call
+            (None, 5, 0),   # None disables the connector-level collect entirely
+            (1, 4, 4),      # collect on every call
+        ],
+    )
+    def test_collect_cadence(self, monkeypatch, frequency, n_calls, expected_collects):
+        assert self._drive(monkeypatch, frequency, n_calls) == expected_collects
```
dsroberts, line 252:
```suggestion
    @pytest.mark.parametrize(
        "frequency, n_calls, expected_collects",
        [
            (3, 9, 3),      # collect every Nth call
            (None, 5, 0),   # None disables the connector-level collect entirely
            (1, 4, 4),      # collect on every call
        ],
    )
    def test_collect_cadence(self, monkeypatch, frequency, n_calls, expected_collects):
        mock_gc = Mock()
        with monkeypatch.context() as m:
            m.setattr("gadopt.gplates.connectors.gc.collect", mock_gc)
            conn = ScalarFieldConnector(
                _DataSource(), GlobalLayerIndicator(), gc_collect_frequency=frequency
            )
            target = _target_coords()
            # Distinct ages (spaced > delta_t=1.0, all <= oldest_age=100) so every
            # call is a cache miss and runs _compute (where the gc counter lives).
            for age in range(90, 90 - 10 * n_calls, -10):
                conn.get_indicator(target, conn.source.age2ndtime(float(age)))
        assert mock_gc.call_count == expected_collects
```
Entirely pointless `_drive` function. Also use `monkeypatch.context()` to avoid affecting garbage collection in the rest of the tests

tests/unit/test_connectors.py
```diff
@@ -0,0 +1,926 @@
...
+    def test_gc_collect_frequency_validated(self):
+        src = _DummySource({"thickness"})
+        with pytest.raises(ValueError, match="gc_collect_frequency"):
+            ScalarFieldConnector(src, GlobalLayerIndicator(), gc_collect_frequency=0)
+
+    def test_defaults_use_module_level_configs(self):
+        src = _DummySource({"thickness"})
+        conn = ScalarFieldConnector(src, GlobalLayerIndicator())
+        assert isinstance(conn.mesh, MeshConfig)
+        assert isinstance(conn.interpolation, InterpolationConfig)
```
dsroberts, line 202: `isinstance(conn.mesh, MeshConfig)` isn't sufficient to determine if `conn` has used the default config. A `MeshConfig` object can be constructed with kwargs, which will change the configuration but still pass this test. Probably best to do 
```
assert conn.mesh == MeshConfig()
assert conn.interpolation == InterpolationConfig()
```

tests/unit/test_connectors.py
```diff
@@ -0,0 +1,926 @@
...
+
+def _target_coords():
+    rng = np.random.default_rng(1)
+    xyz = rng.normal(size=(15, 3))
+    return 2.0 * xyz / np.linalg.norm(xyz, axis=1, keepdims=True)
+
+
+class _MaskedSource(_CapSource):
+    """Stands in for a polygon-style source: provides ``masked_thickness`` and
+    ``membership`` rather than a plain ``thickness`` channel."""
```
dsroberts, line 390: What's this class for? What makes it different to a `_CapSource`?

tests/unit/test_connectors.py
```diff
@@ -0,0 +1,926 @@
...
+        polar = np.linspace(0.30, 0.70, 40)  # cap edge is at 0.5 rad
+
+        for sigma in self.SIGMAS:
+            surface = self._field(src, sigma, polar, 0.0, mesh)
+            above_base = self._field(src, sigma, polar, 190.0, mesh)
+
+            # Where the region is present at all, the two agree to machine
+            # precision — measured max difference 7e-16, not a loose bound.
+            covered = surface > MEMBERSHIP_FLOOR
+            assert covered.sum() > 10, "sweep must straddle the boundary"
```
dsroberts, line 488: 10 is a magic number here. What is the significance of this value?

tests/unit/test_connectors.py
```diff
@@ -0,0 +1,926 @@
...
+        observed = _evaluate_connectors_lockstep({
+            "polygon_indicator": factory.indicator,
+            "polygon_geotherm": factory.geotherm,
+        }, regression_mesh, Q, TEST_AGES)
+        for name in ("polygon_indicator", "polygon_geotherm"):
+            for age in TEST_AGES:
+                _check_reduced(observed[name][age], ref[name][age],
+                               f"{name} age {age}")
+
+    def test_reference_is_age_sensitive(self):
```
dsroberts, line 820: This test and the one that follows is confusing. is this just testing the reference data? How can you treat the reference data as authoritative in the two tests above but not in the following two tests?

tests/unit/test_connectors.py
dsroberts, line 1: `TestConnectorRegression` and `TestSharedSourceConsistency` contain quite long tests. Given `test_lithosphere_pair` and `test_polygon_pair` are dependent on numerical results, I'm wondering if its more appropriate to move them out to the `tests` directory and use a smaller mesh for the remaining unit tests to speed them up a bit?

dsroberts (commented):
Comments on `test_connectors.py` and `test_chord_threshold.py` and the stuff in `tests/unit/data`. Rest of the tests to follow.

tests/unit/test_outputs.py
```diff
@@ -35,3 +60,856 @@ def test_rejects_nonpositive_r_outer(self):
...
+        eps = 1e-6 * width
+
+        def slope_at(r):
+            return (
+                radial_quintic_step(r + eps, base_r, width)
+                - radial_quintic_step(r - eps, base_r, width)
+            ) / (2 * eps)
+        mid_slope = slope_at(base_r - width / 2)
+        assert abs(slope_at(base_r - eps)) < 1e-6 * mid_slope
+        assert abs(slope_at(base_r - width + eps)) < 1e-6 * mid_slope
```
dsroberts, line 201:
```suggestion
        assert pytest.approx(slope_at(base_r - eps), abs=1e-6 * mid_slope) == 0
        assert pytest.approx(slope_at(base_r - width + eps), abs=1e-6 * mid_slope) == 0
```

tests/unit/test_outputs.py
```diff
@@ -35,3 +60,856 @@ def test_rejects_nonpositive_r_outer(self):
...
+        # A zero base depth represents no lithosphere.
+        result = continental_linear(np.array([50e3]), np.array([0.0]))
+        np.testing.assert_allclose(result, 0.0, atol=1e-12)
+
+
+# ---------------------------------------------------------------------------
+# radial_quintic_step — shared radial primitive
+# ---------------------------------------------------------------------------
+
+class TestRadialQuinticTransition:
```
dsroberts: This seems like a really specific set of tests for what is effectively a single line of code. Are there other transitions one might want to implement later, and should any potential transition function meet all of the requirements set out in these tests?

dsroberts, line 148: These tests seem really specific for what is effectively a single line of code. I get that you need to constrain the behaviour of the transition function should you need to change it one day, in that case perhaps the function call (`radial_quintic_step`) should be parameterised such that you don't need to change it in 17 places if you switch to a logistic function or something.

tests/unit/test_outputs.py
```diff
@@ -35,3 +60,856 @@ def test_rejects_nonpositive_r_outer(self):
...
+        surface = out.compute(
+            self._interp(0.0), np.array([mesh.r_outer]), np.array([False]), mesh
+        )
+        np.testing.assert_array_equal(surface, 1.0)
+        below = out.compute(
+            self._interp(0.0),
+            np.array([mesh.r_outer - width / mesh.depth_scale]),
+            np.array([False]), mesh,
+        )
+        np.testing.assert_allclose(below, 0.0, atol=1e-30)
```
dsroberts, line 317: This is beyond the limit of precision for double precision floating point numbers.

gadopt/gplates/outputs.py
```diff
@@ -47,32 +85,799 @@ def __post_init__(self):
...
+    requires = frozenset({"lateral_weight"})
+
+    def weight(
+        self,
+        interpolated: dict[str, np.ndarray],
+        outside_source_range: np.ndarray,
+    ) -> np.ndarray | float:
+        lateral_weight = interpolated["lateral_weight"].copy()
+        lateral_weight[outside_source_range] = 1.0
+        return np.clip(lateral_weight, 0.0, 1.0)
```
dsroberts, line 541:
```suggestion
        result = np.clip(interpolated["lateral_weight"],0.0,1.0)
        result[outside_source_range] = 1.0
        return result
```
`np.clip` copies, no need to explicitly copy beforehand.

gadopt/gplates/outputs.py
```diff
@@ -47,32 +85,799 @@ def __post_init__(self):
...
+        outside_source_range: np.ndarray,
+        mesh: MeshConfig,
+    ) -> np.ndarray:
+        thickness_km = interpolated["thickness"].copy()
+        age_myr = interpolated["age"].copy()
+        thickness_km[outside_source_range] = self.fallback_thickness_km
+        age_myr[outside_source_range] = self.fallback_age_myr
+
+        depth_m = (mesh.r_outer - r_target) * mesh.depth_scale * 1e3
+        z_lab_m = thickness_km * 1e3
```
dsroberts, line 769:
```suggestion
        age_myr = interpolated["age"].copy()
        age_myr[outside_source_range] = self.fallback_age_myr

        depth_m = (mesh.r_outer - r_target) * mesh.depth_scale * 1e3
        z_lab_m = np.where(outside_source_range, self.fallback_thickness_km, interpolated["thickness"]) * 1e3
```
Avoid manual copy.

gadopt/gplates/outputs.py
```diff
@@ -47,32 +85,799 @@ def __post_init__(self):
...
+    requires = frozenset({"membership"})
+
+    def compute(
+        self,
+        interpolated: dict[str, np.ndarray],
+        r_target: np.ndarray,
+        outside_source_range: np.ndarray,
+        mesh: MeshConfig,
+    ) -> np.ndarray:
+        frac = np.clip(interpolated["membership"].copy(), 0.0, 1.0)
```
dsroberts, line 881:
```suggestion
        frac = np.clip(interpolated["membership"], 0.0, 1.0)
```
`np.clip` creates a copy

gadopt/gplates/outputs.py
```diff
@@ -47,32 +85,799 @@ def __post_init__(self):
...
+
+    def base_r(
+        self,
+        interpolated: dict[str, np.ndarray],
+        outside_source_range: np.ndarray,
+        mesh: MeshConfig,
+    ) -> np.ndarray | float:
+        thickness_km = interpolated["thickness"].copy()
+        thickness_km[outside_source_range] = self.fallback_thickness_km
+        return mesh.r_outer - thickness_km / mesh.depth_scale
```
dsroberts, line 429:
```suggestion
        return mesh.r_outer - np.where(outside_source_range, self.fallback_thickness_km,interpolated["thickness"])  / mesh.depth_scale
```
Avoid manual copy

tests/unit/test_sources.py
```diff
@@ -0,0 +1,233 @@
...
+        assert isinstance(_FakeProducer(), AgeCloudSource)
+
+    def test_rejects_a_non_producer(self):
+        with pytest.raises(TypeError, match="AgeCloudSource"):
+            PointCloudSource(object(), _DummyGplates())
+
+    def test_construction_does_not_drive_the_producer(self):
+        producer = _FakeProducer()
+        PointCloudSource(producer, _DummyGplates())
+        assert producer.at_age_calls == 0
```
dsroberts, line 121: A test of a test? I don't think this is necessary. Also, by convention we don't test cases that are handled by type checking.

tests/unit/test_sources.py
dsroberts, line 1: This test is a bit of a mess. Once again `Mock` has been re-invented. You've created this nice `_source` helper function and then proceeded not to use it consistently. Can we please make more of an effort to look over what the LLM has done before handing it over for review.

PR #538, opened 2026-08-18
Add GPU solver options.

dsroberts:
Reorganise solver options to enable use of GPUs in Stokes/GIA solvers if a GPU is detected on the system. Adding GPU offload requires a significant reorganisation of solver options, solver options dicts have been reorganised into "inner" and "outer" sections, where the "inner" section contains the GPU-specific options. GIA iterative solver options have also been moved into `stokes_integrators.py` and given a GPU configuration as well. `set_solver_options` has been refactored to take a default direct and iterative solver configuration and has been split into three main functions, `_use_iterative_solve`, `_handle_gpu_settings` and `_set_monitoring_config`.

~~Currently blocked by~~ https://github.com/firedrakeproject/firedrake/pull/5164 is now merged! 🎉.

### DIFF
--- a/gadopt/stokes_integrators.py
+++ b/gadopt/stokes_integrators.py
@@ -65,6 +65,7 @@ iterative_fieldsplit_0_gpu_parameters: ConfigType = {
             "pc_type": "python",
             "pc_python_type": "firedrake.OffloadPC",
             "offload": {
+                "ksp_type": "preonly",
                 "pc_type": "ksp",
                 "ksp": {
                     "ksp_type": "cg",
@@ -83,12 +84,10 @@ iterative_fieldsplit_0_gpu_parameters: ConfigType = {
     }
 }
 
-iterative_cuda_workarounds_inner: dict[str, dict[str, bool]] = {
-    "ksp": {
-        "pc_gamg_square_0_mat_product_algorithm_backend_cpu": True,
-        "pc_gamg_square_1_mat_product_algorithm_backend_cpu": True,
-        "pc_gamg_square_2_mat_product_algorithm_backend_cpu": True,
-    }
+iterative_cuda_ksp_workarounds_inner: dict[str, bool] = {
+    "pc_gamg_square_0_mat_product_algorithm_backend_cpu": True,
+    "pc_gamg_square_1_mat_product_algorithm_backend_cpu": True,
+    "pc_gamg_square_2_mat_product_algorithm_backend_cpu": True,
 }
 
 # Missing parameter: "fieldsplit_0"
@@ -197,11 +196,13 @@ gia_iterative_cpu_assembled_parameters: dict[str, str | float | int | bool] = {
 }
 
 gia_iterative_gpu_assembled_parameters: ConfigType = {
+    "ksp_type": "preonly",
     "assembled": {
         "ksp_type": "preonly",
         "pc_type": "python",
         "pc_python_type": "firedrake.OffloadPC",
         "offload": {
+            "ksp_type": "preonly",
             "pc_type": "ksp",
             "ksp": {
                 "pc_type": "gamg",
@@ -532,16 +533,14 @@ class StokesSolverBase(SolverConfigurationMixin, abc.ABC):
                 odb = fd.PETSc.Options()
                 odb["matmatmult_backend_cpu"] = None
                 odb["matptap_backend_cpu"] = None
+                odb["inner_C_loc_mat_product_algorithm_backend_cpu"] = None
+                odb["inner_C_oth_mat_product_algorithm_backend_cpu"] = None
                 if gpu_telescope_factor == 1:
-                    for k, v in iterative_cuda_workarounds_inner:
-                        offload_conf["assembled"]["offload"][k] = v
-                    odb["inner_C_loc_mat_product_algorithm_backend_cpu"] = None
-                    odb["inner_C_oth_mat_product_algorithm_backend_cpu"] = None
+                    for k, v in iterative_cuda_ksp_workarounds_inner.items():
+                        offload_conf["assembled"]["offload"]["ksp"][k] = v
                 else:
-                    for k, v in iterative_cuda_workarounds_inner:
-                        offload_conf["fieldsplit_0"]["assembled"]["offload"][
-                            "telescope"
-                        ][k] = v
+                    for k, v in iterative_cuda_ksp_workarounds_inner.items():
+                        offload_conf["assembled"]["offload"]["telescope"]["ksp"][k] = v
             return offload_conf
 
     def _set_monitoring_config(
--- a/gadopt/stokes_integrators.py
+++ b/gadopt/stokes_integrators.py
@@ -477,37 +477,16 @@ class StokesSolverBase(SolverConfigurationMixin, abc.ABC):
             telescope_factor: If PCTELESCOPE is in use, the telescope reduction
             factor (see https://petsc.org/main/manualpages/PC/PCTELESCOPE/)
         """
+        if device_type is not None:
+            if telescope_factor == 1:
+                d = {"assembled": {"offload": {"ksp": d}}}
+            else:
+                d = {"assembled": {"offload": {"telescope": {"ksp": d}}}}
         top_pc = self.solver_parameters.get("pc_type")
         if top_pc == "fieldsplit":
-            if device_type is None:
-                self.add_to_solver_config({"fieldsplit_0": d})
-            else:
-                if telescope_factor == 1:
-                    self.add_to_solver_config(
-                        {"fieldsplit_0": {"assembled": {"offload": {"ksp": d}}}}
-                    )
-                else:
-                    self.add_to_solver_config(
-                        {
-                            "fieldsplit_0": {
-                                "assembled": {
-                                    "offload": {"telescope": {"ksp": d}}
-                                }
-                            }
-                        }
-                    )
+            self.add_to_solver_config({"fieldsplit_0": d})
         else:
-            if device_type is None:
-                self.add_to_solver_config(d)
-            else:
-                if telescope_factor == 1:
-                    self.add_to_solver_config(
-                        {"assembled": {"offload": {"ksp": d}}}
-                    )
-                else:
-                    self.add_to_solver_config(
-                        {"assembled": {"offload": {"telescope": {"ksp": d}}}}
-                    )
+            self.add_to_solver_config(d)
 
     def _handle_gpu_settings(
         self,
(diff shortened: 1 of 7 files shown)


dsroberts:
I think this is good to go now. I've verified it works on Gadi GPU nodes and on Will's PC, just have to do some tests on Setonix. The multi-material CI test seems flaky, but that's using a direct solve, this PR doesn't touch those solver options.

The actual GPU functionality needs https://github.com/firedrakeproject/firedrake/pull/5164 to work, do not merge before that goes in.

gadopt/stokes_integrators.py
```diff
@@ -136,33 +177,71 @@
...
+    "ksp_type": "preonly",
+    "assembled": {
+        "ksp_type": "preonly",
+        "pc_type": "python",
+        "pc_python_type": "firedrake.OffloadPC",
+        "offload": {
+            "ksp_type": "preonly",
+            "pc_type": "ksp",
+            "ksp": {
+                "pc_type": "gamg",
```
sghelichkhani: There is a bit of a difference in how you pass the `ksp_type` here. You don't define `"cg"` and tolerances for `assembled.offload.ksp`. What am I missing?

gadopt/stokes_integrators.py
```diff
@@ -373,54 +456,241 @@ def set_form(self) -> None:
...
+        )
+        if device_type is None:
+            return in_config
         else:
-            self.add_to_solver_config(iterative_stokes_solver_parameters)
+            # in_config can be one of the predefined config dicts, make sure it is not
+            # modified
+            offload_conf: ConfigType = deepcopy(in_config)
+            if gpu_telescope_factor > 1:
+                ksp_params = in_config["assembled"]["offload"]
```
sghelichkhani: Can you say why you are not using `offload_conf` on this line? Doesn't it defeat the purpose?

sghelichkhani: Also it feels like the sort of thing that wee need to have test for? Don't know how though ... usually by intentionally injecting params?

gadopt/stokes_integrators.py
```diff
@@ -373,54 +456,241 @@ def set_form(self) -> None:
...
+                # override it.
+                self.add_to_solver_config({"snes_type": "ksponly"})
+            else:
+                self.add_to_solver_config(newton_stokes_solver_parameters)
+
+        device_type = get_device_type(gpu_extras)
+        if iterative_solve:
+            if self.solver_parameters.get("pc_type") == "fieldsplit":
+                if "fieldsplit_0" not in self.solver_parameters:
+                    # If fieldsplit_0 is missing, decide which set of parameters
```
sghelichkhani: I don't think the logic holds knowing that couple gia solver is using fieldsplit. Are we just ignoring mixed solver for now?

sghelichkhani: I have had separate conversations with Will about this, and I believe we need to unify the two paths, and keep the more general form of the Coupled case, as it is capable of solving a powerlaw rheology.

sghelichkhani (changes requested):
Thanks Dale! This is great work! I think there are a few issues (if not my misunderstanding) about the setup. The gia coupled solver seems to be falling through the cracks which is alright for now. But my thinking right now is that the one that we are probably going to drop is being the focus of this PR. 
I am also thinking it is appropriate to define some of the repeating solver parameters.  Like a structure like:

```
spd_ksp_parameters
gamg_common_parameters
cpu_gamg_parameters
gpu_gamg_parameters
build_spd_solver_parameters(...)
```

I also needed to learn about the reason for swithcing to preonly and then passing on the preconditioners through the offloading, which might be useful for you to teach us at the dev meeting.

gadopt/stokes_integrators.py
dsroberts: So this is OK because ksp_params is never written to again. However, it does appear inconsistent, so I'll change it. We do have tests along these lines in https://github.com/g-adopt/g-adopt/blob/main/tests/unit/test_stokes_solver_configuration.py, but if you think they aren't comprehensive enough I'm happy to add more.

gadopt/stokes_integrators.py
dsroberts: For now, yes, I couldn't think of a good way to do this. However, for now at least coupled GIA has the exact same `fieldsplit_0` settings as iterative Stokes, so maybe the same logic can be used?

gadopt/stokes_integrators.py
dsroberts: Good catch, thanks. Will fix this up.

dsroberts:
Thanks for taking a look at this.

> I am also thinking it is appropriate to define some of the repeating solver parameters. Like a structure like:
> 
> ```
> spd_ksp_parameters
> gamg_common_parameters
> cpu_gamg_parameters
> gpu_gamg_parameters
> build_spd_solver_parameters(...)
> ```

I like this idea, the code would be a lot neater for it.

gadopt/stokes_integrators.py
w3168: I think the coupled gia solver will always use fieldsplit for iterative solvers.

gadopt/stokes_integrators.py
```diff
@@ -136,33 +177,73 @@
   dictionary via the `solver_parameters_extra` argument. This dictionary can also hold
   new pairs of keys and values to extend the default ones.
 """
+gia_outer_solver_parameters: dict[str, str | float] = {
```
w3168: Does this need to be separate from FS0?

gadopt/stokes_integrators.py
```diff
@@ -136,33 +177,73 @@
...
+    "mat_type": "matfree",
+    "snes_type": "ksponly",
+    "pc_type": "python",
+    "pc_python_type": "gadopt.SPDAssembledPC",
+}
+
+gia_iterative_cpu_assembled_parameters: dict[str, str | float | int | bool] = {
```
w3168: I think as @sghelichkhani mentioned - these are are just a copy of the mantle convection defaults (basically because laplacian of velocity and laplacian of displacement are similar)

w3168 (commented):
Thanks Dale this is looking good! 

Can we talk with Sia about setting up a parallel scaling test comparing the Coupled and Interanal variable Solver

gadopt/stokes_integrators.py
sghelichkhani: Yeah! I think that is fine! You never know when testing is enough TBH! Thanks

sghelichkhani:
> Can we talk with Sia about setting up a parallel scaling test comparing the Coupled and Interanal variable Solver

As long as Dale is ok with revisiting this a bit later this can go in and we can revisit this once we figure out if we can move to one of the formulations later.

gadopt/stokes_integrators.py
sghelichkhani: Yes! That is what I meant. Currently the coupled GIA solver received 
fieldsplit_0 -> SPDAss -> CPU GAMG with SOR.  so displacement block does not use the new GPU solver. Is that the intended behaviour here? Have we tried to see if it works with the GPU? should the dispatch explicitly select that backend for coupled GIA?

sghelichkhani: Looking at it again, I think the existing dispatch can handle this if we leave `fieldsplit_0` out of `coupled_gia_solver_parameters`. It would then insert the CPU block on the host and the GPU block for CUDA/HIP, while the coupled solver keeps its own outer fieldsplit and `fieldsplit_1`. I don't think we can just merge the GPU dictionary into the current block because the old CPU `assembled_*` options would remain. Does that sound like the intended setup? It would be useful to have one configuration test that forces CUDA and checks that coupled GIA actually contains `fieldsplit_0.assembled.offload`.

dsroberts: Can do. I built in two overrides for GPU detection, so its pretty easy to trick G-ADOPT into thinking a system has a GPU, so I can put in this test as well.

dsroberts:
I'm not entirely happy with the way `test_stokes_solver_configuration.py` has come out. Rolling in the GIA settings and the additional nesting from the GPU configs have really increased the complexity. I think I need to learn a little more about how fixtures work because I think maybe the right way to do this would be similar to how `test_symmetry.py` looks with the fixture parameters providing the different variants of the tests rather than `pytest.mark.parametrize`. Right now I think there are too many variants of `test_solver_parameters`, using fixture params to get it down to a single test would probably be best.

dsroberts:
I'm happy with where this is at now. Most of the complexity is around constructing the expected parameter sets. There are now a lot of different parameter sets to test, and to test properly these parameter sets need to be constructed independently of the machinery in the solvers. Coverage report confirms everything except print statements in `stokes_solver_manager.py` and everything that has an effect on `StokesSolver.solver_parameters` is tested.

w3168:
Thanks Dale I think the unifying of the solver parameters that you made in [758ce88](https://github.com/g-adopt/g-adopt/pull/538/commits/758ce88332d1603ec258116e55467517afadc61f) looks good from my point of view.

gadopt/stokes_integrators.py
```diff
@@ -373,10 +431,190 @@ def set_form(self) -> None:
...
+                self.add_to_solver_config(d)
+            else:
+                if telescope_factor == 1:
+                    self.add_to_solver_config(
+                        {"assembled": {"offload": {"ksp": d}}}
+                    )
+                else:
+                    self.add_to_solver_config(
+                        {"assembled": {"offload": {"telescope": {"ksp": d}}}}
+                    )
```
angus-g:
```suggestion
        if device_type is not None:
            if telescope_factor == 1:
                d = {"assembled": {"offload": {"ksp": d}}}
            else:
                d = {"assembled": {"offload": {"telescope": {"ksp": d}}}}


        top_pc = self.solver_parameters.get("pc_type")
        if top_pc == "fieldsplit":
                self.add_to_solver_config({"fieldsplit_0": d})
        else:
                self.add_to_solver_config(d)
```

maybe something like this to flatten the nesting and separate the conditions a bit better?

gadopt/stokes_integrators.py
```diff
@@ -373,10 +431,190 @@ def set_form(self) -> None:
...
+                }
+            if device_type == "CUDA":
+                # Add cuda workarounds
+                odb = fd.PETSc.Options()
+                odb.setValue("matmatmult_backend_cpu", True)
+                odb.setValue("matptap_backend_cpu", True)
+                odb.setValue("inner_C_loc_mat_product_algorithm_backend_cpu", True)
+                odb.setValue("inner_C_oth_mat_product_algorithm_backend_cpu", True)
+                self._add_to_offloaded_solver(iterative_cuda_ksp_workarounds_inner, device_type, gpu_telescope_factor)
+            return offload_conf
```
angus-g:
```suggestion

        # in_config can be one of the predefined config dicts, make sure it is not
        # modified
        offload_conf = deepcopy(in_config)
        if gpu_telescope_factor > 1:
            ksp_params = offload_conf["assembled"]["offload"]
            offload_conf["assembled"]["offload"] = {
                "ksp_type": "preonly",
                "pc_type": "telescope",
                "pc_telescope_reduction_factor": gpu_telescope_factor,
                "telescope": ksp_params,
            }
        if device_type == "CUDA":
            # Add cuda workarounds
            odb = fd.PETSc.Options()
            odb.setValue("matmatmult_backend_cpu", True)
            odb.setValue("matptap_backend_cpu", True)
            odb.setValue("inner_C_loc_mat_product_algorithm_backend_cpu", True)
            odb.setValue("inner_C_oth_mat_product_algorithm_backend_cpu", True)
            self._add_to_offloaded_solver(iterative_cuda_ksp_workarounds_inner, device_type, gpu_telescope_factor)
        return offload_conf
```

drop a level of nesting after return if first branch

gadopt/stokes_integrators.py
```diff
@@ -388,39 +626,84 @@ def set_solver_options(
...
+        if "snes_type" not in self.solver_parameters:
+            if not depends_on(self.approximation.mu, self.solution):
+                # If a set of solver parameters already has a "snes_type", do not
+                # override it.
+                self.add_to_solver_config({"snes_type": "ksponly"})
+            else:
+                self.add_to_solver_config(newton_stokes_solver_parameters)
+
+        device_type = get_device_type(gpu_extras)
+        if iterative_solve:
```
angus-g, line 669: I'd consider maybe factoring this to its own function?

```python3
if iterative_solve:
    self._configure_iterative_solver(device_type, gpu_extras)
```

gadopt/stokes_integrators.py
```diff
@@ -388,39 +626,84 @@ def set_solver_options(
...
+                        }
+                        self.add_to_solver_config(
+                            {
+                                "fieldsplit_0": self._handle_gpu_settings(
+                                    gpu_extras,
+                                    device_type,
+                                    additional_params,
+                                )
+                            }
+                        )
```
angus-g: I think this can probably be refactored with the else branch similar to the `_use_iterative_solver` change I suggested: build common parameters, branch on fieldsplit to determine the dictionary prefix?

angus-g (changes requested):
Thanks for case-bashing all of this, an unenviable task! I'd like it if some of the conditional logic was tightened up so it's a bit easier to follow?

dsroberts:
Thanks @angus-g. The subtle differences between the mantle and GIA solver configurations and the CPU and GPU solver configurations make it a bit of a pain to unify them nicely, but it does look much neater now.

angus-g (approved):
Thanks, given I didn't even know there were subtle differences I think they were definitely a bit hidden. I'm fairly happy with the structure, as long as the others are happy with the parameters that end up coming out!

sghelichkhani (approved):
Who am I to disagree with giants?

stephankramer:
Late to the party, but this looks great! :tada:

PR #539, opened 2026-08-26
Revert "Symbolic Jacobian Symmetrisation to deal with nonlinear rheology in SIPG (#370)"

sghelichkhani:
This reverts the merge of #370 (commit af60605). The pull request had auto-merge enabled and it landed on main off stale status checks, from an older state of the branch.

PR #540, opened 2026-08-28
Energy-consistent SIPG boundary residual for nonlinear rheology

### LINKED ISSUE #544: CoupledInternalVariableSolver uses an asymmetric displacement block with weak normal-displacement boundaries
`CoupledInternalVariableSolver` applies CG to its displacement block through the default fieldsplit configuration. That block is expected to be symmetric. On `main`, weak normal-displacement (`"un"`) boundary conditions make it asymmetric.

This was identified while reviewing the SIPG changes in #540, although the symmetry of the coupled formulation had also been raised previously. We do not have a case in which this defect makes a shipped benchmark fail. CG can appear to converge on a mildly asymmetric operator, especially with the current tolerances. The invalid operator/solver combination can nevertheless increase iteration counts, reduce preconditioner performance, or give misleading convergence behavior.

The following minimal example assembles the coupled Jacobian and measures the relative asymmetry of its displacement-displacement block.

```python
"""Reproduce the asymmetric displacement block in CoupledInternalVariableSolver."""

from gadopt import *


# Use one Maxwell element with tau = viscosity / shear_modulus = 1.
shear_modulus = 2.0
viscosity = 2.0
maxwell_time = viscosity / shear_modulus
dt_over_tau = 25.0
dt = dt_over_tau * maxwell_time

mesh = UnitSquareMesh(3, 2)
mesh.cartesian = True

# The coupled unknown contains the displacement and one internal variable.
V = VectorFunctionSpace(mesh, "CG", 2)
S = TensorFunctionSpace(mesh, "DG", 1)
Z = V * S
z = Function(Z)

# Use a generic displacement so that the boundary terms do not vanish.
x, y = SpatialCoordinate(mesh)
z.subfunctions[0].interpolate(
    as_vector(
        [
            x + 0.2 * x * y + 0.1 * y**2,
            y + 0.3 * x**2 - 0.15 * x * y,
        ]
    )
)

DG0 = FunctionSpace(mesh, "DG", 0)
density = Function(DG0).assign(1.0)

approximation = MaxwellApproximation(
    bulk_modulus=3.0,
    density=density,
    shear_modulus=shear_modulus,
    viscosity=viscosity,
    B_mu=0.0,
)

solver = CoupledInternalVariableSolver(
    z,
    approximation,
    dt=dt,
    bcs={1: {"un": 0.0}, 2: {"un": 0.0}},
    solver_parameters="direct",
)

# Assemble the coupled Jacobian and extract its displacement block.
jacobian = assemble(derivative(solver.F, z), mat_type="nest")
A00 = jacobian.petscmat.getNestSubMatrix(0, 0).convert("aij")

# Compute ||A00 - A00^T|| / ||A00||.
A00_transpose = A00.copy()
A00_transpose.transpose()
antisymmetric_part = A00.copy()
antisymmetric_part.axpy(-1.0, A00_transpose)
relative_asymmetry = antisymmetric_part.norm() / A00.norm()

log(f"dt / tau: {dt_over_tau}")
log(f"relative asymmetry: {relative_asymmetry:.6e}")
log(f"symmetric at rtol=1e-12: {relative_asymmetry < 1.0e-12}")
```

On `main` at `7a8e73b`, this gives:

```text
dt / tau: 25.0
relative asymmetry: 3.323729e-02
symmetric at rtol=1e-12: False
```

The coupled momentum stress is

$$
\sigma(u,m)=\mu_0 A(\nabla u)+K(\nabla\cdot u)I-\sum_i2\mu_i m_i,
$$

where $\mu_0=\sum_i\mu_i$. At fixed internal variables, the displacement derivative of the consistency term therefore carries $\mu_0$. On `main`, the symmetrising boundary term instead uses

$$
\eta_{\mathrm{eff}}=\sum_i\frac{\mu_i}{1+\Delta t/\tau_i},
$$

because `CoupledInternalVariableSolver` assigns `approximation.mu = approximation.effective_viscosity(dt)`. The displacement block consequently contains the antisymmetric contribution

$$
(\mu_0-\eta_{\mathrm{eff}})(C-C^T),
$$

which grows with $\Delta t/\tau$.

PR #540 constructs the symmetrising term from the derivative of the actual stress. It also assigns $\mu_0$ as the coupled solver's Nitsche penalty coefficient. The consistency, symmetrising, and penalty terms then all use the elastic coefficient of the coupled displacement block. The reproducer on that branch gives a relative asymmetry of `4.496828e-17`.

sghelichkhani:
This re-lands the symmetric SIPG boundary work. The weak boundary terms of the momentum equation are made energy consistent for a strain-rate-dependent viscosity: the symmetrising term now uses the tangent stress Dσ(u)[φ] rather than σ(φ) with a frozen viscosity, and a penalty-derivative term proportional to Dμ[φ]·w² is added. With these two pieces the boundary residual is the exact first variation of a boundary functional, so derivative(F, z) is a genuine symmetric Newton Jacobian and no custom Jacobian is assembled. For a linear viscosity both new terms vanish and the form is byte-identical to what we had, which the nonlinear guard is careful to preserve (it tests the coefficients underlying the trial, so a linear-but-spatially-varying viscosity like 1 + X[1] stays on the frozen path).

This is the residual formulation that replaces the earlier post-hoc 0.5(J+Jᵀ) Jacobian patch (the quasi-Newton approach from #370). #370 was accidentally auto-merged on stale CI and then reverted in #539, so this is a clean rebuild on top of the post-revert main rather than a reopen.

On testing, besides the Jacobian-symmetry tests in test_symmetry.py, there is a new test_weak_bc_solution.py that pins the actual converged solution of the new residual, not just the symmetry of its Jacobian. It has two manufactured-solution convergence tests, one driving the weak "un" branch through the solver and one driving the weak "u" branch at the equation level, both recovering the expected orders, plus a weak-versus-strong free-slip cross-check on a coarsened viscoplastic (Tosi) case that confirms the weak free-slip solve agrees with the trusted strong-BC formulation and that the two agree better under refinement. Each of these was checked to actually fail under planted sign/coefficient mutations of the boundary terms, including a wrong "u"-branch residual that none of the pre-existing tests caught. The whole new file runs serially in about 13 seconds.

Closes #544


gadopt/momentum_equation.py
```diff
@@ -79,36 +95,65 @@ def viscosity_term(eq: Equation, trial: Argument | Indexed | Function) -> Form:
...
                 2
                 * sigma
                 * inner(outer(eq.n, eq.test), mu * trial_tensor_jump)
                 * eq.ds(bc_id)
             )
-            F -= inner(mu * nabla_grad(eq.test), trial_tensor_jump) * eq.ds(bc_id)
+            # Symmetrising term. For nonlinear mu this is the tangent stress
+            # contracted with the jump; for linear mu it is the frozen-mu form.
+            if mu_nonlinear:
+                tangent = eq.approximation.tangent_stress(trial, eq.test)
```
stephankramer: Shouldn't this also do the right thing for the case `not mu_nonlinear`? Note that `tangent_stress()` already handles the linear case separately, so it doesn't unnecessarily do derivative stuff. If I write it out, I think it would do the same as in the old code in the else-clause.

I was hoping that if we get rid of that else case, we could also get rid of the definition of trial_tensor_jump above, which would address a vulnerability we've discussed previously: we moved the stress definition to the approximation in #121 - which is really good, but then we still replicate the logic of how it's constructed here with "if compressible subtract 2/3 div(.)"; this mean it's very easy for these two definitions to be out-of-sync, e.g. someone might want to do a "true 2D compressible approximation" version with $$\tau=2 sym(\nabla(u)) - div(u)$$.

But I see "trial_tensor_jump" is also used in the penalty - and also the beta coefficient below in the new addition to the penalty-term needs this "if compressible" logic. Hm, does your AI have a neat solution? - ideally I would like all logic about how the stress is assembled be defined in one place in the approximation...

gadopt/momentum_equation.py
```diff
@@ -79,36 +95,65 @@ def viscosity_term(eq: Equation, trial: Argument | Indexed | Function) -> Form:
...
+                F -= w * dot(eq.n, dot(tangent, eq.n)) * eq.ds(bc_id)
```
stephankramer: Same thing here: this should also work for the `not mu_nonlinear` case

gadopt/momentum_equation.py
```diff
@@ -79,36 +95,65 @@ def viscosity_term(eq: Equation, trial: Argument | Indexed | Function) -> Form:
...
+            if mu_nonlinear:
+                tangent = eq.approximation.tangent_stress(trial, eq.test)
+                F -= dot(w, dot(tangent, eq.n)) * eq.ds(bc_id)
+            else:
+                F -= inner(mu * nabla_grad(eq.test), trial_tensor_jump) * eq.ds(bc_id)
             F -= inner(outer(eq.n, eq.test), stress) * eq.ds(bc_id)
+            # Derivative of the penalty through mu, so that the penalty is the
+            # exact second variation of the boundary functional.
+            if mu_nonlinear:
+                beta = 1 - (2 / 3 if compressible_stress else 0)
```
stephankramer: So this does still need to be a special case for `mu_nonlinear` (fine), but also relies on "if compressible subtract 2/3"-logic (not great)

tests/unit/test_weak_bc_solution.py
```diff
@@ -0,0 +1,282 @@
...
+problem and checks optimal convergence. ``test_mms_weak_u_convergence`` (1b)
+drives the weak "u" branch at the ``Equation`` level (``StokesSolver`` converts
+"u" to a strong ``DirichletBC``, so the solver never reaches that branch) on a
+manufactured velocity-only problem. ``test_tosi_weak_freeslip_matches_strong``
+(2) runs a single nonlinear Stokes solve with a realistic Tosi viscoplastic
+rheology and pins the weak free-slip ("un") solution against the trusted
+strong-BC formulation. Both MMS tests are incompressible; the compressible
+branches stay covered only by the symmetry tests. Test 2 is a coarsened,
+single-solve, fixed-temperature version of the Tosi benchmark, not the full
+time-stepped run.
```
stephankramer: This is too much historic information that's going to be meaningless in the long-run. Comments should just describe what this test currently tests in the present tense - possibly in relation to other tests: e.g. "test <other test> tests for symmetry, but not consistency which is tested for here."

tests/unit/test_weak_bc_solution.py
```diff
@@ -0,0 +1,282 @@
...
+
+Known limitation. Mutations of the *penalty coefficients alone* (for example
+``c: 2 -> 4/3`` or ``beta: 1 -> 0`` changed consistently in both the penalty
+term and the penalty-derivative term) keep the method Nitsche-consistent. They
+change stability constants, not the continuous limit, so no solver-output test
+can catch them. They are pinned, for the incompressible "un" branch only, by
+``test_viscosity_term_un_variational_structure`` in ``test_symmetry.py``.
```
stephankramer: I don't understand what any of this means. "changed" refers to what changed in this PR? Why is this a limitation of the test? "They change": who? Catch what?
I think this is AI trying to explain what it's done in this PR and trying to explain to what extent it's tested that - but that shouldn't be done in the comments. Comments should explain the current state of the code. Of course there's the occasional minor exceptions like: "testing regression of bug fixed in ..."

tests/unit/test_weak_bc_solution.py
```diff
@@ -0,0 +1,282 @@
...
+        u_w = _tosi_solve(mesh, bcs_weak, use_switch=False)
+
+        d[N] = (fd.sqrt(fd.assemble(fd.inner(u_w - u_s, u_w - u_s) * fd.dx))
+                / fd.sqrt(fd.assemble(fd.inner(u_s, u_s) * fd.dx)))
+        urms_rel[N] = abs(fd.sqrt(fd.assemble(fd.dot(u_w, u_w) * fd.dx))
+                          / fd.sqrt(fd.assemble(fd.dot(u_s, u_s) * fd.dx)) - 1)
+
+    # observed d_32=7.33e-05 on 2026-08-28, cap = 3x
+    assert d[32] <= 0.05  # sanity ceiling
+    assert d[32] <= 3 * 7.33e-05
```
stephankramer: huh?

stephankramer (commented):
Looks great! Some initial thoughts.

I can see how the symmetrisation-term works (the transpose of the IBP integral) - but the change to the penalty are still a bit of a mystery to me. I one place it's doing 1-2/3 and in another 2-2/3 which feels weird to me. Is there an explanation as to why these are the right coefficients?

gadopt/momentum_equation.py
sghelichkhani: I think in theory you are very right, we should be able to remove the if condition. The only caveat was that the gia code depends on this as approximation in gia does not have the tangent linear definitiong. 

Trick is obviously to override that for gia. But that should be  simple enough t do.

gadopt/approximations.py
```diff
@@ -28,7 +29,38 @@
 ]
 
 
-class BaseApproximation(abc.ABC):
+class DeviatoricStressMixin:
```
w3168, line 31: I think we talked before about if rheology should live in a separate module. Can we discuss this mixin approach at the meeting?

w3168 (commented):
Thanks Sia! Just wanted to flag whether the DeviatoricStressMixin is the clearest way of doing this and if down the line we should try and make the rheology vs buoyancy logic in `approximations.py` clearer?

sghelichkhani:
> Thanks Sia! Just wanted to flag whether the DeviatoricStressMixin is the clearest way of doing this and if down the line we should try and make the rheology vs buoyancy logic in `approximations.py` clearer?

Fair point regarding mixing. The two other altearnative would be a plain module-level function, or providing another commopn base for the two is there any of the two that you think particularly better? Or any other suggestion?

sghelichkhani:
@w3168  Yeah! I think Stephan's point is the same that all stress logic (LHS of stokes) should be seperated, and you are saying the RHS should have the same ...We could try that as a follow up. unless everyone is determined this is the right place to do it.

sghelichkhani:
Pushed a round of changes addressing the review comments (the penalty coefficient consolidation, the wording fixes on the boundary-term comments, and a small loosening of the MMS solver tolerances). Back to you for another look.

gadopt/approximations.py
sghelichkhani: Okido! Let's discuss it on Wednesday.

sghelichkhani: If BaseGIAApproximation and BaseApproximation were sharing a base that could have been simpler. It think with Mixing would be the only way to impose the condition on both, or you need to override that everytime.

tests/unit/test_weak_bc_solution.py
sghelichkhani: This should make better sense now.

gadopt/momentum_equation.py
sghelichkhani: Yes done!

gadopt/momentum_equation.py
sghelichkhani: Done now! Please have a look.

tests/unit/test_weak_bc_solution.py
sghelichkhani: Agreed!

tests/unit/test_weak_bc_solution.py
sghelichkhani: Done! Sorry I usually try to avoid the slop!

tests/unit/test_weak_bc_solution.py
sghelichkhani: I have a proper definition of the convergence there.

gadopt/momentum_equation.py
sghelichkhani: This is sorted now.

sghelichkhani:
> I one place it's doing 1-2/3 and in another 2-2/3 which feels weird to me. Is there an explanation as to why these are the right coefficients?

@stephankramer Those weird expansions are now gone. Both branches now just write `sigma * dmu * inner(jump_gradient, jump_tensor)` for  $\langle G, A(G)\rangle$ , and  no hardcoded 2/3 left in momentum_equation.py  which shoudl be hopefully more readable. The compressible/incompressible split then lives in `deviatoric_stress_shape` and nowhere else.

sghelichkhani:
@w3168 and @Patol75 Can you please have a look at the failing test. As far  as I can see there is no change to the actual values and it'just because BoussinesqApproximation.stress returning 2 * mu * sym(grad(u)) is gone and we now return  `mu * deviatoric_stress_shape(grad(u)) = mu * (2 * sym(grad(u)))`
So bascially, is it ok if I loosen the tolerance considering the test?

gadopt/approximations.py
angus-g: I think it makes sense to take this approach if we are indeed putting this term into approximations. I don't quite understand the argument for separating rheology out? Or is the point that the rheology is provided as an argument to the approximation? As they say: prefer composition over inheritance, so it is something to consider *if* it actually makes sense. If this change is constrained to literally just adding a deviatoric stress form, that does feel a bit premature/overengineered?

sghelichkhani: I guess we could do the same with a function, and use that function everywhere for the deviatorc stress , but then you need to pass a complressible flag to it.  The advantage here is that the feature knows if the approximation is compressible or not to give the right definition. 

Happy to do that if this is overengineering. But I think defining a shared base class for GIA and Stokes is so much more of an overengineering.

gadopt/approximations.py
```diff
@@ -488,6 +560,19 @@ def prefactor_prestress(self, dt):
     def stress(self, u, **kwargs) -> ufl.core.expr.Expr:
         return 0
 
+    def tangent_stress(self, u, v, **kwargs) -> ufl.core.expr.Expr:
```
w3168: I'm struggling to understand this comment whether we need to do something different for the nonlinear (power law) GIA rheology?

w3168: (But I agree with the part about the previous stress and bulk part of the full stress being dealt with separately for the linear case.)

gadopt/approximations.py
angus-g: Sorry, I meant that adding an entirely new module for rheology may be going too far if we only need it for this term. This is definitely preferable to having to pass compressibility into a function.

sghelichkhani: Ah! agreed fully then!

w3168: I think the DeviatoricStressMixin is fine and it seems like a good way to get the incompressible/compressible stress paths out of momentum_equation.py. I wanted to raise it in the meeting just because we have moved how stress is defined over the years so it would be good if everyone is happy.

gadopt/approximations.py
```diff
@@ -28,8 +29,58 @@
...
+
+    All logic about how the deviatoric stress is assembled from a velocity
+    gradient lives in `deviatoric_stress_per_mu`, keyed off the `compressible`
+    property. The full stress, the tangent stress, and the weak (SIPG) boundary
+    terms in the momentum equation all build on this one method, so a subclass
+    that changes the stress form (for example a true 2D compressible model) only
+    overrides here.
+    """
+
+    def deviatoric_stress_per_mu(self, gradient: ufl.core.expr.Expr) -> ufl.core.expr.Expr:
```
w3168: I think I preferred the older name 'deviatoric_stress_shape'. Did you have a look at how stress is defined in the GIAAproximations? I feel like if we renamed 'deviatoric_stress_shape' -> 'deviatoric_tensor' or something then you can resuse this to define the stress later in the gia approximation.

gadopt/approximations.py
w3168: can you use the 'deviatoric_shape' here?

gadopt/momentum_equation.py
```diff
@@ -79,36 +98,55 @@ def viscosity_term(eq: Equation, trial: Argument | Indexed | Function) -> Form:
...
             # We only keep the normal part of stress; the tangential part is assumed to
             # be zero stress (i.e. free slip) or prescribed via "stress".
             F -= dot(eq.n, eq.test) * dot(eq.n, dot(stress, eq.n)) * eq.ds(bc_id)
+            # Derivative of the penalty term through mu, as in the "u" branch
+            # above, restricted to the normal component of the jump.
+            if mu_nonlinear:
+                dmu = expand_derivatives(derivative(mu, trial, eq.test))
+                F += sigma * dmu * inner(jump_gradient, jump_tensor) * eq.ds(bc_id)
 
             if hasattr(eq.approximation, 'bulk_modulus'):
```
w3168: @sghelichkhani I think this is still valid for the (bulk) compressible part of the SIPG formulation. Can you check that your changes are still compatible with this? If it is fine I think we can leave this definition here in the momentum equation.py - what do you think?

w3168 (commented):
Thanks @sghelichkhani I've looked through this a bit more carefully - like Stephan says it's really cool that you've been able to take separate out the incompressible/compressible branches of the sipg stuff in the momentum equation. 

I've got a few questions about how fits in with the existing GIA stuff.

gadopt/approximations.py
sghelichkhani: Sure! I was more motivated by Thomas picking a problem with `shape` happy to revert. Let me have a look there.

gadopt/momentum_equation.py
sghelichkhani: hmm interesting. Let me have a look.

gadopt/momentum_equation.py
```diff
@@ -79,36 +98,55 @@ def viscosity_term(eq: Equation, trial: Argument | Indexed | Function) -> Form:
...
 
         if "u" in bc:
-            trial_tensor_jump = 2 * sym(outer(eq.n, trial - bc["u"]))
-            if compressible_stress:
-                trial_tensor_jump -= (
-                    2 / 3 * identity * (dot(eq.n, trial) - dot(eq.n, bc["u"]))
-                )
+            w = trial - bc["u"]
+            jump_gradient = outer(eq.n, w)
+            jump_tensor = eq.approximation.deviatoric_stress_per_mu(jump_gradient)
```
stephankramer: Loving it!

gadopt/approximations.py
stephankramer: Just to add to the bike-shedding 🚲 - not sure about "deviatoric_stress_shape" either, with shape I would expect something like the (dim, dim) shape of the tensor. The thing that distinguishes this version of deviatoric_stress from the various ones in the subclasses, is that it takes a grad u as input rather than u. So perhaps deviatoric_stress_from_grad ?

stephankramer:
@w3168 : If I follow things correctly, for InternalVariableApproximation the stress that's passed on to the viscosity term is $$\mu\left(2\mathop{sym}(\nabla u) - 2/3 \nabla\cdot u \mathbb{I} \right) + \kappa~\text{bsr}~\nabla \cdot u\mathbb{I} - \sum_i 2 \mu m_i$$ so the extra $$\kappa\ \text{bsr}\ \nabla\cdot u\mathbb{I}$$ is not currently taking into account in the symmetrizing and the penalty term. I think the proposed changes leave that as it is, by redefining `tangent_stress()` used in the symmetrizing term, and because the penalty term uses `deviatoric_stress_per_mu()`.

So I think we first need to decide whether that's correct (ignoring the bulk div(u) term for the Nitsche terms) for the linear case. If it's not correct, i.e. it should be included - then you could *not* redefine `tangent_stress()` - using the fact that it's based on a linearisation of stress(). The penalty term is a bit more complicated. For the linear case, we actually need a version that's *not* divided by $\mu$ - which would suggest redefining `deviatoric_stress_per_mu()` to `stress_from_grad()` which would include $\mu$ and would be defined per approximation, so for Stokes: mu (G+G^T - 2/3 Tr(G) I) and for internal variables: mu (G+G^T - 2/3 Tr(G) I) + kappa Tr(G) I. However for the nonlinear contribution to the penalty term we do need a version that only contains the terms with mu, and is not yet multiplied: G+G^T - 2/3 Tr(G) I in both cases - so how about we call that `deviatoric_stress_from_grad` defined in the Mixin, but then Stokes and InternalVariable approximations each have their own `stress_from_grad()`

When mu is nonlinear (for internal variables app.) you also have to think about what happens with the internal variable terms, they would end up in tangent_stress() which is probably not what you want.

gadopt/approximations.py
stephankramer: deviatoric_strain_rate_from_grad() ?

w3168:
Thanks Stephan I'm going to try and go through the maths again for the viscoelastic case  (I think we talked about it in late October last year)

w3168:
Hi @stephankramer and @sghelichkhani, so I've gone back to the maths and the rough notes I made about SIPG for the (linear) viscoelastic case. Here's a link to zulip conversation we had in October last year ([#GIA-Project > GIA symmetry @ 💬](https://teamgeodynamics.zulipchat.com/#narrow/channel/388111-GIA-Project/topic/GIA.20symmetry/near/546794590)) 

I am pretty sure we are currently including Nitsche terms for the bulk div(u) term, but only for the 'un' bc case. see lines 113-123 in momentum_equation.py in main [here](https://github.com/g-adopt/g-adopt/blob/7a8e73be0cd48fb2cf4e2f6dc04f9b3b1bf7f7f1/gadopt/momentum_equation.py#L113)

Also  we are using an 'effective viscosity'  for the substitute method of viscoelasticity instead of the shear modulus. I think we decided this made sense from applying the backward euler method to rewrite the internal varible in terms of the deviatoric strain (see equation 32 in our [paper](https://gmd.copernicus.org/articles/19/2717/2026/). 

I never generalised this to the 'u' bc case, which is not great... Also I think we (Stephan and I) discussed the case of interior facet terms for DG integrals but decided it would be complicated to apply the `avg(bulk)` terms consistently with stress defined in two places... (see the bullet points in the zulip message). 

@sghelichkhani I think you said that you and claude found a symmetry bug for the CoupledInternalVariableSolver ? I wonder if that is because we should not use the effective_viscosity in place of mu if we don't do the substitution of the internal variable equations (see equation 33 in our paper)... 

I still need to think about the nonlinear case, where the viscosity is a function of the deviatoric stress (so is a function of the displacement and internal variable) (see equation 16 in the paper).

sghelichkhani:
> you and claude found a symmetry bug for the CoupledInternalVariableSolver

So I went to my notes and found this. But I am not sure it is necessarily a bug. The fact is the boundary terms for the Coupled case is not symmetric, yet we are assuming it is in the solve parameters by using cg!  on L151 of stokes solver. 

More from me later.

w3168:
Ok thanks @sghelichkhani I think it sounds like a bug because the FS0 block in the CoupledInternalVariableSolver should still be symmetric. (Btw I think the line number for the `ksp:cg` option has shifted to 61 and 610(? ) now that the gpu /cpu pr went in #538 )

stephankramer:
> I am pretty sure we are currently including Nitsche terms for the bulk div(u) term, but only for the 'un' bc case.

Ah I see, yes sorry I was only looking at "u" branch (because the maths is clearer) - but I guess that's actually a dead branch because we're always using strong DirichletBC for that. Same for internal SIPG terms - only becomes important if we switch to other discretisations...

OK great, I think my point still stands: you can get rid of the "if has bulk_modulus" logic, if we use 4  helpers routines:
1. stress(): as currently
2. tangent_stress(): derivative of stress() wrt u
3. stress_from_grad(): mu (G+G^T - 2/3 Tr(G) I) vs. mu (G+G^T - 2/3 Tr(G) I) + kappa Tr(G) I
4. stress_per_mu_from_grad(): G+G^T - 2/3 Tr(G) I

The only complication is if $\mu$ starts being nonlinear for GIA: I guess you could include internal variable contributions in tangent_stress() and thereby still symmetrize it? Not sure what the correct penalty term is in that case. In any case this would only work if you'd make $\mu$ a function of the 2nd invariant of your entire stress expression, not sure if that makes physical sense - otherwise maybe it's just not meant to be a symmetric operator in that case.

sghelichkhani:
> Ok thanks @sghelichkhani I think it sounds like a bug because the FS0 block in the CoupledInternalVariableSolver should still be symmetric. (Btw I think the line number for the `ksp:cg` option has shifted to 61 and 610(? ) now that the gpu /cpu pr went in #538 )

Yes sorry I had not moved my main fwd. 

nonetheless, I am pretty sure the Coupled Solver is asymmetric as we are using mu and mu effective in the terms that are supposed to sustain symmetry. And that assymetry effectly will scale with d t / tau. 

The reason I said this might not be a bug is that thinking about it I thought the mu eff is coming from the internal variable block which might be changing. I will have a look once I am on my laptop

sghelichkhani:
> 
> The only complication is if $\mu$ starts being nonlinear for GIA: I guess you could include internal variable contributions in tangent_stress() and thereby still symmetrize it? Not sure what the correct penalty term is in that case. In any case this would only work if you'd make $\mu$ a function of the 2nd invariant of your entire stress expression, not sure if that makes physical sense - otherwise maybe it's just not meant to be a symmetric operator in that case.

I have to say I still find the physics of internal variable a bit unusual. BUT, I am pretty sure the nonlinearity comes in form of stress dependence on internal variable. So the displacement block is always linear. Only the whole system becomes asymetric.

sghelichkhani:
image

sghelichkhani:
> image

I think I need to put this as a test if it works out. Thanks @stephankramer and @w3168

sghelichkhani:
Ok I have now pushed the changes @stephankramer suggested and it all looks very neat and consistent. We are now doing symmetry changes for the Coupled solver, which is nice and laying foundartion for the future clean up of the gia solvers.

demos/test_all.py
```diff
@@ -33,7 +33,9 @@
         "extra_checks": ["nu_top", "T_min", "T_max", "entrainment"]
     },
     f"{mm_path}/compositional_buoyancy": {"extra_checks": ["entrainment"]},
-    f"{mm_path}/free_surface": {"extra_checks": ["slab_tip_depth"]},
+    f"{mm_path}/free_surface": {
```
sghelichkhani, line 36: noting this is not related to this PR, just getting rid of a flakey test failure.

gadopt/approximations.py
sghelichkhani: Sorry will, this is now moved around. Can you have another look and raise an issue with this. Not clear what it was refering to. Sorry for not responding to this earlier.

gadopt/approximations.py
sghelichkhani: I went with Stephan! Just because it is harder to disagree with him! Sorry @w3168

sghelichkhani: Originally changed to responde to @Patol75 dislike of the name `shape`

gadopt/approximations.py
sghelichkhani: Again! Probably already done that, can you have another look here?

gadopt/momentum_equation.py
sghelichkhani: This is now gone. Not sure if you are happy with this or not. If so please resolve this.

sghelichkhani:
Failures are unrelated and documented in https://github.com/g-adopt/g-adopt/issues/543

sghelichkhani:
Okay both, I went back and forth a bit on the penalty coefficient for `CoupledInternalVariableSolver`. The substitute method inserts the backward-Euler update of the internal variables into the stress. Its displacement operator therefore has the effective shear modulus

$$
\mu_{\mathrm{eff}} = \sum_i \frac{\mu_i}{1+\Delta t/\tau_i},
$$

despite the existing “effective viscosity” name. Using $\mu_{\mathrm{eff}}$ in its consistency, symmetry, and penalty terms is therefore natural.

The coupled solver is different. It keeps the internal variables as unknowns, so its displacement block is elastic when the internal variables are held fixed. The consistency term and the derivative used for the symmetry term both carry $\mu_0=\sum_i\mu_i$. From the standard Nitsche point of view, the penalty for this block should therefore also use $\mu_0$.

My hesitation was that, after eliminating the internal variables, the coupled system represents the same time-step problem as the substitute method. From that perspective, $\mu_{\mathrm{eff}}$ looks like the natural penalty scale. For a single Maxwell element,

$$
\frac{\mu_0}{\mu_{\mathrm{eff}}}=1+\frac{\Delta t}{\tau},
$$

so using $\mu_0$ gives a larger penalty than the scale of the substituted operator. However, using $\mu_{\mathrm{eff}}$ in the coupled solver means that the penalty has a different shear coefficient from the consistency and symmetry terms.

Will and I therefore settled on $\mu_0$ for the coupled solver. This gives the standard symmetric Nitsche form for its displacement block, while the substitute method continues to use $\mu_{\mathrm{eff}}$. Both formulations remain consistent because their difference is proportional to the boundary-condition error and vanishes when the weak condition is satisfied.

@stephankramer @w3168, please double-check that you agree with this reasoning and that we are not missing a reason to scale the coupled penalty with the substituted operator instead.

w3168:
I agree with all the above! 
> 
> My hesitation was that, after eliminating the internal variables, the coupled system represents the same time-step problem as the substitute method. From that perspective, μ eff looks like the natural penalty scale. For a single Maxwell element,
> 

One small comment I think 'after eliminating the internal variables' sounds to me like you are still describing a schur complement factorisation, whereas on main our default is to solve the coupled system using symmetric multiplicative fieldsplit preconditioning, which is why *probably* it makes most sense to use $\mu_0$ and not $\eta_{eff}$ for the nitche penalty term.  

Anyway I think from our conversations offline we should definitely try using a schur complement factorisation to eliminate the internal variable as it should be an easier solve and potentially approach the speedup of the substitute (pointwise 😉 ) method!

gadopt/approximations.py
```diff
@@ -670,24 +747,99 @@ def __init__(
...
+        N.b. for 2d simulations dividing by 1/3 (instead of 1/2) may be slightly
+        inconsistent. But that seems to be the definition in geodynamic
+        community and therefore analytical tests in
+        tests/viscoelastic_internal_variable/ are setup assuming 3D geometry.
+
+        Returns:
+          A UFL expression for the deviatoric strain.
+
+        """
+        return 0.5 * self.stress_per_mu_from_grad(grad(u))
```
w3168: I raised this offline already but I feel like this is a bit of a step backwards in terms of readability. This is probably bikeshedding again but I wonder if it would make sense to call it a deviatoric_geodynamic style function like we talked about (that assumes that we always divide by 3 for the trace part even in 2d test cases to be consistent with other geodynamic tests...)  and then we don't need to call it divide by mu...

w3168: anyway at the same time it's hard to name things so maybe we (I?) shouldnt get too excited

gadopt/approximations.py
w3168: no worries this got rejigged.

gadopt/approximations.py
w3168: this got resolved.

demos/test_all.py
w3168: I know it's annoying but probably better to just put this in a different pr, since it was responsible for breaking main last week...

w3168:
Also do we have any tests that  that are using the SIPG formulation on interior facets with DG for velocity (or displacement in GIA) (or maybe groundwater...??) 

Probably we should raise an issue so that we don't forget to come back to this at somepoint and make sure everything is still consistent... Or fix now...?

demos/test_all.py
sghelichkhani: Done!

gadopt/approximations.py
sghelichkhani: Done!

gadopt/approximations.py
sghelichkhani: Yeah so per_mu was probably not the best naming. I gave it a bit of though and then went with `deviatoric_tensor_from_grad` instead of a `deviatoric_geodynamic`. In places we called this `per_mu` was  accurate at exactly one of the five cases. `stress_from_grad` multiplies by `self.mu`, but `IncompressibleMaxwellApproximation.stress` multiplies by `effective_viscosity(dt)`, `MaxwellApproximation.deviatoric_stress` multiplies by `mu0`, `deviatoric_strain` multiplies by 0.5 with no modulus at all, and the three SIPG calls in `momentum_equation` pass `tensor_jump(n, w)`, which is neither a stress nor a strain. On the GIA side `self.mu` is a scale the solver assigns before assembly anyway!

Why avoiding naming after a physical quantity: The same expression takes a velocity gradient in the Stokes problems, where it is twice the strain rate with units of 1/s and pairs with a viscosity in Pa s. An incremental displacement gradient in the viscoelastic and GIA classes, where it is twice the strain, dimensionless, and pairs with a shear modulus in Pa. So `strain_rate` would be wrong for GIA and `strain` wrong for Stokes, and the returned tensor has no fixed units. That is your original `deviatoric_tensor` suggestion, with Stephan's `_from_grad` suffix.

On the readability, I think the docstring was still vasgue, so I have written out the three cases explicitly, said which coefficient each caller supplies, and added a note that with the fixed 2/3 the compressible branch is not trace-free in 2D, since in 2D the deviator subtracts half the trace. I left the 2/3 behaviour alone and kept the convention in the docstring instead (so not in the name as that was being vague too).

`DeviatoricStressMixin` stays the same, since it still owns `stress_from_grad`, which really is a stress. The test helper `dev_stress_per_mu` became `deviatoric_tensor` to match.

w3168: thanks sia!

sghelichkhani:
Good point, I have raised #548 for this.

Short version is no, we have no coverage of the interior facet path at all. Nothing in tests or demos puts velocity or incremental displacement in a discontinuous space, so `is_continuous(eq.trial_space)` is always true and the whole dS block never runs in CI. The two tests called `*_DG` are misleading here, they put temperature in DG and keep velocity at CG2.

For the nonlinear case there is a NotImplementedError guard now, so at least it refuses loudly instead of quietly assembling an asymmetric Jacobian, which is what would have happened before. I would leave it at that for the moment.

The linear dS path is the one I am slightly less comfortable about, since it is live code today and this PR did rewrite the expression in it. But nothing in CI checks it, so it is worth coming back to.

OK for an approval whenever you are ready.

stephankramer (approved):
I'm happy with everything now. Thanks @sghelichkhani for being very responsive to all our nitpicks. I think it's improved the core discretisation code a lot - at the expense of some (but reasonable) complication in the approximations - but this should be very helpful for future new discretisations as well. Assuming @w3168 is happy with the GIA changes, I think this is good to go!

angus-g:
Just flagging that this is 3-4 times slower in pytest (30 vs 7 mins), lots of fairly intense tests and we don't run pytest optimally... no parallelisation

angus-g (changes requested):
Move MMS tests to tests/

sghelichkhani:
@angus-g Once it goes green, it would be great to have you review of the tests at least.

angus-g (approved):
Thanks for the restructure, back to previous speeds. Tests feel very "LLM" but I'm not going to have to fix things if you fail them! 😁

PR #541, opened 2026-08-31
RFC: Explore 3-D conformal Killing near-nullspaces for TALA/ALA GAMG

drhodrid:
## Purpose

This is an RFC and experimental pull request. It is intended to share an investigation into whether the additional conformal Killing modes of the three-dimensional deviatoric strain operator can improve GAMG preconditioning for compressible Stokes systems, particularly TALA and ALA models.

I am not proposing that these modes become a default at this stage. The branch contains useful positive results in controlled weak-free-slip shell problems, but it also contains an important negative result from a production-scale GPlates model. The main purpose of the PR is to expose the implementation, evidence, and open questions so that the G-ADOPT community can decide whether any part of this is worth developing further.

## Mathematical motivation

For a three-dimensional TALA or ALA viscous stress, the relevant strain-rate tensor is the trace-free symmetric gradient

```text
epsilon_dev(u) = sym(grad(u)) - div(u) I / 3.
```

In Euclidean three-dimensional space, the kernel of this operator is the ten-dimensional conformal Killing space. A convenient basis consists of:

```text
Translations:          T_i = e_i
Rotations:             R_i = e_i x x
Dilation:              D   = x
Special conformal:     K_i = 2 (e_i . x) x - |x|^2 e_i
```

There are therefore:

* three translations;
* three rotations;
* one dilation mode; and
* three special-conformal modes.

The usual three-dimensional rigid-body near-nullspace contains the first six modes. This branch investigates whether the remaining four modes provide useful additional information to GAMG when the viscous operator is based on the deviatoric strain rate.

These are **near-nullspace candidates**, not necessarily exact null modes of the complete discretised Stokes problem. Boundary conditions, curved geometry, coefficient variation, Nitsche terms, pressure coupling, and discretisation all change whether a global conformal field represents a genuinely slow error component. This distinction turns out to be central to the results below.

The implementation is deliberately three-dimensional. The full six-plus-four conformal construction should not be interpreted as a corresponding ten-mode basis in two dimensions, where the mathematical structure is different.

### `ConformalKillingNearNullspace`

The branch adds a public near-nullspace constructor that can build the rigid six modes or the complete ten-mode conformal space.

The basis is:

* assembled only after the velocity space and strong boundary conditions are known;
* optionally restricted to homogeneous strongly constrained degrees of freedom;
* orthogonalised and normalised before being passed to PETSc; and
* guarded against dependent or numerically negligible vectors.

This allows the same implementation to be used on Cartesian and curved three-dimensional meshes without hard-coding a particular spherical radius or coordinate component.

### Direct ten-mode GAMG enrichment

The `direct10` experiment passes all ten global modes directly to GAMG. This is the simplest expression of the hypothesis, and in the controlled shell tests it often provides the best or nearly best iteration count.

It can, however, increase the aggregate basis rank, coarse-grid size, operator complexity, setup cost, and memory use. It is also less robust when the four conformal modes are poorly aligned with the actual low-energy error space.

### Guarded Ritz selection

`RitzConformalPC` constructs the small projected matrix

```text
H = V^T A V,
```

where `V` contains the ten conformal candidates and `A` is the velocity-block operator. It then selects six low-energy combinations for GAMG instead of passing all ten candidates directly.

The implementation includes:

* eigenvalue and spectral-gap diagnostics;
* positive-semidefinite and finite-value checks;
* optional restriction of candidates on strongly constrained boundaries; and
* a guarded fallback to the ordinary rigid-body basis when the projected problem is not trustworthy.

The intention is to retain a six-vector GAMG basis while allowing the operator and boundary conditions to choose which combinations of rigid and conformal fields are most relevant.

### Balanced six-plus-four correction

`BalancedConformalPC` keeps the ordinary rigid six modes in GAMG and treats the remaining four conformal directions through a separate symmetric low-rank correction. Schematically, the corrected inverse is

```text
Q + (I - Q A) M^-1 (I - A Q),
```

with

```text
Q = W (W^T A W)^-1 W^T,
```

where `M^-1` is the underlying GAMG preconditioner and `W` spans the four additional conformal directions after removal of the rigid contribution.

This avoids feeding a ten-vector basis directly into every GAMG aggregate. It was designed as a conservative prototype that preserves the established rigid-body hierarchy while applying an explicit coarse correction for the four extra global directions.

The correction is algebraically well conditioned in the tests reported below, but its global reductions and vector operations add cost. It also removes only four *global* components. It does not reproduce the local aggregate-wise approximation space generated by giving all ten modes directly to GAMG.

### Nonlinear SIPG and Nitsche tangent contribution

The branch also reapplies the symbolic Jacobian symmetrisation needed for nonlinear rheology in SIPG and weakly applied free-slip/Nitsche terms. This is important for testing these near-nullspaces with the intended nonlinear TALA problem and is related to the work previously discussed in PR #370.

This part is logically separable from the conformal near-nullspace experiment. If any of this work is considered for merging, I expect the tangent-stress/Jacobian change and the near-nullspace/preconditioner work should be reviewed and split independently.

## Controlled weak-free-slip shell results

The most encouraging results come from three-dimensional shell problems with weakly imposed free-slip conditions on both boundaries. These are the cases in which the global conformal fields are least disrupted by strongly clamped velocity degrees of freedom.

For unit viscosity contrast, accumulated velocity-block iterations were:

| Approximation | rotations only | rigid 6 | Ritz 6 | direct 10 |
| --- | ---: | ---: | ---: | ---: |
| Boussinesq | 72 | 90 | 63 | 62 |
| TALA | 152 | 147 | 102 | 96 |
| ALA | 131 | 128 | 88 | 84 |

At a viscosity contrast of `1e3`, the benefit became larger:

| Approximation | rotations only | Ritz 6 | direct 10 |
| --- | ---: | ---: | ---: |
| Boussinesq | 251 | 188 | 186 |
| TALA | 1028 | 332 | 325 |
| ALA | 1024 | 333 | 323 |

The behaviour was not limited to serial execution. In a two-rank TALA shell case at contrast `1e3`, the accumulated iteration counts were:

```text
rotations only: 1080
Ritz 6:          358
direct 10:       358
```

In the larger controlled shell case with 78,438 velocity degrees of freedom, the comparison was:

| Basis | Velocity iterations | Measured solve time |
| --- | ---: | ---: |
| rotations only | 920 | 60.25 s |
| Ritz 6 | 248 | 19.08 s |

These tests suggest that the conformal information can be valuable for a deviatoric velocity operator when the geometry, coefficients, and boundary conditions leave those fields reasonably representative of the slow error space.

## Production-shaped GPlates result

The production-scale test is deliberately much less favourable and is probably more informative for deciding whether this should become a general facility.

The frozen comparison used:

* 152,175,366 velocity degrees of freedom;
* 4,160 MPI ranks;
* the TALA approximation;
* imposed GPlates velocity on the surface;
* weak normal-free-slip at the CMB;
* the same checkpoint and frozen viscosity field for each solver arm;
* a nondimensional viscosity range of approximately `0.390` to `3804`;
* velocity relative tolerance `1e-5`; and
* pressure relative tolerance `1e-4`.

Mean warm velocity-block behaviour was:

| Method | Mean iterations per successful A-block solve | Mean time per solve |
| --- | ---: | ---: |
| rigid 6 | 36 | 12.85 s |
| rigid 6, constrained candidates | 36 | 12.80 s |
| balanced six-plus-four | 36 | 16.19 s |
| Ritz 6, constrained candidates | 51 | 16.92 s |
| Ritz 6 | 52 | 17.37 s |
| direct 10 | 41 | 20.00 s |

The balanced four-mode projected system was well conditioned, with a reported condition number of approximately `1.16`, but this did not reduce Krylov iterations. Its extra global work therefore made it slower than the baseline.

For `direct10`, GAMG operator complexity increased from approximately `1.084` to `1.232`, and the coarsest problem grew from about `360` to `590`. The added hierarchy cost outweighed its modest change in iterations.

The Ritz projected spectrum was approximately:

```text
[1.329, 1.341, 1.364, 1.453, 1.491, 1.521,
 2.878, 2.959, 3.059, 3.122]
```

The spectral gap was approximately `0.435`, so the selection passed its guards and did not fall back. Nevertheless, the selected space was worse than the ordinary rigid-body space for this problem. The dilation coefficient norm was only about `0.0025`, while the special-conformal coefficient norm was about `0.954`, including approximately `0.950` in the constrained-candidate variant. This is a useful example of a projected eigenproblem being numerically stable while selecting a globally plausible but practically unhelpful subspace.

## Partial nonlinear comparison

The limited nonlinear experiment produced the following accumulated warm velocity work:

| Method | Total velocity iterations | Velocity solve time | Final reported SNES norm |
| --- | ---: | ---: | ---: |
| rigid 6 | 2062 | 409.8 s | `3.489857e6` |
| balanced six-plus-four | 2062 | 506.1 s | `3.489851e6` |
| Ritz 6 | 3033 | 591.9 s | `3.489611e6` |

All arms reached the imposed three-iteration SNES limit. These numbers are therefore not converged nonlinear timings and should be interpreted only as supporting evidence that neither conformal prototype reduced the velocity-block cost in this production configuration.

## Interpretation

The controlled shell problem and production GPlates problem differ in several important ways.

### Boundary compatibility

The successful shell benchmark uses weak free-slip conditions on both boundaries. The GPlates case strongly prescribes a spatially varying surface velocity. This destroys the admissibility of the global conformal fields at a large set of surface degrees of freedom. Restricting the candidates to homogeneous strongly constrained degrees of freedom is necessary, but it does not make the resulting global fields approximate low-energy modes of the remaining operator.

### Geometry and resolution

Translations, rotations, dilation, and special-conformal fields are exact Euclidean constructions. Their finite-element representation on a curved shell depends on the geometric approximation and resolution. A global small projected matrix can identify low energy within the candidate span, but that does not guarantee that the selected span matches the errors GAMG needs to represent on each aggregate.

### Strong coefficient heterogeneity

The production viscosity field varies by nearly four orders of magnitude and contains strong lateral structure. The slow modes of this operator are therefore not expected to be captured completely by unweighted global conformal polynomials. An energy or mass weighting based on the actual heterogeneous operator may be needed, and even that may be insufficient without localisation.

### Global versus aggregate-local information

Passing ten modes directly to GAMG supplies ten localised restrictions on each aggregate. The balanced correction removes only four global vector components. These are not algebraically equivalent operations, which explains why a well-conditioned balanced correction need not reproduce the direct-ten iteration count.

The direct ten-mode method also has a real cost. Higher aggregate rank can enlarge the coarse hierarchy, increase setup time and memory, or trigger rank sensitivity. More modes are not monotonically better.

### Ritz selection metric

The current Ritz construction uses `V^T A V`. In the production case this produces a clear and stable spectral split, but the selected six-dimensional subspace performs worse than the rigid basis. Possible alternatives include a mass-weighted generalised eigenproblem, an `M^-1 A` metric based on the preconditioner actually used, aggregate-local Ritz selection, or adaptive enrichment from observed troublesome error components.

## A diagnostic caveat

The frozen benchmark logs contain repeated zero-iteration `DIVERGED_NANORINF` reports during a zero-action full-Schur velocity probe. The same pattern occurs in every solver arm. It appears to be a benchmark/instrumentation edge case in which PETSc is asked to act on a zero right-hand side, rather than evidence that one conformal variant is less stable than another. The benchmark counter should ignore or special-case zero-right-hand-side probes before this diagnostic is used for automated comparisons.

## Current conclusion

The branch demonstrates that the extra conformal modes are mathematically and computationally meaningful in selected three-dimensional weak-free-slip TALA and ALA shell problems. In those controlled cases they can reduce velocity iterations substantially.

It does **not** demonstrate a robust benefit for the production GPlates configuration. In that case:

* ordinary rigid six-mode GAMG remains the fastest tested choice;
* Ritz selection replaces useful rigid information with boundary-incompatible conformal content;
* the balanced correction preserves the baseline iteration count but adds overhead; and
* direct ten-mode enrichment increases hierarchy cost without enough iteration reduction to compensate.

I therefore do not propose changing G-ADOPT defaults. The potentially reusable outcome is an experimental, opt-in framework for constructing and testing conformal candidate spaces, together with benchmark evidence showing where it helps and where it does not.

## Questions for reviewers

I would particularly welcome views on the following:

1. Is a three-dimensional conformal near-nullspace a useful opt-in abstraction for the deviatoric TALA/ALA velocity operator, even if it is unsuitable as a default?
2. Should candidate selection use a mass-weighted generalised eigenproblem or a metric involving the current velocity preconditioner rather than `V^T A V` alone?
3. Would aggregate-local or level-local conformal enrichment be more consistent with GAMG than selecting six global Ritz combinations?
4. Is there a practical PETSc-supported route to adaptive enrichment from observed low-energy Krylov error components?
5. Should strong-boundary compatibility be used as an early criterion to disable the additional four modes?
6. Is the balanced low-rank correction useful as a diagnostic tool even if its global reductions make it unattractive at scale?
7. Which pieces, if any, are appropriate public G-ADOPT API rather than benchmark-only research code?
8. Should the nonlinear SIPG/Nitsche Jacobian changes be split immediately into a separate PR before further discussion of the preconditioner work?

## Files and public interfaces affected

The branch changes or adds:

* `gadopt/nullspaces.py`, including `ConformalKillingNearNullspace`;
* `gadopt/preconditioners.py`, including the Ritz and balanced prototypes;
* `gadopt/momentum_equation.py` and `gadopt/stokes_integrators.py` for the nonlinear tangent-stress/Nitsche treatment;
* `gadopt/approximations.py` and package exports;
* focused unit tests for nullspaces, balanced correction, and symmetry;
* three-dimensional shell and Cartesian compressible benchmarks; and
* documentation of the PETSc/GAMG aggregation considerations encountered during the investigation.

No existing G-ADOPT solver default is intentionally changed by the conformal modes. They are opt-in experiments.

## Verification performed

The following focused checks were run in the current Firedrake environment:

```bash
python -m pytest -q \
    tests/unit/test_nullspaces.py \
    tests/unit/test_balanced_conformal_pc.py \
    tests/unit/test_symmetry.py
```

Result: `117 passed`.

The same focused suite was also run with two MPI ranks:

```bash
mpiexec -n 2 python -m pytest -q \
    tests/unit/test_nullspaces.py \
    tests/unit/test_balanced_conformal_pc.py \
    tests/unit/test_symmetry.py
```

Result: the focused suite passed under two ranks.

The repository lint target was run in the Firedrake environment:

```bash
make lint
```

Result: passed.

The weak-free-slip shell benchmarks were exercised in Boussinesq, TALA, and ALA configurations, including serial, two-rank, viscosity-contrast, and higher-resolution cases. The production-shaped comparison was run separately on Gadi at 4,160 ranks. The production nonlinear comparison was deliberately capped at three SNES iterations and is not presented as a converged result.

## Proposed review status

Please treat this as a draft/RFC. I would prefer discussion of the mathematical premise, boundary-condition dependence, and appropriate PETSc/GAMG integration before trying to make any of these classes merge-ready. If parts are worth retaining, the next step should be to split the branch into small, independently reviewable changes with separate correctness and performance tests.


stephankramer:
Wow, TIL about the 10 Killing modes for deviatoric stress! Also the Ritz selection method seems like a cool idea (although in practice it doesn't seem to work so well yet as a selection criterion). Before going into to the results I got triggered by the word *necessarily* in:

> These are near-nullspace candidates, not necessarily exact null modes of the complete discretised Stokes problem.

What if these extra Killing modes provide actual exact zero modes - we should definitely be aware of that!

So playing around a bit with linear combinations I came up with the following using the notation of T_i, R_i, D, K_i for the Killing modes I came up with the following linear combination:

M_i = K_i - (R_-)^2 T_i

which is a linear combination because R_- (the inner radius) is a constant coefficient. It satisfies no-normal flow at the CMB:

M_i . x_i = 2 (e_i . x) x . x - |x|^2 e_i . x - (R_-)^2 e_i . x = (|x|^2 - (R_-)^2) e_i . x

which is zero for |x|=R_-. It also satisfies free slip everywhere (being in the kernel of stress) - so in a setup with free-slip+no-normal flow at the CMB and a free surface on top, we have three independent exact null modes in the velocity solve.

stephankramer:
Just some clarifications where I thought it was a little unclear in the explanation above vs. looking at the code:
* the constrained candidates/degrees of freedom, which aren't really explained, appear to refer to the idea of applying a homogenized bc for every bc that's imposed strongly
* where it refers to #370 above it appears it's actually based on #540 i.e. symmetrizing the Nitsche boundary terms directly, rather than symmetrizing the Jacobian afterwards making it an approximate Jacobian for the Newton solve (as was done in #370). I think the former (#540) is indeed the right way to go here: the whole analysis of low energy modes really applies to symmetric operators, and in particular here it's important that the right (near) null modes are also left (near) null modes, for which in particular we need the full stress operator being applied on the test as well as the trial in the boundary terms.

So the 3D double free slip case looks to be very promising indeed. Annoyingly, the key timing result to distinguish direct10 from Ritz6 has been left out. I can see that direct10 and Ritz6 have comparable iteration counts which are much better than rigid6, which is great. I would expect indeed with a similar iteration count Ritz6 to be faster than direct10, but I would like to have seen this confirmed by an actual timing. Otherwise there is no evidence in any of the above of the Ritz criterion actually being any beneficial (above just using all 10).

If I understand correctly that double free slip case is using a variable viscosity, but it's not non-linear, is that right? If so, there are two presumably two main differences between the production case and the double free slip case: the bcs, and the linear vs nonlinear viscosity? Or does "frozen viscosity from checkpoint" mean the "production-shaped case" was also linear(ized)? I would be good to distentangle these two. Are the plate velocities applied as a strong bc in all directions, or as a weak tangential Dirichlet condition with no-normal stress?

drhodrid:
# Response to Stephan's comments on PR #541 - with LOTS of help from codex. 

## Exact continuum null modes and the CMB-compatible combinations

Thanks Stephan. Yes, this is an important refinement of the statement in the PR.

Your three fields

$$
M_i = K_i - R_-^2 T_i
$$

span the same space, up to scaling and sign, as

$$
u_i = T_i - \frac{K_i}{R_-^2}
    = -\frac{M_i}{R_-^2}.
$$

Together with the three rotations, this gives a six-dimensional conformal space that is exactly tangent to a spherical CMB and lies in the kernel of the continuum trace-free symmetric-gradient operator. For a concentric shell with an impermeable/free-slip CMB and a genuinely traction-free outer surface, these are therefore six exact null modes of the continuum viscous velocity block.

I think it is useful, however, to distinguish that statement from several related discrete or coupled problems:

1. These are not six exact modes of the double-impermeable free-slip shell used in the existing controlled benchmark. The three $M_i$ fields generally have non-zero normal velocity at the outer boundary, so only the rotations satisfy both impermeability conditions exactly. The success of the other conformal fields in that benchmark is therefore evidence that they can be useful near-nullspace information, rather than evidence that they are exact modes of that boundary-value problem.

2. Being in the kernel of the viscous velocity block $A$ does not necessarily make these fields null modes of the complete TALA or ALA saddle-point operator

   $$
   \begin{bmatrix}
   A & B^T \\
   B & 0
   \end{bmatrix}.
   $$

   TALA and ALA impose the weighted continuity condition

   $$
   \nabla\!\cdot(\rho_0 u)=0.
   $$

   The rotations satisfy this condition for a radially varying reference density, but the three $M_i$ fields generally do not. I therefore think that "six exact null modes of the continuum viscous velocity block" is the precise statement, rather than "six exact null modes of the full TALA/ALA Stokes system."

3. They are not necessarily exact modes of the complete, time-discrete free-surface system. Once surface height and gravitational restoring traction are included, outer-boundary normal velocity couples to the surface equation. A field that is a null mode of the instantaneous viscous velocity block need not be a null mode of that coupled three-field operator.

4. On the curved finite-element mesh, CMB tangency is exact with respect to the radial normal $x/|x|$, but not necessarily with respect to the discrete facet normal used by the Nitsche terms. Interpolation of the quadratic conformal fields into the curved finite-element space introduces another small discrepancy. They therefore remain approximate modes of the assembled discrete operator.

So I agree that the current wording in the PR is too broad. I should explicitly identify this boundary-compatible six-dimensional exact nullspace of the continuum velocity block, while retaining the qualifications for the full saddle-point operator, the coupled free-surface problem, and the curved discrete operator.

Following your observation, I have prototyped locally:

- a dedicated CMB-compatible six-mode constructor using the combinations above; and
- a spherical TALA/ALA benchmark with a weakly impermeable CMB and an actual coupled free surface.

These additions are currently local and have not yet been pushed to the PR branch.

Given the distinctions above, my current inclination is to provide the CMB-compatible six to GAMG as near-nullspace information, rather than register them as an exact PETSc nullspace. Does that seem like the right interpretation to you?

## Clarifications on the implementation and benchmark comparisons

Thanks, these clarifications are all fair. Point by point:

### Constrained candidates

Yes, "constrained candidates" means applying the homogeneous form of every strongly imposed velocity Dirichlet condition to each candidate vector before orthonormalisation. It does not modify the candidates to satisfy weak Nitsche conditions. I will make that explicit in the PR.

### Relationship to PRs #370 and #540

You are also right about the nonlinear boundary formulation. Despite the misleading commit message and reference to #370, the code in this branch corresponds to the energy-consistent residual approach in #540.

It modifies the Nitsche boundary residual so that automatic differentiation produces the true symmetric Newton Jacobian. It does not replace the Jacobian afterwards by $0.5(J+J^T)$, as in the earlier quasi-Newton approach from #370.

As you say, this distinction is particularly important here because the low-energy-mode analysis assumes that right near-null modes are also left near-null modes. That requires the complete stress operator to act consistently on the test and trial functions in the boundary terms.

I will correct the PR description. I also agree that this boundary-residual dependency is logically separate from the conformal-mode experiment and should be split or rebased cleanly if any part of the conformal work is taken forward.

### Ritz6 versus direct10 timings

The missing Ritz6 versus direct10 timings were an omission. The matched warm timings, after construction of the candidates and GAMG hierarchy, are:

| Approximation | Viscosity contrast | Ritz6 | Direct10 |
| --- | ---: | ---: | ---: |
| Boussinesq | 1 | 0.635 s | 0.813 s |
| TALA | 1 | 0.856 s | 1.019 s |
| ALA | 1 | 0.792 s | 0.965 s |
| Boussinesq | $10^3$ | 1.486 s | 1.757 s |
| TALA | $10^3$ | 2.280 s | 2.519 s |
| ALA | $10^3$ | 2.254 s | 2.497 s |

For the two-rank TALA case at contrast $10^3$, Ritz6 and direct10 both required 358 accumulated velocity iterations. Their corresponding warm times were 1.568 s and 1.795 s.

Thus Ritz6 was approximately 10 to 22 percent faster than direct10 in these matched controlled solves. This supports the idea that retaining a six-column GAMG basis can be advantageous when Ritz6 and direct10 produce similar iteration counts.

These warm timings are not yet a complete cost comparison, however. Ritz6 requires ten velocity-operator actions and small global reductions whenever the selection is rebuilt, whereas direct10 increases the width and cost of the GAMG hierarchy. The warm measurements reuse an already constructed candidate basis and hierarchy. A complete comparison should therefore include:

- candidate construction and Ritz-selection time;
- GAMG hierarchy setup time;
- coarse-grid size and operator complexity;
- memory use;
- total setup-plus-solve time; and
- the frequency with which the selection and hierarchy must be rebuilt during a nonlinear solve.

The current measurements demonstrate a repeated-solve advantage in the controlled cases, but they are not yet evidence of a total setup-time or large-scale performance advantage.

### Linear and nonlinear viscosity comparisons

The controlled double-free-slip shell is linear. Its viscosity can vary strongly in space, including radially by a factor of $10^3$, but it is prescribed and does not depend on the velocity solution.

The production "frozen viscosity from checkpoint" comparison is also linearised. The viscosity was evaluated from the nonlinear checkpoint and then held fixed while the different preconditioners were compared.

The separate partial nonlinear comparison restores the complete strain-rate-dependent viscoplastic law. That experiment was capped at three SNES iterations and did not converge, so it is only supporting evidence concerning the relative cost of the inner velocity solves. It should not be presented as a converged nonlinear timing comparison.

This means that the principal differences between the successful controlled shell and the frozen production comparison are not linear versus nonlinear viscosity. They are:

- the velocity boundary conditions;
- the much stronger lateral viscosity structure;
- the geometry and its discrete representation;
- the problem size and parallel decomposition; and
- the relationship between global candidate fields and the low-energy errors seen on individual GAMG aggregates.

The partial nonlinear experiment then provides a separate check of whether the same conclusions persist when the Jacobian and preconditioner are repeatedly updated.

### GPlates surface boundary condition

The GPlates surface velocity is imposed as a strong vector Dirichlet condition in all three components. The CMB condition is weakly imposed $u\cdot n=0$, with zero tangential stress. It is not a weak tangential plate-velocity condition with a natural normal stress.

This helps explain the negative GPlates result. Homogenising the strong surface condition makes the candidate vectors algebraically admissible, but the resulting restricted vectors are no longer global conformal Killing fields or zero-stress modes. The Ritz procedure is then selecting within a candidate space that has already lost the principal mathematical property motivating it.

### Revised interpretation

I think the revised discussion should distinguish three configurations more clearly:

1. A double-impermeable free-slip shell, where the additional conformal fields can provide useful near-nullspace information but are not exact boundary-compatible modes.

2. An impermeable/free-slip CMB with a true free surface, where your three combinations together with the rotations form an exact six-dimensional nullspace of the continuum viscous velocity block, subject to the qualifications concerning weighted continuity, free-surface coupling, and discrete geometry above.

3. The production GPlates case, where the strongly prescribed surface velocity destroys the global conformal nullspace structure and none of the tested conformal enrichments has improved performance.

I will also use "conformal Killing fields" for the complete ten-dimensional space, reserving "Killing fields" for the six rigid translations and rotations.

I will update the PR explanation accordingly, add the missing timing evidence and qualifications, and keep the #540 boundary-residual work logically separate from the conformal near-nullspace experiment.

PR #545, opened 2026-09-07
Fix leak mock object in tests

### LINKED ISSUE #543: test_solver_reset leaks a mocked StokesSolver.set_solver into later tests
`tests/unit/test_stokes_solver_configuration.py:614-615` replaces a class attribute and never restores it:

```python
mock_obj = create_autospec(StokesSolver.set_solver)
StokesSolver.set_solver = mock_obj
```

Every `StokesSolver` built afterwards in the same process gets the mock, so `self.solver` is never assigned and any later `solve()` fails:

```
E  AttributeError: 'StokesSolver' object has no attribute 'solver'
gadopt/stokes_integrators.py:720
```

The suite passes today only because nothing runs after it that solves — `test_solver_reset` is last in its file, and `test_symmetry.py` only inspects `solver.F`. Any new `StokesSolver` solve test hits this.

Fix, matching the idiom already used at line 589:

```python
with patch.object(StokesSolver, "set_solver", autospec=True) as mock_obj:
    ...
    assert mock_obj.call_count == 2
```

Verified locally: reproducer fails after the config file, passes with the change. Introduced in #538.

dsroberts:
Use patch context manager to remove leaked mock object from test_solver_reset. Fixes #543

### DIFF
--- a/tests/unit/test_stokes_solver_configuration.py
+++ b/tests/unit/test_stokes_solver_configuration.py
@@ -1,6 +1,6 @@
 import firedrake as fd
 import pytest
-from unittest.mock import create_autospec, patch
+from unittest.mock import patch
 
 from copy import deepcopy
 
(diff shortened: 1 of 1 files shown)


sghelichkhani (approved):
Great! Thanks

sghelichkhani:
Failure related to the usual:

'''

FAILED demos/test_all.py::test_benchmark[multi_material/free_surface] - AssertionError: Series are different
'''

Cannot be possibly related. I am merging.

